PR #67, opened 2013-12-02
Add opus support

mark3141:
Hello,

I would like to add opus support if possible.

Thanks,
Mark


somilbansal:
Then why is the opus part in SdpInfo.cpp still commented out?

PR #101, opened 2014-04-03
Allow erizo connections to hostnames and via SSL.

byronclark:
In order to get around CORS requirements, it's helpful to be able to
connect by a DNS name instead of an IP address. Allow the hostname
provided in tokens to be configurable. Also allow setups where something
like nginx fronts erizo_controller and provides SSL termination.


andresrivas1506:
I getting error CORS someone can help me?

capture

PR #172, opened 2014-09-22
Separate Host-Specific Configuration to host.js

johnwarden:
This pull request is useful with automatic deployment scripts deploying to cloud providers.  It pulls all host-specific configuration, including public IP and host name, into a separate file called host.js.   This file can be overridden when a new licode box is deployed with just the information specific to that host.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+licode_config/
 licode_config.js
 extras/basic_example/node_modules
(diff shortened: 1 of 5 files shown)


aalonsog:
Thanks for the suggestion (and sorry for the delay). We will take into account it to re-think the configuration. If we finally decide to split it in several files it could be also beneficial to separate nuve config.

PR #179, opened 2014-09-30
Support desktopStreamId parameter in Erizo.Stream. 

johnwarden:
At Minerva Project we have built our own screen sharing extension that we use for Licode and another WebRTC project.  This extension has it's own logic for communicating with the Chrome extension via passing messages on the window -- which is differs from the logic in Erizo.GetUserMedia.

The change in this pull request allows us to use our extension by simply passing the desktopStreamId parameter returned from our extension to the Erizo.Stream constructor.

Couple of other small changes in this pull request:  always set max width/height to the available screen width/height when sharing screens.  And we always set frame rate to 30.  Shared screens look awesome when we do this.

I'm submitting this as a pull in case other people have their own Chrome extensions, and so we don't have merge conflicts if this code evolves in the future.


jcague:

Sorry for the inconvenience.

PR #211, opened 2014-11-13
fix node-gyp rebuild failing with error: flock: ./Release/linker.lock: B...

paericksen:
here is my change to fix the flock error.  I also added the LD_LIBRARY_PATH export here.


aalonsog:
Hi, thanks for the request and apologies for the delay. Could you send two different pull requests (one for the build path and another for the rabbit config)?

aalonsog:
Now I see it is here https://github.com/ging/licode/pull/212 :)

PR #216, opened 2014-11-18
[FIX]On Transport bundled never test for NULL pointer

francoisTemasys:
The only test made is:
```
if (audioSink_ == NULL && videoSink_ == NULL && fbSink_==NULL){
      return;
    }
```
but never individually on audioSink_ or videoSink_. It's possible to have fbSink and no mediaSink_.


jcague:

Sorry for the inconvenience.

PR #229, opened 2015-01-14
Trickle ice

jnoring:


jnoring:
I have more commits on top of this, including some significant improvements to the RTP packet reordering queue, an update to libav 11 and a few other things.  But the big ones are the candidate ordering and stream added event bugs.  Both of those result in setup failure.  Also, the uv_close issue I don't totally understand, but that code fixes it.

lodoyun:
I agree that sending periodic FIRs is not,by far, the optimal solution. However, it allows seeking in the recorded video, otherwise you are forced to re-encode the video. Seeking needs periodic keyframes in the file. It can be argued that almost everybody is transcoding afterwards anyway. Is that what you propose?

I'll accept it and test it for a while, then we can decide what to do with it.

The candidate ordering is a very good catch, thank you!

jnoring:
I don't disagree on the seeking, but I am unable to currently seek in mkv recorded content by licode even _with_ periodic FIR requests.  That was actually part of the reason I updated to libav 11.1 (which is an incoming patch): to see if I could fix seek in those files.  

Also, there are huge differences in media encoding for real-time transfer verses seekability; I think the recommendation to people who need seeking should be to use ffmpeg as a post-process step to re-encode the video.  Or have it be re-encoded in real time on the server (which, honestly, in x264 ultrafast mode, wouldn't be a huge burden on the system).  But I think the worst possible option is to have these periodic FIR requests, which have the potential to impact a person-to-person interaction that is really at the core of a webrtc experience.

PR #250, opened 2015-02-04
Sdp return + add candidate protection

IVBakker:


lodoyun:
That was quick.
Thanks!

PR #256, opened 2015-02-12
Trickle option + script improve

IVBakker:
Main point is the possibility to choose ice trickle or not for webrtcconnection. It's done in 2142e60.

I also added many improvement in script:
-new install dependencies for linux. Faster and smarter: check if the library isn't already there before installing (global and local)
-cmake reconfigured for better usability
-installErizo: can choose which one to install. In any case, without argument it acts the same as before.


lodoyun:
I'm testing it. So far it looks good, the trickle_ice on the server side probably should be a configuration parameter in licode_config

IVBakker:
@lodoyun That's also what I was thinking but without moving the ```nice_agent_gather_candidates(agent_, 1);``` the error here was showing up: https://github.com/ging/licode/blob/master/erizo/src/erizo/WebRtcConnection.cpp#L427

jnoring:
After doing a lot of work with trickle, I think server-side licode should be taking a slightly different approach.

I think licode should always send back full SDP with candidates as an ANSWER to a client; the reason for this is candidate gathering server-side happens so fast (on our production boxes, I literally measure milliseconds) that there is no reason to trickle candidates (possible I'm wrong about this--can anyone thing of a good reason to trickle server-side candidates?).  This approach still conforms to the trickle-ice spec, as endpoints are allowed to include candidates in SDP if they choose.  Also, this conveniently handles the "non-trickle" candidate path.

So what I'd propose: change licode to always respond with a full SDP ANSWER that contains all candidates.

IVBakker:
@jnoring With this change, setting the way the sdp is answered is done as an option on the javascript level. If nothing specified (so by default) it's sending full sdp.

@lodoyun To reproduce the thread problem that happen when the GATHERING is raised before the end of the construction of the DtlsTransport object, I suggest you to test with Firefox.

lodoyun:
Yes, I agree with the no trickle in the server by configuration and it looks good so far in my tests.
Thanks for the other additions also.

jnoring:
@IVBakker Right, I get that--what I'm saying is server should never trickle candidates.  I'd remove the option to do so entirely because there's just no good reason to do this.

PR #259, opened 2015-02-26
installUbuntuDepsUnattended.sh

vovikspbru:
check if components are already installed. Do nothing in that case


lodoyun:
Nice, thanks

PR #261, opened 2015-03-04
Run ErizoController via HTTPS - fixes #81

### LINKED ISSUE #81: Run ErizoController via HTTPS
It is not so simple to secure ErizoController. Port is hardcoded and client's getting server URL is not transparent. Client gets IP address.
When I use Licode via HTTPS, mixed content is reported. Since Firefox doesn't support mixed content by default, page is not correctly loaded by Firefox. Chrome reports "only" warnings.
This is quite important issue, because screen sharing is available only via HTTPS.

panaali:
Since the port was hardcoded ( 'server.listen(8080);' ) and ssl doesn't worked with http server(It needs https server), couple of changes applied to fix these issues.
@jcague

### DIFF
--- a/erizo_controller/erizoController/erizoController.js
+++ b/erizo_controller/erizoController/erizoController.js
@@ -3,8 +3,8 @@ var crypto = require('crypto');
 var rpcPublic = require('./rpc/rpcPublic');
 var ST = require('./Stream');
 var http = require('http');
-var server = http.createServer();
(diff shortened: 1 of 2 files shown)


ttgraham:
Was just looking at adding this myself. Any reason this was closed without merge?

panaali:
It just had a little synaxical

PR #262, opened 2015-03-06
Updated the Socket.io@1.3.4

pawanvivekananda:
Updated to new Socket.IO 1.3.4


DmitriyKarpov:
Hi! I've merged your pull request with my local branch. It seems to be a problem with erizofc.js after upgrading socket version - there is no connection after room.connect() called. I've upgraded socket.io-client too but it didn't help

pawanvivekananda:
While connecting to the server add http while connecting to the server 

Some example would look like this 

var io = require('socket.io-client')
var socket = io.connect('http://localhost:8080');

karth295:
@DmitriyKarpov: this patch works perfectly for me. Since he removed the socket.io client from erizo.js, you need to add a script tag in your HTML file.

Also, if the server that is serving the HTML/JS pages is using socket.io, you will need to upgrade that as well.

PR #274, opened 2015-03-20
RTC objects no longer act as POJSOs.

byronclark:
As of Chrome 43, JSON.stringify no long works on RTCSessionDescription
and RTCIceCandidate objects and we need to expand them manually for
signaling. This change tested on Chrome 41 and Chrome 43.

Upstream bug: https://code.google.com/p/chromium/issues/detail?id=467471


lodoyun:
Great, thanks!

PR #275, opened 2015-03-22
Fix build errors on Mac Yosemite

maojie:
1. error: no template named 'scoped_ptr' in namespace 'boost'; did you
mean 'boost::asio::detail::scoped_ptr'
2. /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/utility:294:15:
error: read-only variable is not assignable

### DIFF
--- a/erizo/src/erizo/DtlsTransport.h
+++ b/erizo/src/erizo/DtlsTransport.h
@@ -5,6 +5,7 @@
 #include <boost/thread/mutex.hpp>
 #include <boost/asio/deadline_timer.hpp>
 #include <boost/asio.hpp>
+#include <boost/scoped_ptr.hpp>
 #include "dtls/DtlsSocket.h"
 #include "NiceConnection.h"
 #include "Transport.h"
--- a/erizo/src/erizo/NiceConnection.h
+++ b/erizo/src/erizo/NiceConnection.h
@@ -8,6 +8,7 @@
 #include <string>
 #include <vector>
 #include <queue>
+#include <boost/scoped_ptr.hpp>
 #include <boost/thread.hpp>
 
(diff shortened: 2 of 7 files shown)


pawanvivekananda:
Thank you for the fix and raising the PULL request. 
I was also facing the same issue, I figured that out, but now i am struck ErizoApi Error

"no type named 'Arguments' in namespace 'v8'; did you mean 'v8::internal::Arguments'?
        static v8::Handle<v8::Value> New(const v8::Arguments& args);
                                               ^~~~~~~~~~~~~
                                               v8::internal::Arguments"

The work around that i found is to down grade the nodejs from 0.12.0 to 0.10.0 

If u have any better way without down grade the nodejs? If so can u please share your thoughts ..

maojie:
I've simply checked the changes in nodejs from v0.12.* to v.10.*, it seems that the definition of v8::Arguments is been removed in v0.12.*, so it'll generate the above errors. I'm not sure if licode works well on Ubuntu with nodejs v0.12.0, but I'll try to find a way to fix that.

mrbatista:
What is your setup? (OS version, dependencies of licode with version)
I Try to install licode with [this] (https://github.com/ging/licode/blob/master/scripts/installMacDeps.sh) script, but i have problem when compile ErizoAPI.
Thanks for your sharing.

maojie:
OS version: OSX Yosemite, Nodejs version: v0.10.6, I think the reason you can't compile ErizoAPI is due to the Nodejs's version, you may switch Nodejs version to v0.10.6 to have a try.

lodoyun:
Looks good,
Thanks!

PR #284, opened 2015-04-08
This fixes exception in setMaxBW when doing p2p audio only call

vibhorsd:
When doing p2p audio only call, `spec.video` is undefined and gets set to true:

    if (spec.video === undefined) {
        spec.video = true;
    }

This causes a crash in `setMaxBW` because sdp does not contain `m=video` media stream line.

    var a = sdp.match(/m=video.*\r\n/);
    if (a == null){
        a = sdp.match(/m=video.*\n/);
    }
    // a is still null because sdp does not contain "m=video" line
    var r = a[0] + "b=AS:" + spec.maxVideoBW + "\r\n";
    sdp = sdp.replace(a[0], r);


lodoyun:
Looks good, thanks!

PR #294, opened 2015-05-07
Moved the mapping of rtcp-fb lines till after we build a rtpmap vector

sherwinsim:
Fixes: Although the original issue was worked around with commit c07f1d46177d0e642a51e21ca761454259997d72, that commit fixes the issue at the client side and not the server side. This fix will allow Erizo to properly parse SDPs natively from Firefox 37 and up

Use Case: We use Licode server as our backend but we do not use the Licode client side.

### DIFF
--- a/erizo/src/erizo/SdpInfo.cpp
+++ b/erizo/src/erizo/SdpInfo.cpp
@@ -562,6 +562,7 @@ namespace erizo {
     std::string line;
     std::istringstream iss(sdp);
     int mlineNum = -1;
+    std::vector<std::string> tmpFeedbackVector;
 
     MediaType mtype = OTHER;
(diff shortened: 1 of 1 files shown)


zevarito:
@sherwinsim Can I ask how is to use Licode just in the backend? What are you using for Client side?

sherwinsim:
Sure, Licode drives some of the MCU based features in the Skylink.io platform. We develop a variety of client side SDKs for platform developers to use.

zevarito:
@sherwinsim that's cool!

lodoyun:
Thanks!
Sorry it took so long

sherwinsim:
No worries, I figure a fix on the server side would help things

PR #310, opened 2015-06-18
Update Homebrew installation URL for Mac dependency script because it has moved

j4y:
Fix for this warning:
```
Installing homebrew... [press Enter]
Whoops, the Homebrew installer has moved! Please instead run:

ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"

Also, please ask wherever you got this link from to update it to the above.
Thanks!
```


lodoyun:
Good, Thanks!

Issue #299, opened 2015-05-25
Interoperability Chrome, Firefox, Opera


mrbatista:
I have tested this browser on the demo [page](http://chotis2.dit.upm.es/), the scenario is the following: three users each with a different browser. The result is (on a MacBook Pro Retina Yosemite 10.10):

1. Room NO p2p: the video call works perfectly; all subscribers receive others' audio/video streams;
2. Room p2p: only Firefox receives all two subscribers, Chrome and Opera do not see Firefox subscribers. No events is fired.

Has anyone encountered this problem with a p2p room?
Version browser:
Chrome: 43.0.2357.65 (64-bit)
Firefox: 38.0.1
Opera: 29.0.1795.60

mrbatista:
The problem with p2p room is also present in my development environment with master branch

mrbatista:
I've investigated and the interoperability Chrome (or Opera) <-> Firefox is broken.
Error is: "setting Remote Description Failed to parse SessionDescription. 
a=rtcp-fb:120 nack pli Invalid SDP line."
The problem seems to be the function [changeOrder](https://github.com/ging/licode/blob/master/erizo_controller/erizoClient/src/webrtc-stacks/FirefoxStack.js#L110) in FirefoxStack.js
If remove this function i have black stream in Chrome and video stream in Firefox.

mrbatista:
@lodoyun any suggestions to resolve this issue?

zevarito:
@mrbatista newbie question: how you determine which is P2P and which is not?

mrbatista:
@zevarito if room is p2p Licode not use MCU. All token created with p2p room not use MCU. This is the logic implemented in Licode.
If you see SDP there is a line that identify a SDP forma MCU with "LicodeMCU"

zevarito:
@mrbatista 

For sure I have not clear the functionality of MCU.

* It is possible for me to select which one to use?
* Is Licode the one who decides if MCU is used or not?

THanks!

nachoaguadoc:
Hi @zevarito 

If you want to create a P2P room you have to include this option as a parameter when you call Nuve in the application script:
```
N.API.createRoom( "MyP2PRoom", function(room) {
      console.log("P2P room created with ID:", room._id);
}, errorCallback, {p2p: true});
```

mrbatista:
@lodoyun I've update to master and I did some testing:
Firefox Desktop <-> Chrome Desktop **broken**; (Mac Yosemite OS X 10.10.3)
[...]

PR #311, opened 2015-06-18
fix problem in peerConnection: "Create Answer can't be called before …

mrbatista:
* fix problem in `peerConnection`: "Create Answer can't be called before setRemoteDescription"
* Added log for debugging Session Description


lodoyun:
I was looking into this, I guess you are working on #299. After I removed changeOrder, subscriptions from chrome to firefox fail randomly, I'd bet it's a race condition.
Polishing the negotiation for p2p is a good idea.

PR #320, opened 2015-07-03
Allow to configure audio bandwidth on licode_config.js.

zevarito:


lodoyun:
I've been taking a look at this and I remembered why we don't do this.
Setting up a maxAudioBW with PCMU (the default codec in Licode) will break audio in Chrome.
We know we need to provide an easier way to provide Codec selection/configuration. We'll revisit this then.

jcague:

Sorry for the inconvenience.

PR #325, opened 2015-07-16
Added a script to allow for unattended CentOS installation.

ForestJay:
Not sure if this would be useful to you, but we use CentOS and made this change in order to automate the install.  

Cheers,
Forest


lodoyun:
Looks good in my tests.

Thanks!

PR #327, opened 2015-07-22
CloudHandler policy with callback.

ivopavlik:
I'd rather use a callback instead of returning id to set EC id for CloudHandler in case you need to read information about assigning EC to rooms from database.


aalonsog:
Thanks! I like the idea of using a callback instead of the sync assignment. 

However, in your code, if no custom method "getErizoController" is defined, erizocontroller id (in line https://github.com/ivopavlik/licode/blob/master/nuve/nuveAPI/cloudHandler.js#L208) would be undefined and no one would be selected. 

To accept the pull request I would also like to have the attempts mechanism implemented.

jcague:

Sorry for the inconvenience.

PR #334, opened 2015-07-29
Support of Socket.io 1.3.x for ErizoController

joshuairl:
Not sure if there is any other place in the codebase where I need to update the socket.io logic?

### DIFF
--- a/erizoAPI/WebRtcConnection.cc
+++ b/erizoAPI/WebRtcConnection.cc
@@ -62,9 +62,6 @@ Handle<Value> WebRtcConnection::New(const Arguments& args) {
   obj->Wrap(args.This());
   uv_async_init(uv_default_loop(), &obj->async_, &WebRtcConnection::eventsCallback); 
   uv_async_init(uv_default_loop(), &obj->asyncStats_, &WebRtcConnection::statsCallback); 
-  //obj->eventMsg = "";
-  //obj->eventSt = 0;
-  obj->statsMsg = "";
   return args.This();
 }
 
@@ -73,6 +70,8 @@ Handle<Value> WebRtcConnection::close(const Arguments& args) {
 
   WebRtcConnection* obj = ObjectWrap::Unwrap<WebRtcConnection>(args.This());
   obj->me = NULL;
+  obj->hasCallback_ = false;
+  
   uv_close((uv_handle_t*)&obj->async_, NULL);
   uv_close((uv_handle_t*)&obj->asyncStats_, NULL);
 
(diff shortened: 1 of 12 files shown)


cracker0dks:
good job. Hope this get merged. 
Have you checked that the erizio.js for the client has the right socket.io version? Or is this generated correct on the fly by setting the socketIO version?

zevarito:
I am little bit concern about SocketIO update, I am developing an IOS Client for Licode using this Objective-C socketIO compatible library: https://github.com/pkyeck/socket.IO-objc, documentation says...

```
The current version of this library does not support socket.io v1.0.
So if you want to use socket.io-objc, please fall back to v0.9.x.
```

There are forks like this one: https://github.com/francoisTemasys/socket.IO-objc that claim to support some features but not all of them.

SocketIO guys just provide a Swift version of the lib.

I am not saying it is a bad idea to update, but I wonder how many clients it could broke, so perhaps is better to think how Licode get beneficed by this update first.

Develop the Licode client for IOS is not the only project I am working on, so I would prefer to focus on update `socket.io-objc` lib after a first version of the port get released.

joshuairl:
Personally, I feel like erizo should not depend on socket.io at all.
I believe it should simply use the standard WebSocket spec.
Since currently Licode only supports browsers that support webrtc (which coincides with browsers that also support WebSockets) there really isn't much need for socket.io.
Socket.io is only beneficial for browsers that don't support websockets, but those browsers also don't support webrtc.

By dropping the dependency, you get a wider community support of native SDK's for standard WebSockets.

Dropping the dependency is probably the best course of action for the future of the Licode project.
But I understand it's not that easy all the time.

There are more developed SDK's for basic WebRTC / WebSockets than are Socket.io.

zevarito:
Yup, I agree, plain Websockets is probably a better option than upgrade the lib just to keep attached to it.

joshuairl:
Actually, the more I think about it...
I think I'm going to create a PR dropping the socket.io dependency all together.

Browsers that support webrtc support websockets.
Absolute non-sense to depend on socket.io for it's vast websocket browser support when all we care about are browsers with WebRTC.

Shouldn't be too hard as long as the erizoController / client support basics of websockets and don't take too much advantage of the generically named event names.

We just need to create a "sendMessage" on the client and the controller  that interprets all of the existing socket.on('eventNames') that are in place.

Also, is there a reason why the client has these `webrtc-stacks` versus using a larger community supported project like adapter.js to deal with the browsers differences?
Just curious.

zevarito:
I wonder how this PR came to "This branch has conflicts that must be resolved", there is a CI for Licode?

jcague:

Sorry for the inconvenience.

:+1: to get rid of socket.io anyways

Issue #309, opened 2015-06-18
Data Channel support?


vipulrastogi:
Are we planning for data channel support as well ?

sherwinsim:
We have a implementation of a working data channel. However it still needs more testing and I need to find time to clean it up before we can do a PR

ajithreddy:
By when can we expect data channel support in Licode. Is there any ETA

Thanks

mitar:
+1

vedyakov:
Is there any news?

mkhahani:
What's preventing from adding support for data channel? It shouldn't be that hard to implement, is it? Perhaps because it's not much useful while we can use WebSockets instead.

As a use case, I'm going to provide a WebRTC test page for users so they would be able to diagnose their Internet or network and see if they can communicate with the server through WebRTC. I think the best option for this test would be data channel.

I'm sure there are more use cases and I hope this happen in a near future.

lodoyun:
We're currently focused on improving videoconferencing and will be, at least for the short-to-mid term. It's not a matter of data channels being "hard" to implement but a matter of how we decide to use our resources and what we want to prioritize.
Having said that, DataChannel would be a nice addition and we would love to see a PR implementing this for Licode.

Issue #233, opened 2015-01-21
Licode should swap to BoringSSL


jnoring:
OpenSSL has some issues with DTLS handling that are causing the webrtc project to swap over to using BoringSSL entirely.  See https://groups.google.com/forum/#!topic/discuss-webrtc/TqZ9N0eTn24 and more importantly, https://code.google.com/p/chromium/issues/detail?id=447431

zevarito:
+1

zevarito:
@jnoring did you attempt to do this?

BoringSSL and OpenSSL share the same API, is just a build change?

jnoring:
I did not attempt this, although I think this would be a wise move for the project.  If I have time in the coming weeks, I'll see if I can put in a change for this.

Issue #232, opened 2015-01-20
Licode should swap to ffmpeg


jnoring:
Licode currently uses libav, which is a fork of the ffmpeg project.  Although it wasn't clear to me a year ago, it's pretty clear that ffmpeg is a better overall project and will be the long-term victor in that debate.  Here's why:

* libav is a subset of ffmpeg features.  ffmpeg generally accepts patches from libav, but not vice-versa.
* the project activity on ffmpeg dwarfs libav.  Looking at messages posted to their usergroup, ffmpeg has more in a day than libav has in some _months_.
* ffmpeg now has periodic releases instead of a "use trunk!" attitude
* may help fix seeking issues in our mkv files.  currently libav isn't handling this well, and it's probably because they really see very little activity these days and their mkv/vp8 muxing is probably doing something wrong.

I'm happy to supply a patch to do this, assuming the devs here agree?

lodoyun:
I agree. There is nothing tying us to libav and it does look like ffmpeg has seen more activity since we decided to go the other way. :+1: 

If you need any help with the patch I can allocate some time for this.

zevarito:
Guys, has Licode switched to FFMpeg ?

pritammobisoft:
This got me thinking about licensing of Licode. Parts of FFMPEG are GPL v3 and parts are LGPLv3. The Licode project's license is stated as MIT. 

But as per [this question](https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l) on open source stack exchange, if you include a GPL v3 dependency in your program then the whole program "might" be required to use GPL v3 license. 

So if Licode wants to stay MIT I think the core developers need to think more about FFMPEG depedency. I am not a lawyer, but just wanted to mention this, as unsuspecting users of Licode might use it because of it's MIT license and later might be surprised that they are required to follow GPL.

PR #339, opened 2015-08-18
Socket io 1

patrickm:
ee13d7c Removes socket.io (client) as dependency for bundling erizo.js
Motivation: many client libraries using distributed erizo.js client code won't need extra bundled socket.io, and it's true they can tools/compile.* it again without included socket.io it would be nice to have it aside. Erizo api consumers will be able to simply take distributed erizo.js file and use own socket.io already present in their environment (amd,requirejs cases)

cc12509 Updates to new socket.io server and client (1.3.6)
Because we can't easily use new socket.io-client with GCC, ee13d7c makes here even more sense.


mkhahani:
I wonder is there any reason this PR has not been merged?! I have requested for the migration here #476 and enumerated advantages of Socket.io 1.

jnoring:
Just speculating, but it likely hasn't been merged because it would be a breaking change for many people.  0.9 is not compatible with 1.0+ servers, so updating for many people is a lot more complicated than a version bump server-side.

mkhahani:
You're right @jnoring, the two versions are not compatible but using the old version in the core also prevents people from using the newer version in their apps and benefiting from a lot of features and enhancements due to version conflicts (on the client side at least).

I don't see any milestone or versioning plan for Licode and this means we have to stick to an outdated, poor and buggy module which is for five years ago!

zevarito:
@mkhahani having and old version do not prevent you from using a newer one, the trick is how you configure `socketIO`, if you serve any version of the lib from `/`, all apps trying to fetch the lib from there will fetch that version, perhaps Licode is wrong about hardcoding the way it serves `socketIO` but if you want to install a newer version to be used by other apps you could use this parameter for example:

```javascript
var io = require('socket.io')(http, {path: "/ackuaria/socket.io/"});
```

cracker0dks:
@zevarito yeah, than everyone who want to use an old version should do this that way.

What you should not do, is using old dependencies because someone have to update his code as well.
My opinion: @mkhahani is right, this project should get a versioning plan.

mkhahani:
Thanks for the responses.

@zevarito the problem is in the client side. Assume a single page application which is talking with ErizoClient via the old version of the Socket.io module and at the same time needs to load the newer version of the Socket.io client to connect to the app server which is running on version 1+.

zevarito:
@mkhahani There is no problem with that, you can easily fetch two versions of SocketIO from client side.

1) At server side you require SocketIO serving it from a custom path:

```javascript
var io = require('socket.io')(http, {path: "/path/socket.io/"});
```

2) Then at client side you load the library like this:

Place the code bellow right before Licode's socketIO is fetched.
```html
<script src="/path/socket.io.js"></script>
    <script type="text/javascript">
       // Keep `io` before Licode SocketIO gets loaded and override it. 
       ioSafe = io;
    </script>
</script>
```
The result should be...
`ioSafe` -> version 1.x
`io` -> Licode's socketIO version (0.96)

3) Then to fetch 1.x version from the server you need to do this in your client side code:

```javascript
var socket = ioSafe('/', {path: '/path/socket.io'});
```

I agree that could be nice to upgrade to SocketIO and that will be cool if Licode starts to versioning releases.

patrickm:
A proper versioning plan is a must, but I suppose noone has time for it now.
Some people probably can't allow to include 2 socket.io versions because of the overhead or simply extra KB.

I don't really see the problem, if licode needs to support older, deprecated socket.io 0.9, then we can:
- support on server both versions, controlled by USE_OLD_SOCKET_IO = true (let core maintainers decide if this should be on/off by default)
- front end clients then can use any (old or new) version, they will just need to adjust USE_OLD_SOCKET_IO on the server.

This way **everyone** is happy, more importantly other users are happier because they are not forced to use deprecated socket.io.

Then deprecation or versioning plan is made and old version can be marked as soon to be deprecated.

mkhahani:
@zevarito Are you sure the trick works just by overriding the `io` variable? Any other global variables or side effects?

@patrickm We don't have to do much for versioning. We can simply start a new version (e.g. v2) just by creating a new branch and targeting some major changes. Then we have a working copy that is supposed to include new features and enhancements. People can also use it with care due to instability. We would have to commit minor changes and bug fixes to both branches for a while but our main focus is now on the new version. This way people have time to migrate to the new version and we can deprecate the old version after the new one got stable.

Right now I can see some [Issues](https://github.com/ging/licode/issues) which are held because they need some surgery to be done or they create incompatibility issues. By creating the new branch we can assign them to a new milestone and work on them without compatibility or stability concerns. Here are some of them and I'm sure there would be more:

- Migrate to Socket.io 1 #476
- Replace browsers stack by Adapter.js  #374 
- Replace OpenSSL by BoringSSL #233 
- Replace libav by ffmpeg #232 
- Add VP9 support #415
- Add Data Channel support #309
- Add Opus support by default #420

zevarito:
@mkhahani it works for me.

image

mkhahani:
@zevarito Thank you. It works well. I just didn't get the point with server path`. The two servers are running on different ports so using `path` does not make sense, right?

zevarito:
@mkhahani doesn't look like SocketIO client is able to define the port as-well, but in that case you can proxy it from your webserver.

mkhahani:
@zevarito The port is part of the URL and I have no problem with it:
```
socket = io('http://www.mydomain.com:3000/');
socketNew = ioNew('http://www.mydomain.com:3001/');
```

Although I'm curios to know which one is preferred, running two SocketIO server instances (despite of the their versions) or a single instance serving on different path's. Any one has idea?

PR #364, opened 2015-11-02
basic example code modified

Mido22:
changes made:
  - removed the unused `getParameterByName` function
  - moved code around, broke them into smaller functions to improve readability( hopefully)
  - added an option to differentiate local stream from remote stream, also slightly modified the logging.
  - finally, now the user can see the streams of the those who joined the room before him.

### DIFF
--- a/extras/basic_example/public/connection_test.js
+++ b/extras/basic_example/public/connection_test.js
@@ -1,90 +1,91 @@
-var serverUrl = "/";
-var localStream, room;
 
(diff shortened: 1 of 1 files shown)


nachoaguadoc:
Hi Mido22,

We appreciate your contribution, but we consider that they're all minor changes not required for a general scenario. As far as we know the possibility to see the streams of those who joined the room before the user has always worked fine.

Best regards

PR #376, opened 2015-12-03
Erizo Controller policy fix

ivopavlik:
Erizo Controller policy always use "ErizoAgent" message queue since bind is hardcoded as amqper.bind("ErizoAgent"); and no value from getErizoAgent method is used.
Should be method getErizoAgent used also for deleteErizoJS?

### DIFF
--- a/erizo_controller/erizoAgent/erizoAgent.js
+++ b/erizo_controller/erizoAgent/erizoAgent.js
@@ -67,8 +67,14 @@ for (var prop in opt.options) {
 var logger = require('./../common/logger').logger;
 var amqper = require('./../common/amqper');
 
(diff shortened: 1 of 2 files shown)


aalonsog:
Many thanks for the request. However, erizo controller policies are though to choose which erizo agent select when trying to allocate a new erizoJS. Thus, does not make sense to apply them inside erizo agents. Agents bind to the common queue "ErizoAgent" but also to their specific queue "ErizoAgent_{id}" (as you can see in [here](https://github.com/ging/licode/blob/master/erizo_controller/erizoAgent/erizoAgent.js#L278). That's the queue used by erizo controller when selecting an specific agent using the policies. 

BR

Issue #375, opened 2015-12-03
Can't stream video only stream with Chrome 47


ivopavlik:
After my Chrome (Windows 8.1) was updated, I can't broadcast stream without audio. Candidates are gathered, but no stream is available in room.remoteStreams.

j4y:
I'm having the same issue.  I wonder if this is a licode bug or a Chrome bug?  It worked fine in Chrome 46.

I opened an issue with the Chromium guys:
https://code.google.com/p/chromium/issues/detail?id=564883

j4y:
I've also confirmed that you can't stream with audio only.

ivopavlik:
My tests with Chrome 47 (note: if you want to use getUserMedia, HTTPS is required):
Licode basic sample (also with test connection functionality) worked fine when both, audio and video are available.
When I set constraints audio (or video) to false, no media data were transferred.
I tried the same on machine with camera only (no microphone is available). It was because I wanted to try https://opentokrtc.com/, where it is quite difficult to affect constraints for getUserMedia. As I supposed, no media information was transferred within Licode. For https://opentokrtc.com/ everything was OK.

So I don't think this is a Chrome issue.

prospdragon:
I have tested on HTTPS with Chrome 47.
And success to access to local media.
But I get error in console ;
https://xxx.com:8080/socket.io/1/?t=1449226725925 net::ERR_CONNECTION_CLOSED
How to fix it?

ivopavlik:
ErizoController doesn't need to run. Otherwise, have you set value config.erizoController.listen_ssl (also with config.erizoController.ssl) to true. You should request page on port 3004 instead of 3001. If you don't have valid certificate, first in the session call https://<ip address>:8080/socket.io/ and confirm certificate exception.

prospdragon:
I fixed this issue as change licode_config.js
Thank you @ivopavlik

j4y:
Note that there is an error sent to STDOUT when running Chrome in debug mode.

`[42142:4935:1204/114557:ERROR:webrtcsession.cc(337)] Failed to set remote answer sdp: Failed to enable BUNDLE.`

From what I understand `BUNDLE` is when use the same port is used for sending video and audio packets.  The packets are determined to be audio or video by inspecting the SSRC ids

lodoyun:
In understand #381 also fixes this

PR #381, opened 2015-12-08
fix sdp offer bundle parsing when streaming video-only or audio-only

j4y:
The fix is to simply change the 4 to a 3 so that the conditional code to add the BundleTag objects to the bundleTags vector.  Credit to Jeremy Noring on the lynckia google group for finding this fix.

In the case of video-only `a=group:BUNDLE video` or audio-only `a=group:BUNDLE audio`, the number of parts is only 3.  Without the bundle tags, the remote description had an invalid bundle definition `a=group:BUNDLE`.  Most of the time with WebRTC the bundle definition will be `a=group:BUNDLE audio video` for audio and video streaming.

This fixes video-only and audio-only streaming in Chrome 47. issue #375

### DIFF
--- a/erizo/src/erizo/SdpInfo.cpp
+++ b/erizo/src/erizo/SdpInfo.cpp
@@ -648,8 +648,9 @@ namespace erizo {
           ELOG_DEBUG("BUNDLE sdp");
           isBundle = true;
         }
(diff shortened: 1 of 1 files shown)


lodoyun:
Thanks!

PR #404, opened 2016-01-21
Fix roomController::removeSubscriptions so that subscribers no longer…

brianfields:
… receive stream bits after disconnecting

This appears to be a simple coding bug, where elements of an array were assumed to be strings when they are really dictionaries. The result of this bug is that a client that subscribes to a stream continues to receive bits for that stream even after she disconnects.

See https://groups.google.com/forum/#!topic/lynckia/1Je8jj-5wUo


lodoyun:
Yes, will check every use of that dictionary... Thanks!

PR #437, opened 2016-04-18
Do not use deprecated Firefox media functions

cracker0dks:
Change to current media functions
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getUserMedia

### DIFF
--- a/erizo_controller/erizoClient/src/Connection.js
+++ b/erizo_controller/erizoClient/src/Connection.js
@@ -67,9 +67,8 @@ Erizo.getBrowser = function () {
 Erizo.GetUserMedia = function (config, callback, error) {
     "use strict";
 
-    navigator.getMedia = ( navigator.getUserMedia ||
+    navigator.getMedia = (navigator.getUserMedia ||
                        navigator.webkitGetUserMedia ||
-                       navigator.mozGetUserMedia ||
                        navigator.msGetUserMedia);
 
     if (config.screen){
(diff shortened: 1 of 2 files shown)


jnoring:
This isn't backwards compatible with old versions of firefox.  What Licode really needs to do is start using adapter.js instead of providing its own shims for getUserMedia().

cracker0dks:
yeah maybe you are right. But who i using webRTC with an old browser ? adapter.js would be nice.

zevarito:
+1 to use AdapterJs.

2016-04-18 13:15 GMT-03:00 Cracker <notifications@github.com>:

> yeah maybe you are right. But who i using webRTC with an old browser ?
> adapter.js would be nice.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/ging/licode/pull/437#issuecomment-211451354>
>



-- 
Alvaro

jnoring:
I have lots of users who unfortunately use WebRTC with old browsers. :(

enricop89:
Why did you close it? It seems to work well

cracker0dks:
because you are not backwards compatible and so this will probably not get merged. It is also better to implement adapter.js than merge this :)

enricop89:
It sound reasonable. Have you already tried to implement adapter.js?

cracker0dks:
no not yet.  If I get problems with the current solution maybe I will. 

PS: You also dont know if they will merge this because there are many other updates like node, socket.io or ssl witch dont get merged (for most undefined reasons). So if you change your code to adapter.js you will probably get trouble every time you pull an update from this code base. #NotWorthATM

PR #445, opened 2016-05-06
bugfix in pull request

kekkokk:
forgot to move fillErizos() inside the aws callack caused this pull request to be ineffective.


aalonsog:
Thank you for the fix. However it seems you have forgotten to add fillErizos function in case no IP and not amazon are set by configuration (line https://github.com/ging/licode/pull/445/files#diff-e3cda258841aa4572d51c64b5078383cL247)

kekkokk:
Ok, I'll try to do a major change for everything concern aws in another pull request.

kekkokk:
switched to: https://github.com/ging/licode/pull/507

PR #469, opened 2016-06-02
custom location for SSL creds

thehunmonkgroup:
adds the following keys to licode_config.js
 - config.erizoController.ssl_key
 - config.erizoController.ssl_cert

allows specifying a custom location for SSL key/cert


aalonsog:
Thanks!

PR #470, opened 2016-06-02
provide config.erizoAgent.instanceLogDir

thehunmonkgroup:
Used to declare the log file location for spawned erizo instances.
Allows erizoAgent to more easily be run as non-root user.


lodoyun:
Nice, thanks!

PR #471, opened 2016-06-02
Provide valid ID for local stream

thehunmonkgroup:
Local streams currently get an undefined ID in getID().


aalonsog:
When a local stream is published in a room it actually has an id. So this is not an acceptable solution. An option could be to check if the stream is local and it has not an id. Only then return "local" as the id (or other string saying it is not published).

thehunmonkgroup:
Ah, I didn't realize it was only for the case of unpublished local streams. I've updated the PR to use the 'local' ID only in this case. Works fine in my tests.

aalonsog:
Great, thanks!

PR #472, opened 2016-06-02
add label/bar options for showing streams.

thehunmonkgroup:
This adds a 'label' and 'bar' key to the options arg for Erizo.Stream.show()

This allows:
- The same stream to be placed in multiple elements on the page using a
  custom label for each invocation of stream.show(), eg {label: 'foo'}
- The bottom bar to not be rendered if {bar: false} is passed. In this
  case, a 'media' prop is set on the stream object, to provide a
  consistent interface to the underlying audio or video object.

It does create the side effect of multiple independent bar controls

### DIFF
--- a/erizo_controller/erizoClient/src/views/AudioPlayer.js
+++ b/erizo_controller/erizoClient/src/views/AudioPlayer.js
@@ -80,14 +80,20 @@ Erizo.AudioPlayer = function (spec) {
         that.div.appendChild(that.audio);
 
         // Bottom Bar
-        that.bar = new Erizo.Bar({elementID: 'player_' + that.id,
-                                  id: that.id,
-                                  stream: spec.stream,
-                                  media: that.audio,
-                                  options: spec.options});
+        if (spec.options.bar !== false) {
+            that.bar = new Erizo.Bar({elementID: 'player_' + that.id,
+                                      id: that.id,
+                                      stream: spec.stream,
+                                      media: that.audio,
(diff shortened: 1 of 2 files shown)


aalonsog:
Thanks for the proposals. The second one is ok. However I see a problem in the first one: you add the possibility of showing a single stream in multiple divs. But you don't provide a way of removing them individually. So the behavior is more or less the same than now (you only add an identifier in the audio/video tag). Here my proposal: 

- A Stream object has a hash table with a relation between html elements and audio/video players.
- When calling function play(elementID), the Stream object creates a new audio/video player and saves the entry elementID:player in the hash.
- If you call the stop() function without arguments, the Stream object removes all the existing players.
- If you call the stop(elementID) function, the Stream object removes the player associated with that element (if it exists)

So my suggestion is that you remove this part from this pull request (keeping only the hide bar feature) and create a new pull request with this behavior.

thehunmonkgroup:
Sure, this makes sense. I'd like to wait on #474 to land before I proceed with your request, as I think that will have implications for the implementation.

jcague:
@thehunmonkgroup I just merged #474 can you continue with this one as @aalonsog suggested please?

thehunmonkgroup:
ok, i've refactored this PR to include only the 'bar' option for Erizo.Stream.show()

will work on the 'label' option in another PR.

let me know if there's anything else you need in order to accept this PR

PR #491, opened 2016-06-28
Fix memory leak ExternalOutput

kekkokk:
Fix a memory leak in ExternalOutput which appear when the frame dimensions are bigger than buffer (especially when resolution > 640 x 480)
The frames are now dropped, preventing erizoJs to crash.


lodoyun:
Thanks

PR #505, opened 2016-07-26
Change shebang to env bash for increased portability.

btamayo:
Previously the shebang line was: `#!/bin/bash` which is not always the location for `bash` for all systems.

Changing it to `#!/usr/bin/env bash` will use the first `bash` found in `PATH` and increase portability across more platforms.

Reference: http://stackoverflow.com/questions/10376206/what-is-the-preferred-bash-shebang


jcague:
:+1: thanks!! can you resolve the conflicts please?

btamayo:
@jcague Done! Sorry for the wait :)

PR #506, opened 2016-07-26
Add a warning when using an unsupported version of node for Erizo

btamayo:
Since many of us use Node Version Manager to switch between different versions of node, it would be helpful to show a warning if we're using the wrong version of node to compile Erizo.


lodoyun:
Useful. Thanks!

PR #507, opened 2016-07-26
Replaced aws-lib with aws-sdk

kekkokk:
related to: https://github.com/ging/licode/pull/445

I decided to switch from aws-lib to aws-sdk (which is the official one I think).
Anyway, it appears that using AWS.Metadata we are able to obtain our public ip from amazon without the need to create an IAM role and hardcode our public / secret aws keys in our code and without specify the region.

rapidly tested and seems working.

### DIFF
--- a/erizo_controller/erizoAgent/erizoAgent.js
+++ b/erizo_controller/erizoAgent/erizoAgent.js
@@ -244,32 +244,34 @@ for (k in interfaces) {
 
 privateIP = addresses[0];
 
-if (GLOBAL.config.erizoAgent.publicIP === '' || GLOBAL.config.erizoAgent.publicIP === undefined){
+if (GLOBAL.config.erizoAgent.publicIP === '' || GLOBAL.config.erizoAgent.publicIP === undefined) {
     publicIP = addresses[0];
-    if(global.config.cloudProvider.name === 'amazon'){ 
-        var opt = {version: '2012-12-01'};
-        if (GLOBAL.config.cloudProvider.host !== '') {
-            opt.host = GLOBAL.config.cloudProvider.host;
-        }
-        ec2 = require('aws-lib').createEC2Client(GLOBAL.config.cloudProvider.accessKey, GLOBAL.config.cloudProvider.secretAccessKey, opt);
-        ec2.call('DescribeInstances', {'Filter.1.Name':'private-ip-address', 'Filter.1.Value':privateIP}, function (err, response) {
+    if (global.config.cloudProvider.name === 'amazon') {
+        var AWS = require('aws-sdk');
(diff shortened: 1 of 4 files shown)


erizo_controller/erizoAgent/erizoAgent.js
```diff
@@ -244,33 +244,35 @@ for (k in interfaces) {
...
 } else {
     publicIP = GLOBAL.config.erizoAgent.publicIP;
+    fillErizos();
 }
 
 // Will clean all erizoJS on those signals
 process.on('SIGINT', cleanErizos); 
 process.on('SIGTERM', cleanErizos);
 
-fillErizos();
```
jcague: why do you change this behaviour? if there's an issue here please submit a different PR.

erizo_controller/erizoAgent/erizoAgent.js
```diff
@@ -244,33 +244,35 @@ for (k in interfaces) {
...
     publicIP = addresses[0];
-    if(global.config.cloudProvider.name === 'amazon'){ 
-        var opt = {version: '2012-12-01'};
-        if (GLOBAL.config.cloudProvider.host !== '') {
-            opt.host = GLOBAL.config.cloudProvider.host;
-        }
-        ec2 = require('aws-lib').createEC2Client(GLOBAL.config.cloudProvider.accessKey, GLOBAL.config.cloudProvider.secretAccessKey, opt);
-        ec2.call('DescribeInstances', {'Filter.1.Name':'private-ip-address', 'Filter.1.Value':privateIP}, function (err, response) {
+    if (global.config.cloudProvider.name === 'amazon') {
+        var AWS = require('aws-sdk');
```
jcague: please put the require at the beginning of the file.

nuve/nuveAPI/cloudHandler.js
```diff
@@ -6,7 +6,7 @@ var logger = require('./logger').logger;
 // Logger
 var log = logger.getLogger("CloudHandler");
 
-var ec2;
+var AWS;
```
jcague: prefer setting `var aws = require('aws-sdk')` at the top of the file

erizo_controller/erizoAgent/erizoAgent.js
kekkokk: cause if you call fillErizos before amazon responded us with an ip or an error, this generated erizo will have a private ip instead of the public one we want. 

I moved this function in the aws callback. 

I'll follow the other suggestions and open a new PR, with everything linted and tested

jcague: I got it now, thanks!

PR #509, opened 2016-07-27
Fix L.Resizer error when bundling / webpack

nionis:
this.L = this.L = {} was the cause of multiple errors while I was making a library for wrapping erizo client for use in webpack / browserify etc, similar to [this](https://www.npmjs.com/package/erizo-client) but with auto-updating and a respiratory for people to contribute


nionis:
[moved to](https://github.com/ging/licode/pull/514)

PR #513, opened 2016-07-30
urgent: fix for mongojs installed which not compatible with node 0.10.x

nionis:
The mongojs lib was updated from 2.3.0 -> 2.4.0 which uses some features like object.assign which are not supported by 0.10.x.

I would suggest making a proper package.json file for all the needed libs. Its actually quite brittle the way it is now.


lodoyun:
I didn't experience any problems with this but it makes sense to lock the version no.
I agree about the proper package.json and we'll get to it soon.

PR #515, opened 2016-07-30
fixed on getBrowser function

nionis:
getBrowser function thinks environment is running in node when its bundled in webpack


nionis:
@jcague basically webpack exposes a global variable 'module' that getBrowser function is using to detect whether it runs or node or not, so it thinks it's on node when it's not

jcague:
I see, thanks for the PR! I've tested it locally and it works in Node.

Can you resolve the conflicts please?

PR #527, opened 2016-09-14
Check payload before setting videoSourceSsrc_

kekkokk:
It happens that sometimes the first packet is not a RED_90000_PT payload type but instead a 72.
This will cause the wrong videoSourceSsrc_ to be set for the Feedback sink, causing:

WARN: WebRtcConnection - Unknown SSRC in deliverFeedback_  3682841650, localVideo 4216015602, remoteVideo 55543, ignoring at every keyframe request


jcague:
what if the video packets come without RED?

jcague:
thanks for the PR @kekkokk! it looks to me that we might consider a different approach, please fill in an issue where we could discuss possibilities.

PR #535, opened 2016-10-05
Add lint option to cmake

jcague:


jcague:
@lodoyun :+1:

jcague:
@lodoyun I moved `cpplint.py` to a folder called `utils`

PR #538, opened 2016-10-05
Added identifier for every erizo log using log4cxx MDC

lodoyun:
Erizo will now include contextual information on every log message.
That information allows to assign any operation to it's publisher, subscriber, externalInput or externalOutput.
It's implemented using log4cxx MDC which means we have to manage the context depending on the thread we're. 
Rule of thumb: we update the context for every call that is triggered by the node main thread and we copy the context for every new thread we start.

We clear the context every time we update it because updating does not work in log4cxx and the MDC constructor and destructor also don't work.

### DIFF
--- a/erizo/src/erizo/DtlsTransport.cpp
+++ b/erizo/src/erizo/DtlsTransport.cpp
@@ -16,10 +16,12 @@ DEFINE_LOGGER(Resender, "Resender");
 
 Resender::Resender(boost::shared_ptr<NiceConnection> nice, unsigned int comp, const unsigned char* data, unsigned int len) : 
   nice_(nice), comp_(comp), data_(data),len_(len), timer(service) {
+    logContext_ = GET_LOG_CONTEXT();
     sent_ = 0;
   }
 
 Resender::~Resender() {
+  UPDATE_LOG_CONTEXT(logContext_);
   ELOG_DEBUG("Resender destructor");
   timer.cancel();
   if (thread_.get()!=NULL) {
@@ -35,6 +37,7 @@ void Resender::cancel() {
 }
 
 void Resender::start() {
+  UPDATE_LOG_CONTEXT(logContext_);
   sent_ = 0;
   timer.cancel();
   if (thread_.get()!=NULL) {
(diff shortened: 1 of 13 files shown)


erizo/src/erizo/logger.h
```diff
@@ -101,4 +102,15 @@
 #define ELOG_FATAL(fmt, args...) \
     ELOG_FATAL2( logger, fmt, ##args );
 
+static const std::string LOG_CONTEXT_ID = "connId";
```
jcague: could it be called just "id"? or "componentId"? I'm saying that because "connId" seems to refer to connection and I'm not sure it will be always the case.

erizo/src/erizo/NiceConnection.cpp
```diff
@@ -59,6 +59,7 @@ namespace erizo {
     mediaType(med), agent_(NULL), loop_(NULL), listener_(listener), candsDelivered_(0), 
     iceState_(NICE_INITIAL), iceComponents_(iceComponents), username_(username), password_(password), iceConfig_(iceConfig),
   receivedLastCandidate_(false){
+      logContext_=GET_LOG_CONTEXT();
```
jcague: consider `logContext_ = GET_LOG_CONTEXT();`

erizo/src/erizo/OneToManyProcessor.cpp
```diff
@@ -75,6 +79,7 @@ namespace erizo {
 
   void OneToManyProcessor::addSubscriber(MediaSink* webRtcConn,
       const std::string& peerId) {
+    UPDATE_LOG_CONTEXT(id_);
```
jcague: why do we need to update the context in all these methods?

erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -16,10 +16,12 @@ DEFINE_LOGGER(Resender, "Resender");
 
 Resender::Resender(boost::shared_ptr<NiceConnection> nice, unsigned int comp, const unsigned char* data, unsigned int len) : 
   nice_(nice), comp_(comp), data_(data),len_(len), timer(service) {
+    logContext_ = GET_LOG_CONTEXT();
```
jcague: what's the value we receive from GET_LOG_CONTEXT() here?

lodoyun:
We've decided to go the manual route for now

PR #539, opened 2016-10-06
Change/third party libs build in cmake [WIP]

jcague:


jcague:
We'll postpone this PR

PR #547, opened 2016-10-11
Fixed bug that could crash ErizoJSController when removing an

lodoyun:
externalInput


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -203,7 +203,7 @@ exports.ErizoJSController = function (spec) {
             var muxer = new addon.OneToManyProcessor(),
                 ei = new addon.ExternalInput(url);
 
-            publishers[from] = {muxer: muxer};
+            publishers[from] = {muxer: muxer, wrtc:ei};
```
jcague: consider `wrtc: ei` (note the blank space)

PR #556, opened 2016-10-14
Update ErizoAPI to node 6

jcague:
This PR is now a complete proposal to update ErizoAPI to be able to use from Node 0.10 to Node 6 (latest LTS).


.travis.yml
```diff
@@ -8,8 +8,11 @@ cache:
   directories:
   - node_modules
   - cache
+  - build/libdeps
 node_js:
   - 0.10
+  - 4
```
lodoyun: I'm not sure about this.
We should probably concentrate supporting a single node version (6 or 4) and remove support for the others

scripts/installUbuntuDeps.sh
```diff
@@ -55,13 +55,27 @@ install_apt_deps(){
...
+}
+
 install_openssl(){
   if [ -d $LIB_DIR ]; then
     cd $LIB_DIR
-    if [ ! -f ./openssl-1.0.1l.tar.gz ]; then
-      curl -O https://www.openssl.org/source/old/1.0.1/openssl-1.0.1l.tar.gz
-      tar -zxvf openssl-1.0.1l.tar.gz
-      cd openssl-1.0.1l
+    OPENSSL_VERSION=`node -pe process.versions.openssl`
```
lodoyun: Same as above, changing openssl versions with node makes me uncomfortable. I would prefer having a fixed node + openssl version as we did before.

jcague:
I agree @lodoyun, I'd support the latest LTS (6) now that we're compatible, but I'd also support Node 0.10 during the transition. At least, until we make sure everything is working right with Node 6. What about having tests running on Node 0.10 and Node 6 for 2 weeks and then drop Node 0.10 support?

PR #562, opened 2016-10-17
Adapt ErizoClient to jshint

jcague:


erizo_controller/erizoClient/lib/socket.io.js
```diff
@@ -1,3 +1,4 @@
+// jshint ignore:start
```
lodoyun: Maybe we could add erizoClient/lib to .jshintignore

PR #564, opened 2016-10-18
Adapt Spine, Erizo Agent and Basic Example to jshint

jcague:


extras/basic_example/public/script.js
```diff
@@ -55,66 +63,65 @@ window.onload = function () {
...
         for (var index in streams) {
           var stream = streams[index];
           if (localStream.getID() !== stream.getID()) {
             room.subscribe(stream);
-            stream.addEventListener("bandwidth-alert", function (evt){
-                console.log("Bandwidth Alert", evt.msg, evt.bandwidth);
-            });
-            
-
+            stream.addEventListener('bandwidth-alert', cb);
```
lodoyun: Just curious, what's wrong with having the callback function here?

spine/erizofc.js
```diff
@@ -0,0 +1,93 @@
+/*
```
lodoyun: This file is copied in installSpine.sh, should be in .gitignore

jcague: good catch! I'll check it, I added all the folder to save time 😞

extras/basic_example/public/script.js
jcague: Creating functions within a loop is discouraged in general: http://airbnb.io/javascript/#functions--in-blocks

PR #576, opened 2016-10-24
Add/nice connection ut

lodoyun:
Added UTs for NiceConnection:
In order to be able to mock libnice, separated the interface and the implementation of the calls to the library.

### DIFF
--- a/erizo/src/test/NiceConnectionTest.cpp
+++ b/erizo/src/test/NiceConnectionTest.cpp
@@ -77,12 +77,19 @@ class NiceConnectionStartTest : public ::testing::Test {
 class NiceConnectionTest : public ::testing::Test {
  protected:
   virtual void SetUp() {
+    const std::string kArbitraryLocalCredentialUsername = "ufrag";
+    const std::string kArbitraryLocalCredentialPassword = "upass";
+    const std::string kArbitraryConnectionId = "a_connection_id";
+    const std::string kArbitraryTransportName = "video";
+    const std::string kArbitraryDataPacket = "test";
+
     libnice = new MockLibNice;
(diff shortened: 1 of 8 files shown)


erizo/src/test/NiceConnectionTest.cpp
```diff
@@ -0,0 +1,426 @@
...
+      "ufrag",
+      "pass");
+  nice.start();
+}
+
+TEST_F(NiceConnectionStartTest, start_Configures_Libnice_With_Port_Range) {
+  char *ufrag = strdup("ufrag");
+  char *pass = strdup("pass");
+
+  ice_config->minPort = 10;
```
jcague: consider using constants for this arbitrary numbers, it improves readability. For instance: kArbitraryMinPort and kArbitraryMaxPort. Or use ice_config->minPort below as you do in other tests.

erizo/src/test/NiceConnectionTest.cpp
```diff
@@ -0,0 +1,426 @@
...
+
+  std::vector<erizo::CandidateInfo> candidate_list;
+  candidate_list.push_back(test_candidate);
+
+  EXPECT_CALL(*libnice, NiceAgentSetRemoteCandidates(_, _, _, _)).Times(1);
+  nice_connection->setRemoteCandidates(candidate_list, true);
+}
+
+TEST_F(NiceConnectionTest, queuePacket_QueuedPackets_Can_Be_getPacket_When_Ready) {
+  char* test_packet = strdup("tester");
```
jcague: maybe it makes sense to move test_packet creation and deletion to the NiceConnectionTest since you're using it in more tests.

PR #589, opened 2016-11-03
Use gcc/g++ 5 in ubuntu

lodoyun:
Also, c++11!


lodoyun:
https://travis-ci.org/ging/licode/builds/173036317#L6946
Yay!

PR #597, opened 2016-11-08
Add a pipeline with handlers inside WebRtcConnection

jcague:
This PR adds a new way to include functionality inside Erizo. It's based on a Pipeline with Handlers that could create/read/update/delete inbound and outbound RTP Packets.

### DIFF
--- a/erizo/src/erizo/WebRtcConnection.cpp
+++ b/erizo/src/erizo/WebRtcConnection.cpp
@@ -16,6 +16,7 @@
 #include "rtp/RtpVP8Parser.h"
 #include "rtp/RtcpAggregator.h"
 #include "rtp/RtcpForwarder.h"
+#include "rtp/RtpRetransmissionHandler.h"
 
 using std::memcpy;
 
@@ -25,7 +26,8 @@ DEFINE_LOGGER(WebRtcConnection, "WebRtcConnection");
 WebRtcConnection::WebRtcConnection(const std::string& connection_id, bool audioEnabled, bool videoEnabled,
     const IceConfig& iceConfig, WebRtcConnectionEventListener* listener)
     : connection_id_(connection_id), audioEnabled_(audioEnabled), videoEnabled_(videoEnabled),
-      connEventListener_(listener), iceConfig_(iceConfig), fec_receiver_(this) {
+      connEventListener_(listener), iceConfig_(iceConfig), fec_receiver_(this),
+      inbound_pipeline_{Pipeline::create()}, outbound_pipeline_{Pipeline::create()} {
   ELOG_INFO("%s, message: constructor, stunserver: %s, stunPort: %d, minPort: %d, maxPort: %d",
       toLog(), iceConfig.stunServer.c_str(), iceConfig.stunPort, iceConfig.minPort, iceConfig.maxPort);
   bundle_ = false;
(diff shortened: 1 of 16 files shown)


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -400,11 +401,25 @@ int WebRtcConnection::deliverFeedback_(char* buf, int len) {
   return len;
 }
 
-void WebRtcConnection::onTransportData(char* buf, int len, Transport *transport) {
+void WebRtcConnection::onTransportData(std::shared_ptr<dataPacket> packet, Transport *transport) {
   if (audioSink_ == NULL && videoSink_ == NULL && fbSink_ == NULL) {
     return;
   }
 
+  packetType type = OTHER_PACKET;
```
lodoyun: I think the way this is now works but it's a little awkward. I'm pretty sure using just SSRC instead of transport type should already be enough. In any case, it's not that important for this PR and could be addressed in the future

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -400,11 +401,25 @@ int WebRtcConnection::deliverFeedback_(char* buf, int len) {
...
+
+  if (transport->mediaType == AUDIO_TYPE) {
+    type = AUDIO_PACKET;
+  } else if (transport->mediaType == VIDEO_TYPE) {
+    type = VIDEO_PACKET;
+  }
+
+  inbound_pipeline_->read(packet);
```
lodoyun: The packet here is a shared_ptr that points to a packet that will change for every packet received. Meanwhile, in the output pipeline, we have a newly allocated shared_ptr for each of the packets in the pipeline. 
As far as I can tell, this works here but will create confusion when we start implementing more pipeline handlers. I would probably make a copy to make it clearer, at the expense of an extra copy that is not really needed right now. Again, it works fine as it is as far as I can tell and it could wait until we clean up more stuff

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -16,6 +16,7 @@
 #include "rtp/RtpVP8Parser.h"
 #include "rtp/RtcpAggregator.h"
 #include "rtp/RtcpForwarder.h"
+#include "rtp/RtpRetransmissionHandler.h"
 
 using std::memcpy;
```
lodoyun: Not needed anymore

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -620,37 +631,31 @@ void WebRtcConnection::updateState(TransportState state, Transport * transport)
 }
 
 // changes the outgoing payload type for in the given data packet
-void WebRtcConnection::queueData(int comp, const char* buf, int length, Transport *transport,
-                                packetType type, uint16_t seqNum) {
+void WebRtcConnection::queueData(std::shared_ptr<dataPacket> packet) {
```
lodoyun: :+1: to this, having the information in the packet is cleared, especially sequence number information.

erizo/src/erizo/WebRtcConnection.cpp
jcague: agree, we should find how to address it in a future PR

PR #600, opened 2016-11-10
Add support to retransmit packets in Erizo

jcague:
Work in progress. I'm still adding unit tests for the new functionality, but I thought it would be interesting to create the PR already to discuss the solution.


kekkokk:
can you provide some info on this functionality? 
is active and working or it's a skeleton?

PR #602, opened 2016-11-14
Removed audio and video constraints from WebrtcConnection constructor, added hints to CreateOffer

aalonsog:


erizo/src/erizo/WebRtcConnection.h
```diff
@@ -82,7 +82,7 @@ class WebRtcConnection: public MediaSink, public MediaSource, public FeedbackSin
    */
   bool setRemoteSdp(const std::string &sdp);
 
-  bool createOffer();
+  bool createOffer(bool videoEnabled, bool audioEnabled, bool bundle);
```
jcague: I don't want to be too picky here, so this is only informational: the new coding style says we prefer variable names in Snake Case (`video_enabled`, `audio_enabled`, ...). But you can leave them as they're now because we're not changing existing variables yet.

erizoAPI/WebRtcConnection.cc
```diff
@@ -133,7 +131,11 @@ NAN_METHOD(WebRtcConnection::init) {
 NAN_METHOD(WebRtcConnection::createOffer) {
   WebRtcConnection* obj = Nan::ObjectWrap::Unwrap<WebRtcConnection>(info.Holder());
   erizo::WebRtcConnection *me = obj->me;
-  bool r = me->createOffer();
+  bool v = info[0]->BooleanValue();
```
jcague: please check also the number of arguments:

```
if (info.Length() < 3) {
  ThrowException(Exception::TypeError(v8::String::New("Wrong number of arguments")));
}
```

erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -177,7 +177,7 @@ exports.ErizoJSController = function () {
         });
         if (options.createOffer === true){
             log.debug('message: create offer requested, id:', wrtc.wrtcId);
-            wrtc.createOffer();
+            wrtc.createOffer(true, true, true);
```
jcague: please consider using this for future readability:
```
var videoEnabled = true;
var audioEnabled = true;
var bundle = true;
wrtc.createOffer(videoEnabled, audioEnabled, bundle);
```

.gitignore
```diff
@@ -15,4 +15,5 @@ erizo_controller/erizoAgent/erizo-*.log
 .project
 .cproject
 spine/erizofc.js
-extras/basic_example/public/assets/
+spine/spineClientsConfig.json
```
lodoyun: This will cause problems, as it is the default file used by `runSpineClients`, if you want to make changes for your tests you should create a new file and use `runSpineClients -s config_file.js`

erizo_controller/erizoJS/erizoJSController.js
lodoyun: :+1:

spine/nativeClient.js
```diff
@@ -120,7 +118,7 @@ exports.ErizoNativeConnection = function (spec){
                 }
             }, {});
 
-            wrtc.createOffer();
+            wrtc.createOffer(true, true, true);
```
lodoyun: 👍

PR #603, opened 2016-11-15
Add/codec configuration

lodoyun:
Moved codec configuration to rtp_media_config.js
There is a default config file that is copied to the root dir in installErizo.sh

As an added bonus, I've reworked some of the media negotiation and now it's more dynamic. Added UTs for that functionality.

**Important note**
Now everything has to be configured via the config file, including supported feedback packets and fmtp parameters. Whatever is not in the file will be rejected in the sdp negotiation.


mkhahani:
What a huge enhancement!!
Thank you. I always had to apply my modifications into the source code and re-build.
Good job :thumbsup:

PR #605, opened 2016-11-16
[FEATURE-REVIEW] Single Peer Connections

jcague:
In this PR I want to propose to change our current approach to support Single Peer Connections.

Main advantage: reduce usage of UDP ports and time to subscribe to new streams.

Please, provide as much feedback as possible to this proposal!!

### DIFF
--- a/feature-review/single-peer-connection.md
+++ b/feature-review/single-peer-connection.md
@@ -134,6 +134,56 @@ endif;
 @enduml
 )
 
+#### Coud Handler Policy
+
+There should be a new ErizoController Cloud Handler policy that takes into account these new requirements:
+
+```js
+/*
+Params
+
+roomId: string,
+agents: object with the available agents
+  agent_id : {
+    info: {
+      id: String,
+      rpc_id: String
+    },
+    metadata: Object,
+    stats: {
+      perc_cpu: Int
+    },
+    timeout: Int		// number of periods during the agent has not respond
+  }
+
+Returns
+rpc_id: agent.info.rpc_id field of the selected agent.
+
+*/
+
+var hash = function(roomId) {
+  var hash = 0, i, chr, len;
+  if (this.length === 0) return hash;
+  for (i = 0, len = roomId.length; i < len; i++) {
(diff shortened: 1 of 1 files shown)


notedit:
the two message flow  is the same.  maybe you refer the wrong image

jcague:
good catch @notedit ! I've fixed it already.

zevarito:
@jcague will this change leave Multiple Peer Connections still working and add Single Peer Connection as an option? Also, wasn't clear to me after reading your diagrams if it will provide backward compatibility or not? Thanks and good work!

jcague:
@zevarito yep, that's something that we should tackle in the document. Thanks for the heads up!

jcague:
@aalonsog @lodoyun @notedit @zevarito I updated this document with further logic / architecture / comments, can you review it again and provide feedback, please?

lodoyun:
In general I do like the direction you are proposing for this.
A couple of thoughts:

- ErizoClient should be in charge of absorbing the implementation differences in the browsers. We could gradually enable single PeerConnection in different browsers as we implement it. 

- We should define in this stage a working policy for ErizoController. For instance, by default, the policy should try to keep rooms within the same ErizoJS whenever possible and evenly distribute the rooms evenly among the available ErizoJs processes. I know there are a million ways to optimize this but it's a start.

jcague:
@lodoyun good catch! I just added a simple policy that could help us as a starting point and commented your first point in the document.

jcague:
I'd like to finish this review on Monday if I don't receive more feedback/requests. We will be able to change it in the future whenever we find issues, 

@jnoring any thoughts or feedback on this feature?

notedit:
hi  @jcague    that means   we can use  multi  streams over on peerconnection  in  client side?

notedit:
i see the  this planB  https://tools.ietf.org/html/draft-ietf-rtcweb-rtp-usage-26

notedit:
it seems that  webrtc will support  Unified Plan.

https://bugs.chromium.org/p/chromium/issues/detail?id=465349

PR #607, opened 2016-11-17
Add the possibity to include metadata in the logs

jcague:
With this change both publish() and subscribe() methods in clients will accept a new field in options called metadata. 

Metadata is an Object with keys and values, both should be strings. Below you can find an example:
```
{metadata: {type: 'publisher', external_id: 'id1'}}
```

All fields inside metadata will be printed in logs in Erizo in this PR. And we will add them to ErizoController logs too.

### DIFF
--- a/erizo/src/erizo/logger.h
+++ b/erizo/src/erizo/logger.h
@@ -39,10 +39,9 @@ class LogContext {
   void setLogContext(std::map<std::string, std::string> context) {
     context_ = context;
     context_log_ = "";
-    std::for_each(context.begin(), context.end(),
-                  [this](std::pair<std::string, std::string> item) {
+    for (const std::pair<std::string, std::string> &item : context) {
       context_log_ += item.first + ": " + item.second + ", ";
-    });
+    }
   }
 
   void copyLogContextFrom(LogContext *log_context) {
--- a/erizo/src/erizo/logger.h
+++ b/erizo/src/erizo/logger.h
@@ -25,6 +25,39 @@
 #include <log4cxx/logger.h>
 #include <log4cxx/helpers/exception.h>
 
+#include <map>
+#include <string>
+#include <utility>
+
+class LogContext {
+ public:
+  LogContext() : context_log_{""} {
+  }
+
+  virtual ~LogContext() {}
+
+  void setLogContext(std::map<std::string, std::string> context) {
+    context_ = context;
+    context_log_ = "";
+    std::for_each(context.begin(), context.end(),
+                  [this](std::pair<std::string, std::string> item) {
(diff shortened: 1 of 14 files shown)


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -41,7 +41,7 @@ exports.ErizoJSController = function () {
     /*
      * Enables/Disables slideshow mode for a subscriber
      */
-    setSlideShow = function (slideShowMode, from, to){
+    setSlideShow = function (slideShowMode, from, to) {
```
lodoyun: I think minor cosmetic problems like this in the future should be grouped in another PR, they add too much noise and create  merge conflicts when there shouldn't be.

erizo_controller/erizoClient/src/Room.js
```diff
@@ -720,7 +724,8 @@ Erizo.Room = function (spec) {
                 // 1- Subscribe to Stream
 
                 if (that.p2p) {
-                    sendSDPSocket('subscribe', {streamId: stream.getID()});
+                    sendSDPSocket('subscribe', {streamId: stream.getID(),
```
lodoyun: Does this have any effect on p2p streams as it is now?

erizo/src/erizo/logger.h
```diff
@@ -25,6 +25,38 @@
 #include <log4cxx/logger.h>
 #include <log4cxx/helpers/exception.h>
 
+#include <map>
+#include <string>
+#include <utility>
+
+class LogContext {
```
lodoyun: :+1:

erizo_controller/erizoClient/src/Room.js
jcague: nope, but it will have if we add logs to ErizoController in the short future (about an hour 😄 )

lodoyun: 👏

extras/basic_example/public/script.js
```diff
@@ -93,15 +93,15 @@ window.onload = function () {
...
+            room.subscribe(stream, {slideShowMode: slideShowMode, metadata: {type: 'subscriber'}});
             stream.addEventListener('bandwidth-alert', cb);
           }
         }
       };
 
       room.addEventListener('room-connected', function (roomEvent) {
 
-        room.publish(localStream, {maxVideoBW: 300});
+        room.publish(localStream, {maxVideoBW: 300, metadata: {type: 'publisher'}});
```
aalonsog: Ok, with this I understand why metadata is desired :)

jcague:
@lodoyun completely agree. I updated the description with the changes in the public API.

PR #611, opened 2016-11-17
npm packages needed by BasicExample

wolmi:


jcague:
@wolmi if don't use those packages afaik. That's probably something you added in your local repo.

wolmi:
The packages are used in licode/extras/basic_example/basicServer.js

https://github.com/ging/licode/blob/2cf67accf21472fe83d72e95c9bb947bc03d0018/extras/basic_example/basicServer.js#L5-L6

Now I see the packages are installed in installBasicExample.sh instead of putting a new packages.json for basicExample server.

I think this is not the correct way, because the dependencies in packages.json are for development same as the basic example server, I think is logic put it all together in packages.json equal as express or body-parser.

If some of dependencies are used in other server like erizo or nuve this may be put in dependencies instead of devDependencies.

Sorry but I want to understand this.

Thanks

jcague:
@wolmi they're only used by basicServer, so I don't think we should need them in the general package.json. Instead it might be interesting to have a different package.json for basicServer. Can you do it?

PR #621, opened 2016-11-23
Added status log when closing WebRTCConnections

lodoyun:
We should now always have an status: 105 message when connections are closed.
Regarding the implementation, I'm beginning to think we should wrap the addon.WebRTCConnection object in JS and explicitly add the context to it. Init and close should be functions of that object. What we have now is a little awkward.

### DIFF
--- a/erizo_controller/erizoJS/erizoJSController.js
+++ b/erizo_controller/erizoJS/erizoJSController.js
@@ -20,6 +20,7 @@ exports.ErizoJSController = function () {
         SLIDESHOW_TIME = 1000,
         PLIS_TO_RECOVER = 3,
         initWebRtcConnection,
+        closeWebRtcConnection,
         getSdp,
         getRoap,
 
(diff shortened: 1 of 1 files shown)


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -141,6 +143,14 @@ exports.ErizoJSController = function () {
         callback('callback', {type: 'initializing'});
     };
 
+    closeWebRtcConnection = function (wrtc) {
+        var associatedMetadata = wrtc.metadata || {};
+        wrtc.close();
+        log.info('message: WebRtcConnection status update, ' +
```
aalonsog: Just a doubt: should we wait to a callback form wrtc confirming that it has been successfully closed?

lodoyun: In an ideal world that would be the case, and that was what I was aiming for. 
However, it creates race conditions that we want to avoid. In general, you can assume that the connection is no longer useful or active with this log, and you can diagnose the complete destruction of the objects in lower level logs.

jcague:
@lodoyun with this PR it will be a lot easier to track status, nice work!

PR #624, opened 2016-11-24
Add a ThreadPool for WebRtcConnections

jcague:
This PR sets a ThreadPool that will be used to get Workers and assign them to WebRtcConnections. These workers will be threads processing RTP packets, but they won't do IO tasks.

I've added a new parameter in the configuration called numWorkers, that is the total number of threads that will be used by Erizo to process RTP packets.

### DIFF
--- a/erizo/src/erizo/WebRtcConnection.h
+++ b/erizo/src/erizo/WebRtcConnection.h
@@ -64,7 +64,7 @@ class WebRtcConnection: public MediaSink, public MediaSource, public FeedbackSin
    * Constructor.
    * Constructs an empty WebRTCConnection without any configuration.
    */
-  WebRtcConnection(const std::string& connection_id, const IceConfig& iceConfig,
+  WebRtcConnection(std::shared_ptr<Worker> worker, const std::string& connection_id, const IceConfig& iceConfig,
       const std::vector<RtpMap> rtp_mappings, WebRtcConnectionEventListener* listener);
   /**
    * Destructor.
--- a/erizo/src/erizo/thread/ThreadPool.h
+++ b/erizo/src/erizo/thread/ThreadPool.h
@@ -10,37 +10,12 @@ namespace erizo {
 
 class ThreadPool {
  public:
-  explicit ThreadPool(unsigned int num_workers) : workers_{} {
-    for (unsigned int index = 0; index < num_workers; index++) {
-      workers_.push_back(std::make_shared<Worker>());
-    }
-  }
+  explicit ThreadPool(unsigned int num_workers);
+  ~ThreadPool();
 
-  ~ThreadPool() {
-    close();
-  }
-
-  std::shared_ptr<Worker> getLessUsedWorker() {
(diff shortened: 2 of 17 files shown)


erizo_controller/erizoJS/erizoJS.js
```diff
@@ -73,7 +75,10 @@ var controller = require('./erizoJSController');
 // Logger
 var log = logger.getLogger('ErizoJS');
 
-var ejsController = controller.ErizoJSController();
+var threadPool = new addon.ThreadPool(10);
```
aalonsog: If I'm right, you should use here the number of workers configured in config file

jcague: :+1: you got it

erizo/src/erizo/thread/ThreadPool.h
```diff
@@ -0,0 +1,50 @@
...
+      workers_.push_back(std::make_shared<Worker>());
+    }
+  }
+
+  ~ThreadPool() {
+    close();
+  }
+
+  std::shared_ptr<Worker> getLessUsedWorker() {
+    std::shared_ptr<Worker> choosed_worker = workers_.front();
```
lodoyun: consider using `chosen_worker`

erizo/src/erizo/thread/ThreadPool.h
```diff
@@ -0,0 +1,50 @@
...
+#define ERIZO_SRC_ERIZO_THREAD_THREADPOOL_H_
+
+#include <memory>
+#include <vector>
+
+#include "thread/Worker.h"
+
+namespace erizo {
+
+class ThreadPool {
```
lodoyun: I generally don't like having the full implementation in the header file. I know workers are already that way...
In this case I guess ThreadPool is simple enough that we don't need a cc file but I wanted to make sure this decision was taken consciously. Maybe we should discuss where we want to draw the line because I would prefer this not becoming a trend.

erizo/src/erizo/WebRtcConnection.h
```diff
@@ -64,7 +64,7 @@ class WebRtcConnection: public MediaSink, public MediaSource, public FeedbackSin
    * Constructor.
    * Constructs an empty WebRTCConnection without any configuration.
    */
-  WebRtcConnection(const std::string& connection_id, const IceConfig& iceConfig,
+  WebRtcConnection(std::shared_ptr<Worker> worker, const std::string& connection_id, const IceConfig& iceConfig,
```
lodoyun: This kills spine as it uses erizoAPI directly

scripts/licode_default.js
```diff
@@ -140,6 +140,9 @@ config.erizo = {};
 //Erizo Logs are piped through erizoAgent by default
 //you can control log levels in [licode_path]/erizo_controller/erizoAgent/log4cxx.properties
 
+// Number of workers that will be used to handle WebRtcConnections
+config.erizo.numWorkers = 24;
```
lodoyun: 👍

PR #628, opened 2016-11-29
Added a Scheduler to run tasks at a given time

jcague:
This PR adds a new Scheduler to Workers, to be able to schedule tasks that we want to execute in the future. It's interesting to create periodic tasks, timeouts, etc.

### DIFF
--- a/erizo/src/erizo/thread/Scheduler.cpp
+++ b/erizo/src/erizo/thread/Scheduler.cpp
@@ -7,8 +7,10 @@
 
 
 Scheduler::Scheduler(int n_threads_servicing_queue)
-: n_threads_servicing_queue_(0), stop_requested_(false), stop_when_empty_(false) {
-  for (unsigned int index = 0; index < n_threads_servicing_queue; index++) {
+: n_threads_servicing_queue_(n_threads_servicing_queue), stop_requested_(false), stop_when_empty_(false) {
+  stop_requested_ = false;
+  stop_when_empty_ = false;
+  for (int index = 0; index < n_threads_servicing_queue; index++) {
     group_.create_thread(boost::bind(&Scheduler::serviceQueue, this));
   }
 }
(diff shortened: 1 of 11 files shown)


erizo/src/erizo/thread/Scheduler.cpp
```diff
@@ -0,0 +1,90 @@
+#include "thread/Scheduler.h"
+
+#include <assert.h>
+
+#include <boost/bind.hpp>
+#include <utility>
+
+
+Scheduler::Scheduler(int n_threads_servicing_queue)
+: n_threads_servicing_queue_(0), stop_requested_(false), stop_when_empty_(false) {
```
lodoyun: shouldn't this be `Scheduler::Scheduler(int n_threads_servicing_queue)
 +: n_threads_servicing_queue_(n_threads_servicing_queue), stop_requested_(false), stop_when_empty_(false)`

erizo/src/erizo/thread/Scheduler.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+  new_task_scheduled_.notify_all();
+  group_.join_all();
+}
+
+void Scheduler::schedule(Scheduler::Function f, std::chrono::system_clock::time_point t) {
+  {
+    std::unique_lock<std::mutex> lock(new_task_mutex_);
+    task_queue_.insert(std::make_pair(t, f));
```
lodoyun: Considering including a comment or a reminder that the tasks are ordered by time_point in the multimap, to make it easier to contribute and avoid confusions

erizo/src/erizo/thread/Scheduler.cpp
jcague: well, internally we increase n_threads_servicing_queue_ from each thread. Either way is good IMO so if you feel strong about setting it in the constructor I can't change the code.

erizo/src/erizo/thread/Scheduler.cpp
jcague: done

erizo/src/erizo/thread/Scheduler.cpp
jcague: I finally apply your comment to avoid potential race conditions while running unit tests.

jcague:
@lodoyun I've added  Unit Tests for the Scheduler.

PR #630, opened 2016-11-29
Including DTLS roles for active-passive decision

aalonsog:


aalonsog:
For the moment passive answerer peers are not supported. When you reset the DtlsTransport to switch server to client (passive to active) ICE fails. I will try to improve this in the future.

erizo/src/erizo/SdpInfo.cpp
```diff
@@ -581,6 +606,20 @@ namespace erizo {
         isFingerprint = true;
         ELOG_DEBUG("Fingerprint %s ", fingerprint.c_str());
       }
+      if (isSetup != std::string::npos) {
+        std::vector<std::string> parts;
+        parts = stringutil::splitOneOf(line, ":", 1);
+        if (parts[1].compare("passive ") == 0) {
```
jcague: why do we need a blank space here?

consider something like `if (parts[1].find("passive") != std::string::npos) {` to avoid issues with spaces.

std::size_t found = str.find(str2);
  if (found!=std::string::npos)

PR #633, opened 2016-12-01
Change libnice to v 0.1.7

lodoyun:
libnice 0.1.13 was causing connection problems in some cases - We'll downgrade to 0.1.7 as it has proven to work without problems until we further investigate the issue

### DIFF
--- a/scripts/installMacDeps.sh
+++ b/scripts/installMacDeps.sh
@@ -97,9 +97,9 @@ install_openssl(){
 install_libnice(){
   if [ -d $LIB_DIR ]; then
     cd $LIB_DIR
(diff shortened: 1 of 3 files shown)


thehunmonkgroup:
Just a note that I also struggled finding a version of libnice that behaved well. The older versions were missing some important bugfixes, and the 0.1.13 release had the connectivity issues you observed.

I eventually settled on using their master branch as of June 23rd, 2016 (commit 58fdd9b5f7e04e5f77bbc50ba2ac24f592a10b84) and this has worked without issue for me.

I've also inquired about why they haven't made a new release in forever, and weren't interested in cutting a new point release at the time.

PR #635, opened 2016-12-02
Enable Node 6 (LTS) installation as dependency

jcague:


jcague:
good catch @kekkokk, thanks! actually we're installing `curl` just a couple of lines below! :joy: I'll fix it.

aalonsog:
Great!

PR #644, opened 2016-12-07
Add/mute stream

lodoyun:
Added the ability to mute a stream in Licode.
Calling `muteAudio` in a *remote stream* will instruct licode to stop sending audio to that client.

### DIFF
--- a/erizo/src/test/rtp/RtpAudioMuteHandlerTest.cpp
+++ b/erizo/src/test/rtp/RtpAudioMuteHandlerTest.cpp
@@ -149,16 +149,73 @@ MATCHER_P(ReceiverReportHasSequenceNumber, seq_num, "") {
 }
 
 TEST_F(RtpAudioMuteHandlerTest, basicBehaviourShouldReadPackets) {
-    auto packet = createDataPacket(kArbitrarySeqNumber, VIDEO_PACKET);
+    auto packet = createDataPacket(kArbitrarySeqNumber, AUDIO_PACKET);
 
     EXPECT_CALL(*reader.get(), read(_, _)).With(Args<1>(HasSequenceNumber(kArbitrarySeqNumber))).Times(1);
     pipeline->read(packet);
 }
 
 TEST_F(RtpAudioMuteHandlerTest, basicBehaviourShouldWritePackets) {
-    auto packet = createDataPacket(kArbitrarySeqNumber, VIDEO_PACKET);
+    auto packet = createDataPacket(kArbitrarySeqNumber, AUDIO_PACKET);
 
     EXPECT_CALL(*writer.get(), write(_, _)).With(Args<1>(HasSequenceNumber(kArbitrarySeqNumber))).Times(1);
     pipeline->write(packet);
 }
 
+TEST_F(RtpAudioMuteHandlerTest, shouldNotWriteAudioPacketsIfActive) {
+    auto audio_packet = createDataPacket(kArbitrarySeqNumber, AUDIO_PACKET);
+    auto video_packet = createDataPacket(kArbitrarySeqNumber+1, VIDEO_PACKET);
+    audio_mute_handler->muteAudio(true);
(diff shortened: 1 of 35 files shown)


erizo_controller/erizoClient/src/Stream.js
```diff
@@ -280,25 +280,41 @@ Erizo.Stream = function (spec) {
         }
     };
 
+    that.muteAudio = function (audioActive, callback) {
```
jcague: having muteAudio with audioActive seems like a bit complex to understand for the purpose of this feature IMO. Consider other options like:

that.muteAudio = function(audioMuted,....

or

that.setAudioActive = function(audioActive,....

or 

that.muteAudio = function(...
that.unmuteAudio = function(...

erizo/src/test/rtp/RtpAudioMuteHandlerTest.cpp
```diff
@@ -0,0 +1,221 @@
...
+    if (type == AUDIO_PACKET) {
+      header->setSSRC(kAudioSsrc);
+    } else {
+      header->setSSRC(kVideoSsrc);
+    }
+
+    return std::make_shared<dataPacket>(0, reinterpret_cast<char*>(header), sizeof(erizo::RtpHeader), type);
+  }
+
+  std::shared_ptr<dataPacket> createNack(uint16_t seq_number, packetType type, int additional_packets = 0) {
```
jcague: maybe this is the right PR to create a common helper for all these methods (createData, createNack, Matchers, etc.) and to share them with other UTs. Feel free to add it here, or we can do it in a separate PR if you prefer.

erizo_controller/erizoClient/src/Stream.js
lodoyun: You're right, thats a leftover from a previous name, will fix.

erizo_controller/erizoClient/src/Stream.js
```diff
@@ -280,25 +280,41 @@ Erizo.Stream = function (spec) {
...
+        if (that.room && that.room.p2p){
+            L.Logger.warning('muteAudio is not implemented in p2p streams');
+            callback ('error');
```
aalonsog: Maybe we can send a message to the stream publisher to ask it to close the audio track. This way the feature will work in both scenarios.

erizo_controller/erizoClient/src/Stream.js
aalonsog: As methods in erizo are named "muteStream" I guess we are going to include the possibility of muting video in the future. Why we don't use this nomenclature also in the client (stream.mute(audioActive, videoActive))?

erizo/src/test/rtp/RtpAudioMuteHandlerTest.cpp
lodoyun: I think I'll do it in a separate PR for clarity :+1:

erizo_controller/erizoClient/src/Stream.js
lodoyun: It may take a while before we have video mute (if we ever implement it). I wanted to avoid the confusion in ErizoClient API while leaving the rest of the path flexible enough so we don't have to change many things in the future.
I still prefer muteAudio but I could be convinced to change it :)  @jcague, any opinion on this?

jcague: I'd say `muteAudio()/muteVideo()` is easier to understand and to develop than `mute(audio,video)` because it avoids developers to keep state of both in the same call. But I don't have a strong opinion though.

erizo_controller/erizoClient/src/Stream.js
lodoyun: I think the cost/benefit of implementing something like this in P2P is a lot less appealing. The big win here is in big rooms, as we're saving audio bandwidth. In small rooms/p2p muting the audio in the client side should be sufficient in most cases. I'm not even convinced it's that practical considering the potential delay.

An alternative would be to just to mute the audio track here if we're in p2p.

PR #645, opened 2016-12-07
Clean files that are not used by DTLS anymore

jcague:
It's one of those PR where I only remove old things....


jcague:
you're right. I removed OpenSSLInit also because we weren't using it.

PR #648, opened 2016-12-09
Use monotonic clock

jcague:


erizo/src/erizo/lib/Clock.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include <chrono>  // NOLINT
+
+namespace erizo {
+
+using clock = std::chrono::steady_clock;
+using time_point = std::chrono::steady_clock::time_point;
+using duration = std::chrono::steady_clock::duration;
+
+inline int64_t durationToMs(erizo::duration duration) {
```
lodoyun: Consider putting this utility functions inside a class, even if they are still static. For instance, `ClockUtils::durationToMs(...)` would be more readable in my opinion

erizo/src/erizo/media/ExternalOutput.h
```diff
@@ -82,7 +83,7 @@ class ExternalOutput : public MediaSink, public RawDataReceiver, public Feedback
   // and that value is added to every timestamp we write.
   long long firstVideoTimestamp_;  // NOLINT
   long long firstAudioTimestamp_;  // NOLINT
-  long long firstDataReceived_;  // NOLINT
+  clock::time_point firstDataReceived_;  // NOLINT
```
lodoyun: This was a good opportunity to ditch CamelCase in the variables involved in this PR.

PR #650, opened 2016-12-09
[WIP] Enable build in Travis OSX and disable build for Node 0.10

jcague:
Work in progress.

In this PR I want to investigate builds in OSX again now that it takes short time, and I'll disable builds for Node 0.10.


jcague:
We won't support OSX anymore 😞

PR #652, opened 2016-12-12
Pass stunServerUrl and turnServer configurations to Clients

alireza-m:


aalonsog:
Why do we need to include separated parameters for stun and turn? Currently you can use iceServers array to include several servers of both types. See the format [here](https://github.com/ging/licode/blob/master/scripts/licode_default.js#L39)

aalonsog:
Any new about this?

PR #656, opened 2016-12-14
Adds Bandwidth Estimation in Erizo

jcague:


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -141,6 +143,8 @@ bool WebRtcConnection::setRemoteSdp(const std::string &sdp) {
   localSdp_.setOfferSdp(remoteSdp_);
   extProcessor_.setSdpInfo(localSdp_);
 
+  bwe_handler_->updateExtensionMaps(extProcessor_.getVideoExtensionMap(), extProcessor_.getAudioExtensionMap());
```
lodoyun: 👍 to using the extensionProcessor :)

erizo/src/test/rtp/BandwidthEstimationHandlerTest.cpp
```diff
@@ -0,0 +1,224 @@
...
+    pipeline->addBack(writer);
+    pipeline->addBack(bwe_handler);
+    pipeline->addBack(reader);
+    pipeline->finalize();
+  }
+
+  virtual void TearDown() {
+  }
+
+  std::shared_ptr<dataPacket> createDataPacket(uint16_t seq_number, packetType type) {
```
lodoyun: We really should do a helper for this at some point as all filters need this.

kekkokk:
Hi guys, Is this an rtcp termination implementation? Could you be more specific in "important" commits like these to explain what's the purpose of the commit? in this way we could be more helpful in our tests and feedbacks

jcague:
@kekkokk sure! I'd like to propose templates to all the PRs just to make sure we add more information in all PR (important or not). I'll try to add those templates by today EOD.

wrt this PR, it's not a RTCP Termination per se, it just starts sending REMB packets with the estimated bandwidth from Erizo to the Publishers. It will be helpful for instance when there's no subscriber (that's the case of streams being recorded), or when the subscribers are not sending proper feedback back to the publisher (that's the case of the slide show mode).

kekkokk:
I saw, I'm trying your implementation right now cause it's like christmas for me today. I waited this feature for months cause I need to record streams when the publisher is alone. 
Until now I was forced to auto-subscribe my localstream and it was an "awful" method.
It seems that the bandwidth estimation works flawlessly when alone, it's not capped a 300kb anymore and it can reach the full bandwidth assigned in the maxVideoBW.

However, the recordings are full of artifacts anyway, like when it was capped @ 300kb.
I think another type of feedback is missing. 
Are you aware of this behaviour? 
Can you clarify to me how much work and what is still missing?

Trying to document myself

jcague:
The most reasonable cause for your problem is that we aren't handling packet losses in the External Output, so we should add such functionality inside. The idea would be to send NACKs from EO whenever we detect a packet (or group of packets) has been lost.

jcague:
@kekkokk I don't know if you did it yet, but can you create an issue explaining your problem. I estimate that we'll start working on it in the next weeks. Thanks for the heads up!

kekkokk:
thank you very much for your time, I really appreciate your work.
I'll open an issue asap, 
have a nice day!

erizo/src/test/rtp/BandwidthEstimationHandlerTest.cpp
jcague: Agree, let's talk about it this week

PR #662, opened 2016-12-19
Add a new Synthetic Input for future stress tests

jcague:
This PR adds a new Synthetic Input component that is capable of creating fake VP8 and Opus packets without much CPU consumption. We want to use it in future stress tests. 

Note: The fake packets cannot be transcoded by current browsers at this point but they'll compute in stats as real packets and all the logic inside Erizo will handle them as real packets, so they'll be forwarded to subscribers too.

- [x] SyntheticInput in Erizo
- [x] SyntheticInput Unit Tests
- [x] SyntheticInput in ErizoAPI
- [x] SyntheticInput in Spine
- [x] Example of usage of SyntheticInput from Spine


erizo/src/erizo/media/SyntheticInput.cpp
```diff
@@ -0,0 +1,223 @@
...
+  }
+  delete header;
+}
+
+void SyntheticInput::close() {
+  running_ = false;
+}
+
+void SyntheticInput::calculateSizeAndPeriod(uint32_t video_bitrate, uint32_t audio_bitrate) {
+  ELOG_TRACE("Bitrate received: %d", video_bitrate);
```
lodoyun: First time we've used TRACE, wouldn't DEBUG be good enough?

erizo/src/erizo/media/SyntheticInput.cpp
```diff
@@ -0,0 +1,223 @@
...
+          switch (chead->getBlockCount()) {
+            case RTCP_PLI_FMT:
+            case RTCP_FIR_FMT:
+              sendPLI();
+              break;
+            case RTCP_AFB:
+              char *unique_id = reinterpret_cast<char*>(&chead->report.rembPacket.uniqueid);
+              if (!strncmp(unique_id, "REMB", 4)) {
+                uint64_t bitrate = chead->getBrMantis() << chead->getBrExp();
+                calculateSizeAndPeriod(bitrate, kDefaultAudioBitrate);
```
lodoyun: :+1:

erizo/src/test/media/SyntheticInputTest.cpp
```diff
@@ -0,0 +1,283 @@
...
+    input->setVideoSink(&sink);
+    input->setAudioSink(&sink);
+    input->start();
+  }
+
+  virtual void TearDown() {
+    input->close();
+  }
+
+  void executeTasksInNextMs(int time) {
```
lodoyun: 👏

erizo/src/erizo/media/SyntheticInput.cpp
jcague: Actually I'll remove it since it's not needed anymore. Just used it to check that everything worked as expected during my manual tests.

PR #664, opened 2016-12-19
Fix External Input and External Output destruction

jcague:


jcague:
@kekkokk this should fix your problem when stopping recordings.

kekkokk:
@jcague unfortunately no, after the first stop (which is slightly delayed too), there is still a considerable delay.

kekkokk:
I confirm, tried with a fresh install. same issue

jcague:
@kekkokk I did more changes and tested them successfully in my machine, can you try it again please?

kekkokk:
works like a charm, thank you!

PR #665, opened 2016-12-20
Saving ErizoController-Room relations in the database

aalonsog:
Now relations between rooms and the erizoController where it is assigned are stored in the nuve database. This way, crashes in Nuve avoids service interruption.

### DIFF
--- a/nuve/nuveAPI/cloudHandler.js
+++ b/nuve/nuveAPI/cloudHandler.js
@@ -67,8 +67,26 @@ var getEcQueue = function (callback) {
     });
 };
 
+var assignErizoController = function (erizoControllerId, room, callback) {
+    room.erizoControllerId = erizoControllerId;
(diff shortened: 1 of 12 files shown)


nuve/nuveAPI/cloudHandler.js
```diff
@@ -151,94 +165,125 @@ exports.addNewErizoController = function (msg, callback) {
...
+                    roomRegistry.updateRoom(room._id, room);
+                    getErizoControllerForRoom(room, callback);
+                }
+            });
+            return;
         }
 
-        if (id !== undefined) {
+        var attempts = 0;
+        var intervarId;
```
lodoyun: Maybe you meant intervalId? 😄

aalonsog: ok ok... it seems it [doesn't exist](http://www.wordreference.com/definition/intervar) in English. Fixed!

PR #670, opened 2016-12-23
Fixes when checking users from Cloud Handler to Erizo Controller

aalonsog:
When Cloud Handler receives a timeout when checking users in a room, the response is now correctly handled in Nuve Client.

This fixes a break in basic example when removing old rooms.

I've also updated status responses in nuve to new express notation.

### DIFF
--- a/extras/basic_example/basicServer.js
+++ b/extras/basic_example/basicServer.js
@@ -83,20 +83,28 @@ var deleteRoomsIfEmpty = function (theRooms, callback) {
         callback(true);
         return;
     }
-    var theRoom = theRooms.pop();
-    N.API.getUsers(theRoom._id, function(userlist) {
+    var theRoomId = theRooms.pop()._id;
+    N.API.getUsers(theRoomId, function(userlist) {
         var users = JSON.parse(userlist);
         if (Object.keys(users).length === 0){
-            N.API.deleteRoom(theRoom._id, function(){
+            N.API.deleteRoom(theRoomId, function(){
                 deleteRoomsIfEmpty(theRooms, callback);
             });
         } else {
             deleteRoomsIfEmpty(theRooms, callback);
         }
-    }, function () {
-        N.API.deleteRoom(theRoom._id, function(){
-            deleteRoomsIfEmpty(theRooms, callback);
-        });
+    }, function (error, status) {
+        console.log('Error getting user list for room ', theRoomId, 'reason: ', error);
+        switch (status) {
+            case 404:
+                deleteRoomsIfEmpty(theRooms, callback);
+                break;
+            case 503:
+                N.API.deleteRoom(theRoomId, function(){
+                    deleteRoomsIfEmpty(theRooms, callback);
+                });
+                break;
+        }
     });
 };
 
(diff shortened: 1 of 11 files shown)


nuve/nuveAPI/resource/usersResource.js
```diff
@@ -27,19 +27,19 @@ exports.getList = function (req, res) {
...
             log.info('message: getUserList - room not found, roomId: ' + req.params.room);
-            res.send('Room does not exist', 404);
+            res.status(404).send('Room does not exist');
             return;
         }
 
         log.info('message: getUsersList success, roomId: ' + currentRoom._id +
                  ', serviceId: ' + currentService._id);
         cloudHandler.getUsersInRoom(currentRoom._id, function (users) {
-            if (users === 'error') {
```
jcague: I understand that we're not handling error responses anymore, could this be a problem?

extras/basic_example/basicServer.js
```diff
@@ -84,19 +84,19 @@ var deleteRoomsIfEmpty = function (theRooms, callback) {
...
-                }
+                deleteRoomsIfEmpty(theRooms, callback);
             });
         } else {
-            if (theRooms.length > 0){
-                deleteRoomsIfEmpty(theRooms, callback);
-            }
+            deleteRoomsIfEmpty(theRooms, callback);
         }
+    }, function () {
```
lodoyun: The error callback may very well be caused by a timeout, we either don't handle it or we should take into account what kind of errors it will report.
Consider the possible errors, if the room does not exist then deleting it will also fail, if it's a timeout (nuve is not available) then it also won't work. I don't see the value in the callback as it is.

extras/basic_example/basicServer.js
```diff
@@ -84,19 +84,19 @@ var deleteRoomsIfEmpty = function (theRooms, callback) {
         return;
     }
     var theRoom = theRooms.pop();
-    N.API.getUsers(theRoom._id, function(userlist){
+    N.API.getUsers(theRoom._id, function(userlist) {
         var users = JSON.parse(userlist);
         if (Object.keys(users).length === 0){
             N.API.deleteRoom(theRoom._id, function(){
-                if (theRooms.length > 0){
```
lodoyun: 👍

nuve/nuveAPI/cloudHandler.js
```diff
@@ -252,9 +251,6 @@ exports.getUsersInRoom = function (roomId, callback) {
         }
         var rpcID = 'erizoController_' + room.erizoControllerId;
         rpc.callRpc(rpcID, 'getUsersInRoom', [roomId], {'callback': function (users) {
-            if (users === 'timeout') {
```
lodoyun: 👍

nuve/nuveAPI/resource/usersResource.js
aalonsog: The response in "users" can only be the list of users (an array) or the string "timeout" in case there is a problem in the call to EC. 

So I think at this point it is enough with handling this kind of error

extras/basic_example/basicServer.js
aalonsog: Thanks for the comment!! Now I take into account the two possible errors Nuve could return (404 and 503). I've also modified Nuve Client to return error code in the response.

nuve/nuveAPI/resource/usersResource.js
jcague: :+1:

PR #681, opened 2017-01-03
Enable client API to mute audio on publishers

jcague:
This PR aims to:

- [x]  Reduce locks in audio mute feature (Erizo)
- [x]  Create a single list of publishers in ErizoJS
- [x]  Add the possibility to mute audio from a publisher

With this PR we can mute audio from Publisher and it will stop sending bytes to all the subscribers, saving bandwidth in these cases.


jcague:
@lodoyun @aalonsog I merge this PR now, but please review it once you come back. I'll apply changes in a separate PR.

PR #693, opened 2017-01-12
Enable Subscribe only option when loading basic example

jcague:


extras/basic_example/public/script.js
```diff
@@ -100,8 +100,8 @@ window.onload = function () {
       };
 
       room.addEventListener('room-connected', function (roomEvent) {
-
-        room.publish(localStream, {maxVideoBW: 300, metadata: {type: 'publisher'}});
+        var onlySubscribe = getParameterByName('onlySubscribe') == '1' || false;
```
lodoyun: Consider it making it like "screen" where it being present is enough. Either change the screen to having to be '1' or change this one.

PR #697, opened 2017-01-16
Handlers con now be disabled by configuration

jcague:
With this PR we'll be able to disable handlers (and functionality) by default in Erizo. 

Each handler will now use a unique name to be identified and we can disable them with a line like this one in licode_config.js :
```js 
config.erizo['disabled_handlers'] = ['bwe', 'retransmissions'];
```

### DIFF
--- a/erizo_controller/erizoClient/src/Stream.js
+++ b/erizo_controller/erizoClient/src/Stream.js
@@ -7,7 +7,7 @@
 var Erizo = Erizo || {};
 Erizo.Stream = function (spec) {
     var that = Erizo.EventDispatcher(spec),
-        getFrame;
+        getFrame, controlHandler;
 
     that.stream = spec.stream;
     that.url = spec.url;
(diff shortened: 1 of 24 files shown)


erizo_controller/erizoClient/src/Stream.js
```diff
@@ -291,6 +291,24 @@ Erizo.Stream = function (spec) {
         that.pc.updateSpec(config, callback);
     };
 
+    controlHandler = function (handlers, publisherSide, enable) {
+      publisherSide = !(publisherSide !== true);
+      var handlers = (typeof handlers === 'string') ? [handlers] : handlers;
+      handlers = (handlers instanceof Array) ? handlers : [];
+
+      if (handlers.length > 0) {
```
lodoyun: We should, at some point, agree to a unified error reporting for publicly exposed API functions.

PR #698, opened 2017-01-17
Added client-side API to access stats from Erizo

lodoyun:
ErizoClient can be used to access real-time stats gathered directly by Erizo. They contain mostly from RTCP packets from the clients but also a couple of Erizo-calculated values: actual bitrates and the output from the Erizo bitrate estimator.

The API may change but for now they are accessed with:
`room.getStreamStats(streamID, callback)` where the callback takes one parameter, the actual stats data in JSON format.

Also, this PR introduces a new _permission_:`STATS` that can be used to limit which roles can access the stats.

### DIFF
--- a/erizo/src/erizo/Stats.cpp
+++ b/erizo/src/erizo/Stats.cpp
@@ -25,6 +25,7 @@ namespace erizo {
   }
 
   uint32_t Stats::processRtpPacket(char* buf, int len) {
+    boost::recursive_mutex::scoped_lock lock(mapMutex_);
     RtcpHeader *chead = reinterpret_cast<RtcpHeader*> (buf);
     if (chead->isRtcp())
       return 0;
@@ -34,7 +35,7 @@ namespace erizo {
     if (bitrate_bytes_map.find(ssrc) == bitrate_bytes_map.end()) {
       bitrate_bytes_map[ssrc] = 0;
     }
-    bitrate_bytes_map[ssrc]+=len;
+    bitrate_bytes_map[ssrc] += len;
     time_point nowms = clock::now();
     time_point start = bitrate_calculation_start_;
     duration delay = nowms - start;
--- a/erizo/src/erizo/Stats.cpp
+++ b/erizo/src/erizo/Stats.cpp
@@ -26,15 +26,29 @@ namespace erizo {
   }
 
   uint32_t Stats::processRtpPacket(char* buf, int len) {
-    rtpBytesReceived_+=len;
+    RtcpHeader *chead = reinterpret_cast<RtcpHeader*> (buf);
+    if (chead->isRtcp())
+      return 0;
+
+    RtpHeader* head = reinterpret_cast<RtpHeader*>(buf);
+    uint32_t ssrc = head->getSSRC();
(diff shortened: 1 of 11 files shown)


erizo/src/erizo/Stats.cpp
```diff
@@ -26,15 +25,29 @@ namespace erizo {
   }
 
   uint32_t Stats::processRtpPacket(char* buf, int len) {
-    rtpBytesReceived_+=len;
+    RtcpHeader *chead = reinterpret_cast<RtcpHeader*> (buf);
```
jcague: I think we need to protect some of this code with a mutex. And also check that we're protecting getStats.

erizo/src/erizo/rtp/BandwidthEstimationHandler.cpp
```diff
@@ -216,4 +216,8 @@ void BandwidthEstimationHandler::OnReceiveBitrateChanged(const std::vector<uint3
   last_send_bitrate_ = bitrate_;
   sendREMBPacket();
 }
+
+uint32_t BandwidthEstimationHandler::getLastSendBitrate() {
+  return last_send_bitrate_;
```
jcague: we need to make last_send_bitrate_ atomic I think.

erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -485,5 +485,55 @@ exports.ErizoJSController = function (threadPool) {
...
+                var unfilteredStats = JSON.parse(publisher.subscribers[sub].getStats());
+                for (var channel in unfilteredStats) {
+                    var ssrc = unfilteredStats[channel].sourceSsrc;
+                    var isReceivedBitrate = false;
+                    if (ssrc === undefined) {
+                        ssrc = unfilteredStats[channel].ssrc;
+                        isReceivedBitrate = true;
+                    }
+                    if (!stats[sub][ssrc]) {
+                        stats[sub][ssrc] = {};
```
jcague: stats[sub][ssrc] =  stats[sub][ssrc] || {};

erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -485,5 +485,55 @@ exports.ErizoJSController = function (threadPool) {
...
+                    }
+                    if (isReceivedBitrate) {
```
jcague: you can use Object.assign() to merge objects if you prefer.

erizo/src/erizo/Stats.cpp
```diff
@@ -26,15 +25,29 @@ namespace erizo {
...
+    if (chead->isRtcp())
+      return 0;
+
+    RtpHeader* head = reinterpret_cast<RtpHeader*>(buf);
+    uint32_t ssrc = head->getSSRC();
+    if (bitrate_bytes_map.find(ssrc) == bitrate_bytes_map.end()) {
+      bitrate_bytes_map[ssrc] = 0;
+    }
+    bitrate_bytes_map[ssrc]+=len;
```
jcague: prefer blank spaces around ` =+ `: bitrate_bytes_map[ssrc] += len;

PR #705, opened 2017-01-23
RR Generation Handler

kekkokk:
Here's my implementation of RR Generation handler.
Tried to follow all your suggestion (https://github.com/ging/licode/pull/687#discussion_r96769698).
Let me know what do you think

### DIFF
--- a/erizo/src/erizo/rtp/RRGenerationHandler.cpp
+++ b/erizo/src/erizo/rtp/RRGenerationHandler.cpp
@@ -10,7 +10,7 @@ namespace erizo {
 DEFINE_LOGGER(RRGenerationHandler, "rtp.RRGenerationHandler");
 
 RRGenerationHandler::RRGenerationHandler(WebRtcConnection *connection) :
-    temp_ctx_{nullptr}, enabled_{true} {}
+    temp_ctx_{nullptr}, enabled_{true}, rientra(false) {}
 
 
 void RRGenerationHandler::enable() {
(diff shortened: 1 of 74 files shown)


erizo/src/erizo/rtp/RRGenerationHandler.cpp
```diff
@@ -0,0 +1,251 @@
...
+  enabled_ = false;
+}
+
+bool RRGenerationHandler::isRetransmitOfOldPacket(std::shared_ptr<dataPacket> packet) {
+  RtpHeader *head = reinterpret_cast<RtpHeader*>(packet->data);
+  if (packet->type == VIDEO_PACKET) {
+    if (!rtpSequenceLessThan(head->getSeqNumber(), video_rr_.max_seq) || jitter_video_.jitter == 0) {
+      return false;
+    }
+    int64_t time_diff_ms = static_cast<uint32_t>(packet->received_time_ms) - video_rr_.last_recv_ts;
```
jcague: I think we're still duplicating lines of code between video and audio cases. Is there a possible to create generic functions? let the code as is if you don't find a quick and easy solution.

kekkokk: I thought about that and I really can't find an easy and fast way to do this and i must move on other task at work. 
I'll probably refactor this in  a second time

kekkokk:
I will really appreciate you help with UTs, cause I never did one and I really haven't time to do them in these days.
I'd love if you can implement them!

kekkokk:
@lodoyun have I done something wrong with the auto merge? all tests passed in my local environment

lodoyun:
Yep, they passed for me locally also. I'm on it, trying with a clean build.

PR #706, opened 2017-01-23
Move some functionality to handlers (stats, FEC and RTCPProcessor)

jcague:
I think it's a lot easier to review this PR commit by commit.

Here I move the rest of the functionality we had in WebRtcConnection to handlers, to make it easier to implement Streams in another PR later on.


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -38,18 +39,22 @@ WebRtcConnection::WebRtcConnection(std::shared_ptr<Worker> worker, const std::st
...
+  audio_mute_handler_ = std::make_shared<RtpAudioMuteHandler>(this);
+  bwe_handler_ = std::make_shared<BandwidthEstimationHandler>(this, worker_);
+  fec_handler_ = std::make_shared<FecReceiverHandler>(this);
+  rtcp_processor_handler_ = std::make_shared<RtcpProcessorHandler>(this, rtcp_processor_);
 
   // TODO(pedro): consider creating the pipeline on setRemoteSdp or createOffer
   pipeline_->addFront(PacketReader(this));
+  pipeline_->addFront(rtcp_processor_handler_);
+  pipeline_->addFront(IncomingStatsHandler(this));
+  pipeline_->addFront(fec_handler_);
```
lodoyun: This is not specific to this PR but it's easier to see now with the handler stack here... Could the `FecReceiverHandler` affect the bitrate estimation in `BandwidthEstimationHandler`?

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -92,7 +97,7 @@ void WebRtcConnection::close() {
 }
 
 bool WebRtcConnection::init() {
-  rtcpProcessor_ = boost::shared_ptr<RtcpProcessor>(
+  rtcp_processor_ = std::shared_ptr<RtcpProcessor>(
```
lodoyun: Nice housekeeping 👍

erizo/src/erizo/rtp/RtcpProcessorHandler.cpp
```diff
@@ -0,0 +1,45 @@
...
+  } else {
+    if (connection_->getStats().getLatestTotalBitrate()) {
+      processor_->setPublisherBW(connection_->getStats().getLatestTotalBitrate());
+    }
+  }
+  processor_->checkRtcpFb();
+  ctx->fireRead(packet);
+}
+
+void RtcpProcessorHandler::write(Context *ctx, std::shared_ptr<dataPacket> packet) {
```
lodoyun: This looks good but I'm not 100% it's exactly the same as before. In any case, testing it should be pretty straightforward and my choice would be to adapt the handler to make it work here.

erizo/src/erizo/rtp/StatsHandler.cpp
```diff
@@ -0,0 +1,45 @@
...
+  connection_{connection} {}
+
+
+void IncomingStatsHandler::enable() {
+}
+
+void IncomingStatsHandler::disable() {
+}
+
+void IncomingStatsHandler::read(Context *ctx, std::shared_ptr<dataPacket> packet) {
```
lodoyun: 👍

erizo/src/test/utils/Matchers.h
```diff
@@ -0,0 +1,24 @@
+#ifndef ERIZO_SRC_TEST_UTILS_MATCHERS_H_
+#define ERIZO_SRC_TEST_UTILS_MATCHERS_H_
+
+#include <rtp/RtpHeaders.h>
+#include <MediaDefinitions.h>
+
+MATCHER_P(HasSequenceNumber, seq_num, "") {
```
lodoyun: Consider `RtpHasSequenceNumber`

erizo/src/test/utils/Tools.h
```diff
@@ -0,0 +1,86 @@
...
+
+namespace erizo {
+
+static constexpr uint16_t kVideoSsrc = 1;
+static constexpr uint16_t kAudioSsrc = 2;
+static constexpr uint16_t kArbitrarySeqNumber = 12;
+static constexpr uint16_t kFirstSequenceNumber = 0;
+static constexpr uint16_t kLastSequenceNumber = 65535;
+
+class Tools {
```
lodoyun: What do you think about making this PacketTools and have more classes here if we need other things?

erizo/src/test/utils/Matchers.h
```diff
@@ -0,0 +1,24 @@
+#ifndef ERIZO_SRC_TEST_UTILS_MATCHERS_H_
+#define ERIZO_SRC_TEST_UTILS_MATCHERS_H_
+
+#include <rtp/RtpHeaders.h>
+#include <MediaDefinitions.h>
+
```
lodoyun: Could we have this matchers also under the `erizo` namespace so it's easier to identify that they are part of the test utils?

erizo/src/erizo/WebRtcConnection.cpp
jcague: It might affect if they were configured in a different order I think. We should check it anyway in the future.

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -152,6 +152,9 @@ bool WebRtcConnection::setRemoteSdp(const std::string &sdp) {
   extProcessor_.setSdpInfo(localSdp_);
 
   bwe_handler_->updateExtensionMaps(extProcessor_.getVideoExtensionMap(), extProcessor_.getAudioExtensionMap());
+  if (remoteSdp_.supportPayloadType(RED_90000_PT) || slide_show_mode_) {
```
jcague: good catch!

erizo/src/test/utils/Matchers.h
jcague: you're right, I'll include them in the namespace too

erizo/src/test/utils/Matchers.h
jcague: agree

erizo/src/test/utils/Tools.h
jcague: I like the idea, will do.

PR #710, opened 2017-01-25
Minor fix in data base collections creation

aalonsog:
We forgot to initialize erizo controllers collection when introducing the new feature of saving them in the DB. This crashes fresh installations of Licode when no erizo controller is registered in Nuve yet (and it is searching for them in a keep alive check).

Fixes [issue 708](https://github.com/ging/licode/issues/708)


garcianavalon:
haha great! 💃

PR #711, opened 2017-01-26
SlideshowMode can be configured at different periods now

lodoyun:
By passing a period in milliseconds instead of `true` to the slideShowMode option from erizoClient, you can now adjust the slideShow frame rate.


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -440,7 +441,7 @@ exports.ErizoJSController = function (threadPool) {
     /*
      * Enables/Disables slideshow mode for a subscriber
      */
-    that.setSlideShow = function (slideShowMode, from, to) {
+    that.setSlideShow = function (slideShowMode, from, to, period) {
```
jcague: we don't need this period here, right?

PR #713, opened 2017-01-26
Add/frame rate constraints

lodoyun:
Video frameRate for a local stream can now be configured in the constructor `Erizo.Stream(config)` by setting `config.videoFramerate = [minFrameRate, maxFrameRate]`. 
Note this are _optional_ or _ideal_ constraints meaning getUserMedia will ignore the constraint if it cannot be complied


kekkokk:
I'll suggest to cap default @ 25 fps to avoid problem with bwe handler. 
In cases where framerate >=30 the bwe estimation drop as described https://github.com/ging/licode/issues/689 here

jcague:
@kekkokk we've seen cases @30fps and it seems to work fine, although we need to keep working on it. There's an open PR #712 that could help you, please try it.

lodoyun:
We can consider setting a reasonable default value but for now I'm merging this.

PR #724, opened 2017-01-31
[PoC] Test VP9 SVC

jcague:
This is a proof of concept to test VP9 SVC functionality in Chrome. It's being helpful for us to investigate RTP data, the standard, and do previous checks of feasibility.

**Please, don't use it in production environments, we won't merge this PR with master because there's much work to be done before having it working properly.**


jcague:
This has been already implemented in the master branch.

PR #725, opened 2017-01-31
[BUGFIX] delete old peerConnection.

kekkokk:
ATM, in a p2p call, if you unsubscribe to a stream and you try to re-subscribe you will get an error.


kekkokk:
we are actually deleting stream.stream, not stream.pc.
so, if you are on p2p, unsubscribe from a stream and try to resubscribe, you will get an error cause its trying to do things on the old peerconnection, that is actually closed.
apparently, if we delete the peerconnection it will create a new one and everything's fine

kekkokk:
can you please merge? so Travis will not be driven crazy

PR #730, opened 2017-02-02
[BUGFIX] subscribe on room in connecting state

kekkokk:
fix a bug where “stream-added” was called between the room.connect() method and its event callback “room-connected”.

this caused the subscribe method to try to open a peer connection but spec aren’t available yet (such as iceServers, maxVideoBW, etc.).
This is an issue when we have to rely on TURN, cause the iceServers list is empty and the peerconnection fails without notifying us.

### DIFF
--- a/erizo_controller/erizoClient/src/Room.js
+++ b/erizo_controller/erizoClient/src/Room.js
@@ -726,6 +726,12 @@ Erizo.Room = function (spec) {
 
     // It subscribe to a remote stream and draws it inside the HTML tag given by the ID='elementID'
     that.subscribe = function (stream, options, callback) {
+        if (that.state !== CONNECTED) {
+            if (callback)
+                callback(undefined, 'Unable to subscribe. Room is not connected yet, retry');
+            L.Logger.error('Unable to subscribe. Room is not connected yet, retry');
+            return;
+        }
 
         options = options || {};
 


lodoyun:
I'm failing to reproduce this, are there any special conditions that could help me reproduce it?. Do you have logs of this happening? Could you trace the control messages over the websocket? Client side should be enough but erizoController logs could be helpful, too.
Just by looking at the code, I cannot see how `onAddStream` would be received before the `token` call succeeds, fills in the Ice information triggers `room-connected`.

kekkokk:
yeah, in a few minutes i'll post you a log. 
meanwhile my scenario is:
remote licode server on aws.
2 laptops (one behind a symmetricNat).
room set in p2p mode.
turn parameters set in config.

write the address of the remote basicexample in each pc and join them almost at the same time.
the one behind nat will publish his stream but get an infinite spinning wheel

kekkokk:
ok, so, apparently this issue could be caused by a bad implementation of mine.
on the erizoController, in the socket.on('token') I added a function that call a REST to generate turn credentials BEFORE returning the callback 'success'. this can cause a slightly delay between the effective enstablishing of the socket connection and the callback with the iceServers that will fire 'room-connected'.

Anyway, this suggest me that it could happen also with a network-delayed callback, so this check could help in some rare case. 
Try to post a log anyway

lodoyun:
Ok... That would probably explain it. The thing is that in the current implementation, there is no way (unless there's a bug somewhere) the `onAddStream` would be sent before the success callback from the token, the adding the socket to the list and the callback are in the same synchronous block.
Even if there is network delay or heavy jitter, we would always send the callback _before_ the `onAddStream` and we trust TCP to deliver our packets in order.
If you add an asynchronous call in the middle and delay the success callback AFTER adding the websocket to the room.sockets list, you're introducing a race condition and the check would be needed but I suggest adding the socket to the list in the callback of the REST call. Or, even better, get the TURN credentials out-of-band and not within Licode to prevent this kinds of issues.

This, however, opens a totally different issue: we're not handling ICE failures correctly in p2p rooms.

kekkokk:
totally agree. thanks

PR #739, opened 2017-02-07
Erizofc logger

aalonsog:
Added logger in erizofc stack. 

I've also exported logger to make it available when requiring erizofc module. 

IMPORTANT. This changes the way of including erizofc library in Node applications: 

```
var Erizo = require('./licode/erizofc.js').Erizo;
var L = require('./licode/erizofc.js').L;
```
Exporting logger is useful for managing log levels:
```
L.Logger.setLogLevel(2);
```

Note: if approved, I would update the[ API documentation](http://lynckia.com/licode/client-api.html#erizofc)


lodoyun:
Does this affect spine? If so, could you please fix it in this PR?
LGTM otherwise 👍

aalonsog:
@lodoyun, it just matter of how erizojs module is required, and it is already modified in the original PR for Spine.

PR #743, opened 2017-02-08
add support for Microsoft Edge

mccob:
Microsoft Edge (in Windows 10 inisder) support WebRTC 1.0 (but no datachannel)
https://blogs.windows.com/msedgedev/2017/01/31/introducing-webrtc-microsoft-edge/

also have to add adapter.js before erizo.js call in index.html :

    	<script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>

This PR add Edge recognition and use BowserStack that's work well with Edge.

### DIFF
--- a/doc/client_api.md
+++ b/doc/client_api.md
@@ -391,8 +391,20 @@ room.publish(localStream, {maxVideoBW:300}, function(id, error){
   } else {
     console.log("Published stream", id);
   }
-});            
+});
 ```
+
+### Publish Simulcast Streams
+
+You can enable Simulcast in the publisher by adding the next option:
+
+```
+room.publish(localStream, {simulcast: {numSpatialLayers: 2}});
+```
+
+Being `numSpatialLayers` the max number of spatial layers the publisher will send.
+
+**Note:** Simulcast only works with Google Chrome and it's not compatible with Licode's recording yet.
 
 ## Subscribe to a remote stream
 
--- a/erizo_controller/erizoClient/src/Room.js
+++ b/erizo_controller/erizoClient/src/Room.js
@@ -441,11 +441,10 @@ Erizo.Room = function (spec) {
             options.minVideoBW = spec.defaultVideoBW;
         }
 
-        // TODO(javier): remove dangling once Simulcast is stable
-        options._simulcast = options._simulcast || false;
+        options.simulcast = options.simulcast || options._simulcast || false;
 
         // 1- If the stream is not local or it is a failed stream we do nothing.
-        if (stream && stream.local && !stream.failed && 
+        if (stream && stream.local && !stream.failed &&
             that.localStreams[stream.getID()] === undefined) {
 
             // 2- Publish Media Stream to Erizo-Controller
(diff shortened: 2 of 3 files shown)


aalonsog:
Tkanks! This is a very interesting feature. However I have a doubt, why do we need to import the adapter?

Unfortunately right now I have not access to an Edge installation in Windows Insider Preview so I cannot test this :(

mccob:
MS mention that ". If you need support for objects or advanced features such as multi-stream and simulcast with the current WebRTC 1.0 API, we recommend the adapter.js library, which now supports Microsoft Edge."

Without, it's not working. I haven't search why for now.

Insider Preview is available after few hour after joining program if you have a Windows 10 test computer (because it's unstable).

aalonsog:
Thanks for the explanation. From my point of view, right now we should make it compatible without having to import the adapter. So... could you try to find out what is needed and include it in a new Edge stack?

On the other hand, perhaps it is a good moment to discuss if we want to migrate our Browser stacks to the adapter...

kekkokk:
With the upcoming windows 10 creators update, which will be released today, I can confirm that simply adding adapter to the basicexample index.html edge (ver. >40) and modding erizo to use bowser stack works.
There is still work to do about bwestimation etc. to make it work "properly" but I think that we should consider abstract the getusermedia etc. layer and let the job be done to adapter.

I'm testing right now and let you know about my findings

kekkokk:
It seems only working in NON p2p scenario, and I noticed that the rembs are ignored by edge which sends always low rate ~300kbit/s (I have a 3mbit cap).

also debugging into the feedback, when subscribing with chrome, these are not delivered by licode to edge 'cause of a wrong ssrc. any news about that? @mccob are you able to use edge?

kekkokk:
for those who wants to help: [Discourse Discussion](http://discourse.lynckia.com/t/edge-supports-vp8-vp9-on-newest-creators-update/469)

mccob:
@kekkokk still able to use Edge but with default scenario/configuration yes (P2P is not default I think)

kekkokk:
@mccob yeah but monitoring the network are you able to get more than 300kbit/s??
seems that edge is ignoring my rembs to get an higher bitrate

mccob:
yes seems, actually Edge is  sending at 1 Mbits/s

kekkokk:
thanks, I'll dig into it

mccob:
seems to be an issue with Edge calls records (MKV), after few seconds, video always cut,  Sound is OK

kekkokk:
@mccob can you provide me some detailed infos on how you implemented edge and maybe do a pr? I am not able to ramp up over 300, tried everything

HemanthAnakapalle:
**Hi @mccob ... Can you please provide the details to integrate with Edge?**

I added below stated adapter.js.
<script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>

And modified Connection.js, Erizo.getBrowser() ... to return the value as "bowser"

But how ever ... we are facing "webkitRTCPeerConnection .... undefined" Issue in BowserStack.js.
And we changed "webkitRTCPeerConnection" with "window.RTCPeerConnection" ... which is part of adapter-latest.js

Now we are facing "Invalid Argument" for the below line.
adapter-latest.js
      var event = new Event('negotiationneeded');
      self.dispatchEvent(event); -------> Issue with this line.

In client licode ... dispatchEvent accepts Erizo.Event right?

**How did you managed to make it work?**

Thanks,
Hemanth

mccob:
update : licode no longer provide bowser stack. So I close this PR.

Actually edge is not working https://github.com/lynckia/licode/issues/1330

mccob:
https://blogs.windows.com/windowsexperience/2018/12/06/microsoft-edge-making-the-web-better-through-more-open-source-collaboration/

tl;tr: edge move to chromium as on Android

Definitely close this subject

PR #749, opened 2017-02-10
Add new generic stats data

jcague:
New data structure to store stats.

### DIFF
--- a/erizo/src/erizo/stats/StatNode.cpp
+++ b/erizo/src/erizo/stats/StatNode.cpp
@@ -90,7 +90,6 @@ uint64_t RateStat::value() {
 }
 
 std::string RateStat::toString() {
-  checkPeriod();
   return std::to_string(value());
 }
 
--- a/erizo/src/erizo/stats/StatNode.cpp
+++ b/erizo/src/erizo/stats/StatNode.cpp
@@ -61,8 +61,8 @@ StatNode& CumulativeStat::operator+=(uint64_t value) {
   return *this;
 }
 
-RateStat::RateStat(duration period, double rate_factor, std::shared_ptr<Clock> the_clock)
-  : period_{period}, rate_factor_{rate_factor}, calculation_start_{the_clock->now()}, last_{0},
+RateStat::RateStat(duration period, double scale, std::shared_ptr<Clock> the_clock)
+  : period_{period}, scale_{scale}, calculation_start_{the_clock->now()}, last_{0},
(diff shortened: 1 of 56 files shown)


erizo/src/erizo/stats/StatNode.h
```diff
@@ -0,0 +1,115 @@
...
+  StatNode() {}
+  virtual ~StatNode() {}
+
+  virtual StatNode& operator[](std::string key);
+
+  virtual StatNode& operator[](uint64_t key) { return (*this)[std::to_string(key)]; }
+
+  template <typename Node>
+  void insertStat(std::string key, Node&& stat) {  // NOLINT
+    node_map_.insert(std::make_pair(key, std::make_shared<Node>(std::forward<Node>(stat))));
```
lodoyun: Can you explain the use of std::forward here? Maybe even leave a comment so we don't break this in the future

erizo/src/erizo/stats/StatNode.cpp
```diff
@@ -0,0 +1,106 @@
...
+  checkPeriod();
+}
+
+uint64_t RateStat::value() {
+  checkPeriod();
+  return last_period_calculated_rate_;
+}
+
+std::string RateStat::toString() {
+  checkPeriod();
```
lodoyun: You don't need this as you already do `checkPeriod()` in `value()`

erizo/src/erizo/stats/StatNode.cpp
```diff
@@ -0,0 +1,106 @@
...
+  return text.str();
+}
+
+StatNode& CumulativeStat::operator=(uint64_t initial) {
+  total_ = initial;
+  return *this;
+}
+
+StatNode CumulativeStat::operator++(int value) {
+  CumulativeStat node{total_};
```
lodoyun: Couldn't you return `*this`? like in `operator=`
Am I missing something?

jcague: It's just the prototype used for the postincrement operator. There's more info here: http://en.cppreference.com/w/cpp/language/operator_incdec

erizo/src/erizo/stats/StatNode.h
jcague: sure! will add a line to remind us its meaning in the future.

erizo/src/erizo/stats/StatNode.cpp
jcague: good catch!

PR #752, opened 2017-02-13
Create stats using the new data structure

jcague:
With this change it will be way too easier to add new stats. I've also made it asynchronous to be able to remove some mutex.

### DIFF
--- a/erizoAPI/WebRtcConnection.cc
+++ b/erizoAPI/WebRtcConnection.cc
@@ -69,6 +69,7 @@ NAN_MODULE_INIT(WebRtcConnection::Init) {
   Nan::SetPrototypeMethod(tpl, "setVideoReceiver", setVideoReceiver);
   Nan::SetPrototypeMethod(tpl, "getCurrentState", getCurrentState);
   Nan::SetPrototypeMethod(tpl, "getStats", getStats);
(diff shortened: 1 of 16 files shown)


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -508,33 +508,35 @@ exports.ErizoJSController = function (threadPool) {
         }
     };
 
+    /* eslint no-param-reassign: ["error", { "props": false }] */
+    const getWrtcStats = (label, stats, wrtc) => {
+      const promise = new Promise((resolve) => {
+        wrtc.getStats((statsString) => {
+          console.log('OKKKKK', arguments, statsString);
```
lodoyun: OKKKKKK

erizoAPI/WebRtcConnection.cc
```diff
@@ -16,6 +18,29 @@ using v8::Exception;
 using v8::Value;
 using json = nlohmann::json;
 
+StatCallWorker::StatCallWorker(Nan::Callback *callback, std::weak_ptr<erizo::WebRtcConnection> weak_connection)
```
lodoyun: 👍  to async'ing this

PR #764, opened 2017-02-17
Generate a 18 digits safe integer

simonemazzoni:
Hello,
I noticed that the function `id = Math.random() * 1000000000000000000;` can return a floating point numbers (i.e. `if Math.random()` return `0,04738571228`, then the id would be `473857122775589.44`)

Since the stream id should be an integer, and is passed as integer and not as String through socket messages, I think that use a function that ensures this id to be a non floating number is mandatory to avoid parsing errors.


simonemazzoni:
Hi @jcague, I see you merged master into the branch of this PR.
Any plan to merge this?

jcague:
thanks @simonemazzoni !

simonemazzoni:
Great! 👍

kekkokk:
It appears that a 18 digit integer is never considered a "safe integer".
for ex. the controller just generated a: 542260870009211140

in Internet Explorer:
`var test = 542260870009211140;
console.log(test);
542260870009211130`

this can cause a mess!
I think the way to go would be:
`id = Math.floor(Math.random() * (Number.MAX_SAFE_INTEGER));`

Issue #616, opened 2016-11-19
Can not re-subscribe after stream failure


mkhahani:
Trying to handle `stream-failed` event by re-subscribing to the same stream passed by the event with no luck. The SDP is sent but no response from the server is received and there is no error on the server side.

It's while the same approach works well for publisher, that is when a publish stream fails I can re-publish by handling `stream-failed` event.

PS: I use NetLimiter in order to drop or limit a connection(stream).

simonemazzoni:
I have the very same  problem.
I noticed that in case of `stream-failed` the property `failed` of the stream is set to 'true'

That value is then checked on signaling messages received, both from the server and other peers (not p2p and p2p), not processing the signaling message in case the stream has the `failed` property set to `true`.

Here the two part of `Room.js` where all happen.
 ```
// if !p2p
that.socket.on('signaling_message_erizo', function (arg) {
    var stream;
    if (arg.peerId) {
        stream = that.remoteStreams[arg.peerId];
    } else {
        stream = that.localStreams[arg.streamId];
    }

    if (stream && !stream.failed) {
        stream.pc.processSignalingMessage(arg.mess);
    }
});

// if p2p "You can see that the message is processed anyway"
that.socket.on('signaling_message_peer', function (arg) {

    var stream = that.localStreams[arg.streamId];

    if (stream && !stream.failed) {
        stream.pc[arg.peerSocket].processSignalingMessage(arg.msg);
    } else {
        stream = that.remoteStreams[arg.streamId];

        if (!stream.pc) {
            createRemotePc(stream, arg.peerSocket);
        }
        stream.pc.processSignalingMessage(arg.msg);
    }
});
```

The first workaround I find is to manually delete the `failed` property from the stream on `stream-failed`. This bypass that check and allow a stream to be published again.

The problem is that the `stream-failed` event could be fired both for a publisher and for a subscriber.
In the Subscriber `stream-failed` case I don't think it is correct or useful to allow a user to re-subscribe the same failed stream (since isn't on the server anymore).

In the Publisher `stream-failed` case I think instead that allow the user to republish his local stream is the correct thing to do.

I created a pull request with a possible fix. You can check it [here](https://github.com/lynckia/licode/pull/744)

Let me know.

Simone

aalonsog:
Republishing a stream that has previously failed is a very sensitive issue. There are several checks implied in both client and server side. So we have to handle it carefully. My opinion is that we would have to analyze every possible scenario and try to search a robust solution. 

[...]

PR #765, opened 2017-02-17
Added callbacks error in case o f publish and subscribe of a failed s…

simonemazzoni:
Hello,
I reimplemented the fixes discussed [here #616](https://github.com/lynckia/licode/issues/616)

Now a callback error is returned every time someone tries to publish or subscribe a failed stream.

Also the property `failed` is removed from a stream after a successful unpublish of that stream. This allow the publisher to try to republish it local stream.


simonemazzoni:
@aalonsog great!

Please, let me know when you start reviewing #616 stuff so as I can see if I can help.

PR #770, opened 2017-02-21
Add a new Sequence Number translator for Simulcast/SVC

jcague:
With this new translator we should be able to seamlessly send packets that belong to different temporal and spatial layers in SVC and temporal layers in Simulcast.

### DIFF
--- a/erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
+++ b/erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
@@ -1,0 +1,114 @@
+#include "rtp/SequenceNumberTranslator.h"
+
+#include <vector>
+
+namespace erizo {
+
+constexpr uint16_t kMaxSequenceNumberInBuffer = 511;  // = 65536 / 128 and more than 10s of audio @50fps
+constexpr uint16_t kMaxDistance = 200;
+
+DEFINE_LOGGER(SequenceNumberTranslator, "rtp.SequenceNumberTranslator");
+
+SequenceNumberTranslator::SequenceNumberTranslator()
+    : in_out_buffer_{kMaxSequenceNumberInBuffer},
+      out_in_buffer_{kMaxSequenceNumberInBuffer},
+      first_input_sequence_number_{0},
+      last_input_sequence_number_{0},
+      initial_output_sequence_number_{0},
+      initialized_{false}, reset_{false} {
+}
+
+bool SequenceNumberTranslator::lessThan(uint16_t first, uint16_t last) const {
+  uint16_t result = first - last;
+  return result > 0xF000;
+}
+
+void SequenceNumberTranslator::add(SequenceNumber sequence_number) {
+  in_out_buffer_[sequence_number.input % kMaxSequenceNumberInBuffer] = sequence_number;
+  out_in_buffer_[sequence_number.output % kMaxSequenceNumberInBuffer] = sequence_number;
+}
+
+uint16_t SequenceNumberTranslator::fill(const uint16_t &first,
+                                        const uint16_t &last) {
+  const SequenceNumber& first_sequence_number = get(first);
+  uint16_t output_sequence_number = first_sequence_number.output;
+
+  if (first_sequence_number.type != SequenceNumberType::Skip) {
+    output_sequence_number++;
+  }
+
+  for (uint16_t sequence_number = first_sequence_number.input + 1;
+       lessThan(sequence_number, last);
+       sequence_number++) {
+    add({sequence_number, output_sequence_number++, SequenceNumberType::Valid});
+  }
+
+  return output_sequence_number;
+}
+
+SequenceNumber& SequenceNumberTranslator::internalGet(uint16_t input_sequence_number) {
+  return in_out_buffer_[input_sequence_number % kMaxSequenceNumberInBuffer];
+}
+
+SequenceNumber SequenceNumberTranslator::get(uint16_t input_sequence_number) const {
+  return in_out_buffer_[input_sequence_number % kMaxSequenceNumberInBuffer];
+}
+
+SequenceNumber SequenceNumberTranslator::get(uint16_t input_sequence_number, bool skip) {
+  if (!initialized_) {
+    SequenceNumberType type = skip ? SequenceNumberType::Skip : SequenceNumberType::Valid;
+    uint16_t output_sequence_number = reset_ ? initial_output_sequence_number_ : input_sequence_number;
+    add({input_sequence_number, output_sequence_number, type});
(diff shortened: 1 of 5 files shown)


erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
```diff
@@ -0,0 +1,114 @@
...
+  uint16_t output_sequence_number = first_sequence_number.output;
+
+  if (first_sequence_number.type != SequenceNumberType::Skip) {
+    output_sequence_number++;
+  }
+
+  for (uint16_t sequence_number = first_sequence_number.input + 1;
+       lessThan(sequence_number, last);
+       sequence_number++) {
+    add({sequence_number, output_sequence_number++, SequenceNumberType::Valid});
```
lodoyun: I really don't like this way of passing new objects. It may be less verbose but I think it's less readable.

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
```diff
@@ -0,0 +1,114 @@
...
+SequenceNumber SequenceNumberTranslator::get(uint16_t input_sequence_number, bool skip) {
+  if (!initialized_) {
+    SequenceNumberType type = skip ? SequenceNumberType::Skip : SequenceNumberType::Valid;
+    uint16_t output_sequence_number = reset_ ? initial_output_sequence_number_ : input_sequence_number;
+    add({input_sequence_number, output_sequence_number, type});
+    last_input_sequence_number_ = input_sequence_number;
+    first_input_sequence_number_ = input_sequence_number;
+    initialized_ = true;
+    reset_ = false;
+    return internalGet(input_sequence_number);
```
lodoyun: Is there a reason to use `internalGet` if you're not going to use the reference?

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  if (lessThan(input_sequence_number, first_input_sequence_number_)) {
+    return {input_sequence_number, 0, SequenceNumberType::Skip};
+  }
+
+  if (lessThan(last_input_sequence_number_, input_sequence_number)) {
+    uint8_t output_sequence_number = fill(last_input_sequence_number_, input_sequence_number);
```
lodoyun: fill returns `uint16_t`

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
```diff
@@ -0,0 +1,114 @@
...
+SequenceNumberTranslator::SequenceNumberTranslator()
+    : in_out_buffer_{kMaxSequenceNumberInBuffer},
+      out_in_buffer_{kMaxSequenceNumberInBuffer},
+      first_input_sequence_number_{0},
+      last_input_sequence_number_{0},
+      initial_output_sequence_number_{0},
+      initialized_{false}, reset_{false} {
+}
+
+bool SequenceNumberTranslator::lessThan(uint16_t first, uint16_t last) const {
```
jcague: I agree, I can implement a generic RtpUtils and start adding this specific case, wdyt?

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
jcague: I can see your point. I think it's computationally cheaper because we can avoid some copy. Anyway I can send values as arguments if you prefer.

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
jcague: good catch!

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
jcague: I wanted to be sure I don't copy the same object too many times, but probably the compiler knows how to do it, I'll check if there's some additional copy if I use the public version.

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
jcague: we'll use `add(SequenceNumber{sequence_number, output_sequence_number++, SequenceNumberType::Valid});` as the best approach.

erizo/src/erizo/rtp/SequenceNumberTranslator.cpp
lodoyun: Yep, that sounds good

nicolashuSE:
@jcague I'm a newbie of audio/video multipart application, and working on licode currently. About the SequenceNumberTranslator I have some questions as follows:
1）Why needs to do 'sequence number translation'? What is the principle/mechanism/specification 'sequence number translation' is accordance with？
2）You say "Add a new Sequence Number translator for Simulcast/SVC", why needs to do 'sequence number translation' in RtpSlideShowHandler, RtpPaddingRemovalHandler, RtpPaddingGeneratorHandler?

I really hope that you can explain to me about these questions in detail. Or, give me some material Link.
Thanks a lot!

nicolashuSE:
@jcague @lodoyun please give some help about the questions mentioned before, thanks u all!

kekkokk:
I don't know if its an rfc or not, but chrome, if switching from one spatial to another, and the pictureID is not incremental, freezes and can't decode frame for few sec, until a keyframe re-start the decoding

See: https://github.com/lynckia/licode/issues/952

nicolashuSE:
@kekkokk thanks your reply, I have checked #952. I think #952 is not my question, which is about the pictureId rewriting. My question is about the RTP sequence number translation of this commit, and this commit happens before #952 .

nicolashuSE:
@jcague @lodoyun @kekkokk please give me a little time about this question, I'll appreciate it very much.

yayuntian:
+1

marshuman:
my question the same as nicolashuSE,can @jcague can u help us?

kekkokk:
when you receive a simulcast stream, you receive 2 stream with 2 different resolutions. these streams are identified by their ssrc, but each starts from a different sequence number.

the idea is to send the best layer to te person who is receiving the stream, based on his available bandwidth.

The problem is that the subscriber decoder, doesn't know about the stream being simulcasted, so you need to send him a stream that is more spec compliant possible.

To do this, the decoder expects a stream with sequence number which are growing monotonically.
When you decide in your codebase to drop a frame (slideshow handler or padding removal handler), or change the spatial layer (resolution), you need to rewrite the seq. numbers to be monotonically growing so the decoder does not compain and sends back nacks or plis.

PR #775, opened 2017-02-22
Handler to filter packets from higher quality layers

jcague:
**Description**
This PR adds a new handler to drop packets that belong to spatial/temporal layers that we don't want to forward to the subscribers.

**Changes in the Client API**
It also adds a mechanism to choose the spatial and temporal layer from the client (stream._setQualityLayer(...)). The method is prefixed with underscore because the feature is not finished yet and we don't want people to start using it in production yet.

I'll update client's documentation once we make the API public.


erizo/src/erizo/rtp/QualityFilterHandler.cpp
```diff
@@ -0,0 +1,130 @@
...
+  enabled_ = true;
+}
+
+void QualityFilterHandler::disable() {
+  enabled_ = false;
+}
+
+void QualityFilterHandler::handleFeedbackPackets(std::shared_ptr<dataPacket> packet) {
+  RtpUtils::forEachRRBlock(packet, [this](RtcpHeader *chead) {
+    RtpUtils::updateREMB(chead, 400000);
```
lodoyun: Why 400000? Even if we have proper feedback soon, it probably would make more sense to use `maxVideoBW` here

erizo/src/erizo/rtp/RtpUtils.h
```diff
@@ -1,13 +1,27 @@
...
 
+#include <memory>
+
 namespace erizo {
 
 class RtpUtils {
  public:
   static bool sequenceNumberLessThan(uint16_t first, uint16_t second);
+
+  static void forEachRRBlock(std::shared_ptr<dataPacket> packet, std::function<void(RtcpHeader*)> f);
```
lodoyun: 👍 to this... very helpful

erizo/src/erizo/rtp/QualityFilterHandler.cpp
```diff
@@ -0,0 +1,130 @@
...
+      if (ssrc == video_sink_ssrc_) {
+        translator_.get(sequence_number, true);
+      }
+      return;
+    }
+
+    checkSSRCChange(ssrc);
+    rtp_header->setSSRC(video_sink_ssrc_);
+
+    if (packet->compatible_spatial_layers.back() == target_spatial_layer_ && packet->ending_of_layer_frame) {
```
lodoyun: If I understand this right, if we put this after current line 95, we won't be setting the marker for packets that we're going to discard anyway

erizo/src/erizo/rtp/QualityFilterHandler.cpp
jcague: you got it!

erizo/src/erizo/rtp/QualityFilterHandler.cpp
jcague: :+1:

PR #781, opened 2017-02-23
Assign controller to room

wolmi:
**Description**

*The assignment of erizoController to a room don't work on sharded mongodb deployments, it only works in monolitic mongodb deployments*

*I re-write complete assignErizoControllerToRoom function removing the use of db.eval(), this function don't work on sharded mongodb deployments. For more info: https://docs.mongodb.com/manual/reference/method/db.eval/#sharded-data*

### DIFF
--- a/nuve/nuveAPI/mdb/roomRegistry.js
+++ b/nuve/nuveAPI/mdb/roomRegistry.js
@@ -49,13 +49,14 @@ exports.addRoom = function (room, callback) {
 };
 
 exports.assignErizoControllerToRoom = function(room, erizoControllerId, callback) {
-  // TODO: Add error handling
   db.rooms.findOne({_id: db.ObjectId(room._id)}, function(err, findRoom){
+    if (err) log.warn('message: assignErizoControllerToRoom error, ' + logger.objectToLog(err));
     if (!findRoom) {
        callback(undefined);
     }
     if (findRoom.erizoControllerId) {
       db.erizoControllers.findOne({_id: findRoom.erizoControllerId}, function(err, assignedErizoController){
+        if (err) log.warn('message: assignErizoControllerToRoom error, ' + logger.objectToLog(err));
         if (assignedErizoController) {
           callback(assignedErizoController);
         }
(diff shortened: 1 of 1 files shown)


wolmi:
Hi

No, is not an option because we are using mongo cluster and are planing to expand mongo with replica sets geo-localized.

I'll work on this issue a bit more to convert to atomic transaction.

Thanks

wolmi:
Hi,

To complete the info I shared in the first comment:
https://docs.mongodb.com/manual/reference/method/db.eval/
You can see that on the same page but just at the begging:

image

Please can you consider seriously the refactor of this function?

If you review my code is almost atomic, the only thing is that I define two scenarios to perform an update or insert. But if you prefer make new approach please let me know.

aalonsog:
Hi @wolmi, think in an scenario in which a room is empty, so no erizoController is assigned to it yet. Then...

- User 1 connects to the room and Cloud Handler calls `assignErizoControllerToRoom` method to know where to allocate the user. An asynchronous query to the DB is performed to check if there is an ErizoController already assigned to the room.
- Before having a response of the previous call, User 2 connects to the same room and Cloud Handler calls again `assignErizoControllerToRoom` method. An asynchronous query to the DB is performed to check if there is an ErizoController already assigned to the room.
- The result of both async calls is that no Erizo Controller is assigned to the room so each flow continue assigning a new Erizo Controller that, depending in the priority list or the scheduling algorithm could be different Erizo Controllers. The final result is that two users connected to the same room are assigned to different Erizo Controllers. 

You have to understand that "almost" atomic is not enough here :) I agree that the it would be convenient to support mongo clusters but we would have to find a solution to support them without compromising the reliability of the Erizo Controllers assignments.

Of course, any idea is welcome!

wolmi:
Hi @aalonsog 

Thanks for the details of case of use you are in mind, is very helpfull.

But remember that the deprecation is for mongodb 3.0 and above not only clusters, I found this by a planet alignment when I migrated licode to my centralized mongodb  with 3.2.4, it doesn't matter it is a cluster.

I will think in a better solution.

Thanks

aalonsog:
Ok, please let us updated about any news. By the moment this means we are not compatible with Mongo 3.0. 

I close this PR.

PR #784, opened 2017-02-28
Fix several loops in SequenceNumberTranslator

jcague:
**Description**

This PR fixes an issue in SequenceNumberTranslator when there are several loops of sequence numbers in total. I added a Unit Test to cover this case too.

No needed.


unclerunning:
The output seq of packet with seq 6 will mess up with seq 16 if the input packet sequence like below:
1, PacketState::Forward,

10, PacketState::Skip,

20, PacketState::Forward,  

5, PacketState::Forward,
6, PacketState::Forward,
16,PacketState::Forward,

PR #787, opened 2017-03-02
P2p fix

kekkokk:
**Description**

Small bugfix that prevented the user to subscribe-unsubscribe-resubscribe to the same video in p2p.
Also deleted the "end" candidate in chrome stable stack. seems that chrome doesn't like this candidate anymore causing an error.
The commits should be self explaining btw.


aalonsog:
Hi @kekkokk, thanks for the contribution!

Ok to the p2p thing. However, I don't understand the second point. When this candidate is failing? In p2p mode?

kekkokk:
yeah, just make a p2p call chrome to chrome and see the console output. also check the chrome://webrtc-internals

aalonsog:
Ok, I see the error. However, the `end` candidate has to be sent in case we use a centralized room. Erizo needs it in some cases. 

I propose to fix the issue by ignoring the received `end` candidates in Chrome (in processSignalingMessage function of the stack). 

What do you think?

kekkokk:
Sorry, I missed the part where erizo needs that. I just tried doing some tests and everythings worked fine. 
However you are totally right, I'll PR this change in the afternoon.

aalonsog:
Thanks!

PR #788, opened 2017-03-03
Add a handler to control PLI sending rate

jcague:
**Description**

Here we stop sending too many PLIs when multiple clients are subscribed to one publisher. It will also send a FIR packet after waiting too long for keyframes.

Not needed.


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -29,6 +29,7 @@
 #include "rtp/LayerDetectorHandler.h"
 #include "rtp/QualityFilterHandler.h"
 #include "rtp/QualityManager.h"
+#include "rtp/PLIPacerHandler.h"
```
lodoyun: I've been using camelcase also for abbreviations lately (NackGenerator). I don't have an strong opinion on what's better but we should be consistent

erizo/src/erizo/rtp/RtpUtils.cpp
```diff
@@ -35,10 +35,32 @@ void RtpUtils::forEachNack(RtcpHeader *chead, std::function<void(uint16_t, uint1
...
+        }
+  });
+  return is_fir;
+}
+
 std::shared_ptr<dataPacket> RtpUtils::createPLI(uint32_t source_ssrc, uint32_t sink_ssrc) {
   RtcpHeader pli;
   pli.setPacketType(RTCP_PS_Feedback_PT);
-  pli.setBlockCount(1);
+  pli.setBlockCount(RTCP_PLI_FMT);
```
lodoyun: 👍

PR #790, opened 2017-03-07
Add layer bitrate calculation handler + Test page

lodoyun:
**Description**
This PR includes:
- Layer bitrate calculation handler: This handler calculates the bitrate of each available layer and stores it in the stats under `["qualityLayers"][spatialLayer][temporalLayer]` where `spatialLayer` and `temporalLayer` are the numeric values corresponding to the layer.
Layers that have never been active will not exist in the map whereas layers previously active will have 0 bitrate.

- A layer test page that can be run similarly to the basicExample but under `https://basicexampleserverurl.com/quality_layers.html` This page subscribes to the first stream that arrives (or publish and subscribe to the same stream with the `shouldPublish` option). Then, it will chart all the available layer bitrate information, as well as the total calculated bitrate and the estimated available bitrate for the subscriber.

None.


kekkokk:
is there any advice on how to use the test page? did a ./installErizo and ./installBasicexample but i can only see me. no graphs or chart in the page is shown

PR #807, opened 2017-03-17
Added the possibility of dynamically creating p2p rooms in basic example

aalonsog:
**Description**

Now you can specify the room type when dynamically creating rooms in basic example. Connecting to `https://basic_server:3004/?room=1234&type=p2p` creates a p2p room called '1234'.

No changes

### DIFF
--- a/extras/basic_example/basicServer.js
+++ b/extras/basic_example/basicServer.js
@@ -72,7 +72,7 @@ var getOrCreateRoom = function (name, type = 'erizo', callback) {
             }
         }
 
-        let extra = {data: {basicExampleRoom:true}};
+        let extra = {data: {basicExampleRoom: true}};
         if (type === 'p2p') extra.p2p = true;
 
         N.API.createRoom(name, function (roomID) {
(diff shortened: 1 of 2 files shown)


extras/basic_example/basicServer.js
```diff
@@ -142,21 +147,22 @@ app.get('/getUsers/:room', function(req, res) {
...
+    
+    let username = req.body.username;
+    let role = req.body.role;
+    
+    let room = defaultRoomName, type;
 
-    var username = req.body.username,
-    role = req.body.role;
+    if (req.body.room && !isNaN(req.body.room)) room = req.body.room;
+    if (req.body.type) type = req.body.type;
```
jcague: Could we use something like:
```
let room = req.body.room || defaultRoomName;
let type = req.body.type;
```
or even:
```
let {body: {room = defaultRoomName, type}} = req;
```
😄

aalonsog: We have to check that req.body.room is an integer (`!isNaN(req.body.room)`). So I'm not sure if just a line is recommended (because probably is possible :D)

PR #814, opened 2017-03-22
Add RoomId option to quality layers

jcague:
**Description**

Add a new option to quality layers to load info from a known room Id.

Not needed.

### DIFF
--- a/erizo_controller/erizoAgent/log4cxx.properties
+++ b/erizo_controller/erizoAgent/log4cxx.properties
@@ -1,5 +1,5 @@
 # Set root logger level to DEBUG and its only appender to A1.
-log4j.rootLogger=DEBUG, A1
+log4j.rootLogger=INFO, A1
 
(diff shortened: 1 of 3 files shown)


erizo_controller/erizoAgent/log4cxx.properties
```diff
@@ -1,5 +1,5 @@
 # Set root logger level to DEBUG and its only appender to A1.
-log4j.rootLogger=DEBUG, A1
+log4j.rootLogger=INFO, A1
```
aalonsog: Do you want to change the log level in this PR?

jcague: yes, it's an intentional change because I don't think we need to enable DEBUG level by default. But I can do it in a separate PR if we prefer.

aalonsog: It's ok for me. Just to check it was't a mistake :)

PR #821, opened 2017-03-23
QualityManager will be disabled unless is_scalable

lodoyun:
**Description**

The `QualityManager` will now be disabled by default (and padding also), until `QualityFilterHandler` detects the first packet of an scalable layer. 


erizo/src/erizo/rtp/QualityManager.cpp
```diff
@@ -10,13 +10,27 @@ constexpr duration QualityManager::kActiveLayerInterval;
...
   last_activity_check_{the_clock->now()}, clock_{the_clock} {}
 
+void QualityManager::enable() {
+  ELOG_DEBUG("message: Enabling QualityManager");
+  enabled_ = true;
+  setPadding(true);
+}
+
+void QualityManager::disable() {
+  enabled_ = false;
```
jcague: should we also a log here to be consistent?

PR #822, opened 2017-03-24
Disable Retransmissions when available bw is lower than base layer

jcague:
**Description**

It now disables retransmissions when the bandwidth estimation is lower than the base layer in Simulcast/SVC. It helps in cases with very low bandwidth connections.

Not needed.


jcague:
we're following a different approach with slideshow so I close this PR, which is not needed anymore.

PR #826, opened 2017-03-27
Add a Token Bucket to limit RTX rate

jcague:
**Description**

Implements a Token Bucket mechanism to limit retransmission rates.

Not needed.


erizo/src/erizo/rtp/RtpRetransmissionHandler.cpp
```diff
@@ -1,14 +1,20 @@
...
 
-RtpRetransmissionHandler::RtpRetransmissionHandler()
-  : connection_{nullptr},
+RtpRetransmissionHandler::RtpRetransmissionHandler(std::shared_ptr<erizo::Clock> the_clock)
+  : clock_{the_clock},
+    connection_{nullptr},
     audio_{kRetransmissionsBufferSize},
-    video_{kRetransmissionsBufferSize} {}
+    video_{kRetransmissionsBufferSize},
+    bucket_{static_cast<uint64_t>(300000. * kMarginRtxBitrate), kBurstSize, clock_},
```
lodoyun: Consider making 300000  another constant. I'm assuming it's 300kbps because of the fast start and it's more or less an _unofficial_ standard.

erizo/src/erizo/lib/TokenBucket.cpp
```diff
@@ -0,0 +1,92 @@
...
+    time_per_burst_{0},
+    clock_{the_clock} {}
+
+TokenBucket::TokenBucket(const uint64_t rate, const uint64_t burst_size,
+            std::shared_ptr<erizo::Clock> the_clock)
+  : time_ {0},
+    time_per_token_{1000000 / rate},
+    time_per_burst_{burst_size * time_per_token_},
+    clock_{the_clock} {
+  assert(rate > 0 && "Rate should be greater than 0");
```
lodoyun: Not convinced about using `assert` mostly because we're not using it in general.
We should discuss if we want to use it in general

Issue #604, opened 2016-11-16
Publishing audio on Firefox fails after recent changes


mkhahani:
Testing on the latest code (for 2 days ago), Firefox is stuck at **"Gathered all candidates. Sending END candidate"** and does not go further when publishing audio, no problem with video. Checked the SDP and everything seems fine. I could not go further but I'm almost sure this is about the server code.

Can any body confirm this?

nguyendatdh93:
Yes. I have the same problem. How to solve it?

mkhahani:
Still doesn't work.
Team, this is serious and users are complaining about it. Are you going to drop the support for Firefox?!!

PR #854, opened 2017-04-18
candidates won’t be added when Firefox in audio publish only.

kekkokk:
**Description**

testing the current branch, I noticed that firefox won't publish in audio only mode. digging into the code i found this "return" statement if the video transport wasn't specified. 
Resolved adding another check but don't know if it's ok.
Tell me what do you think.

Probably needs but doesn't include


mkhahani:
Thanks! So this might fix #604.

PR #855, opened 2017-04-18
Revert libnice to version 0.1.4 temporarily

jcague:
**Description**

We've found more ICE failures happening with v0.1.7 so we'll revert version 0.1.4 temporarily while we discuss a better approach for handling ICE (update to latest versions or even switch to another library).

Not needed.


kekkokk:
it's just a statistic or you insulated a failure in some scenarios?

jcague:
We gathered results after some A/B testing during the past weeks and they showed 0.1.4 performed better in terms of ICE failures. We don't want this solution to be permanent, so we'll prioritize some work to update to newer versions in the future.

kekkokk:
great! I also suggest to update installMacDeps

jcague:
good catch @kekkokk! I'll do it in a separate PR as this one is already closed.

PR #856, opened 2017-04-18
[PoC] Test more things to support FF simulcast

jcague:
**Description**

This is a PoC that is still Work in Progress. Please, don't review or test it yet. I've been playing a bit with Firefox Simulcast since we still need to make some progress in order to support it. I'm stuck at some point where Firefox sends multiple streams, but I can't filter them properly. 

There are some caveats: Firefox sends unannounced SSRCs for each RID, temporal layers seem to be somehow broken, I still don't understand how to properly configure layers in Firefox, etc.

Not needed.


kekkokk:
:(

PR #861, opened 2017-04-19
Make Simulcast API public

jcague:
**Description**

It allows developers to start using Simulcast in Chrome by setting a new option called simulcast.

There's a new small section to show an example of a publisher sending a Simulcast stream.

### DIFF
--- a/doc/client_api.md
+++ b/doc/client_api.md
@@ -404,6 +404,8 @@ room.publish(localStream, {simulcast: {numSpatialLayers: 2}});
 
 Being `numSpatialLayers` the max number of spatial layers the publisher will send.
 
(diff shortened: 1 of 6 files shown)


doc/client_api.md
```diff
@@ -391,9 +391,19 @@ room.publish(localStream, {maxVideoBW:300}, function(id, error){
   } else {
     console.log("Published stream", id);
   }
-});            
+});
 ```
 
+### Publish Simulcast Streams
+
+You can enable Simulcast in the publisher by adding the next option:
```
lodoyun: I'd warn here of the current simulcast limitation: i.e. Chrome only and no recording.

PR #864, opened 2017-04-21
fixed MovingAverageStat gives wrong average when new value smaller than old one

cxxrocks:
changes:
- added code handling value < old_value for MovingAverageStat
- +0.1 when casting current_average_ to uint64_t
- added some test case


cxxrocks:
going to squash commits and re-submit this PR

PR #865, opened 2017-04-21
fixed MovingAverageStat when value < old_value

cxxrocks:
- fixed MovingAverageStat when value < old_value
- +0.1 when casting _current_average to uint64, to avoid 3.999... -> 3
- added some test cases for MovingAverageStat

### DIFF
--- a/erizo/src/erizo/stats/StatNode.cpp
+++ b/erizo/src/erizo/stats/StatNode.cpp
@@ -1,5 +1,6 @@
 #include "stats/StatNode.h"
 
+#include <cmath>
 #include <sstream>
 #include <string>
 #include <memory>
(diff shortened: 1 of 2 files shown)


erizo/src/erizo/stats/StatNode.cpp
```diff
@@ -249,7 +249,8 @@ StatNode& MovingAverageStat::operator+=(uint64_t value) {
 }
 
 uint64_t MovingAverageStat::value() {
-  return static_cast<uint64_t>(current_average_);
+  // without +0.1 4.999999 will be cast to 4
+  return static_cast<uint64_t>(current_average_ + 0.1);
```
jcague: good catch! why don't we do something like `return static_cast<uint64_t>(round(current_average_));`

erizo/src/erizo/stats/StatNode.cpp
```diff
@@ -266,7 +267,11 @@ void MovingAverageStat::add(uint64_t value) {
       / (next_sample_position_ + 1);
   } else {
     uint64_t old_value = (*sample_vector_.get())[next_sample_position_ % window_size_];
-    current_average_ = current_average_ + static_cast<double>(value - old_value) / window_size_;
+    if (value > old_value) {
```
jcague: good catch!

erizo/src/erizo/stats/StatNode.cpp
cxxrocks: Yes, round(current_average_) is better. I will change that line right now.

PR #868, opened 2017-04-24
Fix/padding removal handler

lodoyun:
**Description**

This PR fixes `PaddingRemovalHandler`. It adds sequence number translation for NACKs to avoid problems when there is Padding in non-simulcast connections. Also, the handler is moved in the pipeline to avoid conflicts with `FeedbackGenerationHandler`

### DIFF
--- a/erizo/src/erizo/rtp/RtpPaddingRemovalHandler.cpp
+++ b/erizo/src/erizo/rtp/RtpPaddingRemovalHandler.cpp
@@ -18,14 +18,24 @@ void RtpPaddingRemovalHandler::disable() {
 
 void RtpPaddingRemovalHandler::read(Context *ctx, std::shared_ptr<dataPacket> packet) {
   RtcpHeader *chead = reinterpret_cast<RtcpHeader*>(packet->data);
+  RtpHeader *rtp_header = reinterpret_cast<RtpHeader*>(packet->data);
 
   if (!chead->isRtcp() && enabled_ && packet->type == VIDEO_PACKET) {
-    if (!removePaddingBytes(packet)) {
+    uint32_t ssrc = rtp_header->getSSRC();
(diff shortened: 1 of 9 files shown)


erizo/src/erizo/rtp/RtpPaddingRemovalHandler.cpp
```diff
@@ -58,19 +95,38 @@ bool RtpPaddingRemovalHandler::removePaddingBytes(std::shared_ptr<dataPacket> pa
...
 }
 
+std::shared_ptr<SequenceNumberTranslator> RtpPaddingRemovalHandler::getTranslatorForSsrc(uint32_t ssrc,
+  bool should_create) {
+    auto translator_it = translator_map_.find(ssrc);
+    std::shared_ptr<SequenceNumberTranslator> translator;
+    if (translator_it != translator_map_.end()) {
+      ELOG_DEBUG("Found Translator for %u, %s", ssrc, connection_->toLog());
+      translator = translator_it->second;
+    } else {
```
jcague: `} else if (should_create) {`?

erizo/src/erizo/rtp/RtpPaddingRemovalHandler.cpp
```diff
@@ -58,19 +95,38 @@ bool RtpPaddingRemovalHandler::removePaddingBytes(std::shared_ptr<dataPacket> pa
...
+    return translator;
+  }
+
 void RtpPaddingRemovalHandler::notifyUpdate() {
   auto pipeline = getContext()->getPipelineShared();
   if (!pipeline) {
     return;
   }
-
+  connection_ = pipeline->getService<WebRtcConnection>().get();
```
jcague: consider moving this line below, after the `if(initialized_)`

PR #871, opened 2017-04-27
New dockerfile and start script

lodoyun:
**Description**

This PR implements a new Dockerfile and docker init script.
Instead of pulling the repository, the new dockerfile uses `COPY` to generate the context for the build.

The new script allows for starting different parts of licode as well as starting the whole thing.
You can for instance run it with `--nuve` to only start nuve. For now, no further customization of the configuration is provided.

*For users of the older Dockerfile* 
This new Dockerfile is located in the root of the repository to be able to generate the context via `COPY`
Also, `CMD` has been changed to `ENTRYPOINT` to make it easier to pass parameters to the init script. Remember you have to use `--entrypoint` to change the entrypoint instead of just passing the script as a parameter.

*Add a detailed description of any change in the public APIs.*


Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+COPY . /opt/licode
+
+# Clone and install licode
+WORKDIR /opt/licode/scripts
+
+RUN ./installUbuntuDeps.sh --cleanup && \
+  ./installErizo.sh && \
+  ./../nuve/installNuve.sh && \
+	./installBasicExample.sh
```
jcague: please check indentation

extras/docker/initDockerLicode.sh
```diff
@@ -1,7 +1,167 @@
...
+    exit 1
+  fi
+}
+run_rabbitmq() {
+  echo "Starting Rabbitmq"
+  rabbitmq-server -detached
+  sleep 3
+}
+
+run_mongo() {
```
aalonsog: Perhaps we can reuse the code in /scripts/installNuve.sh by passing parameters depending on the parts we want to execute. If so we should split the populate code in that scripts in two parts: run mongo and populate mongo.

lodoyun: I agree that we're duplicating code here, but I would do it as a part of a bigger refactoring of the scripts. Right now the populating the database is tied to copying the configuration and that can cause all sorts of problems if we make it an option in the default scripts.

lodoyun:
@jcague @aalonsog 
I've gone ahead and added the doc changes here so everything stays synchronized.

PR #877, opened 2017-05-03
Add IceConnection Interface

jcague:
**Description**

It adds a new C++ interface for IceConnection. It will help us using libraries others than libNice in the future. I also create an empty instantiation of a NicerConnection that also inherits from IceConnection. We will be able to enable it in the future by adding -DUSE_NICER to the builds.

Not needed.


erizo/src/erizo/IceConnection.h
```diff
@@ -0,0 +1,132 @@
...
+struct CandidatePair{
+  std::string erizoCandidateIp;
+  int erizoCandidatePort;
+  std::string clientCandidateIp;
+  int clientCandidatePort;
+  std::string erizoHostType;
+  std::string clientHostType;
+};
+
+class IceConfig {
```
jcague: @lodoyun can you review the new IceConfig? I added media_type transport_name, connection_id, ice_components, username and password and removed them from the NiceConnection constructor.

lodoyun: LGMT, at some point (not in this PR) we should probably refactor how we pass the configuration. Right now it comes directly from erizoAPI and is filled in progressively in several steps, there's probably a more consistent way to do it.

erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -10,7 +10,8 @@
 
 #include "./SrtpChannel.h"
 #include "rtp/RtpHeaders.h"
-#include "lib/LibNiceInterface.h"
+#include "./NiceConnection.h"
+#include "./NicerConnection.h"
```
kekkokk: this is funny

jcague: yep 😄 , I renamed NiceConnection to LibNiceConnection anyway to avoid confusions in the future, as Pedro suggested before.

PR #881, opened 2017-05-04
Remove Simulcast WIP message

jcague:
**Description**

Removed the alert/warning that shows up every time we publish a simulcast stream.

Not needed.


kekkokk:
can we consider simulcast "ready" also in Firefox?

jcague:
Yes, but only for subscribers, for publishers Firefox won't be supported for a while probably. I mean, publishers in Firefox won't be able to send simulcast streams yet.

PR #884, opened 2017-05-05
Add io thread + nicer init logic

jcague:
**Description**

Adds logic to initialize NicerConnection, and a IOWorker to read/write data from sockets and process ICE messages.

Not needed.


erizo_controller/erizoJS/erizoJS.js
```diff
@@ -79,7 +81,10 @@ var log = logger.getLogger('ErizoJS');
 var threadPool = new addon.ThreadPool(GLOBAL.config.erizo.numWorkers);
 threadPool.start();
 
-var ejsController = controller.ErizoJSController(threadPool);
+var ioThreadPool = new addon.IOThreadPool(GLOBAL.config.erizo.numIOWorkers);
+ioThreadPool.start();
```
lodoyun: We could potentially not start the `ioThreadPool` if we're not using nicer right?

scripts/licode_default.js
```diff
@@ -149,6 +149,9 @@ config.erizo = {};
 // Number of workers that will be used to handle WebRtcConnections
 config.erizo.numWorkers = 24;
 
+// Number of workers what will be used for IO (including ICE logic)
+config.erizo.numIOWorkers = 1;
```
lodoyun: Just a reminder that we should probably tweak this number

erizo/src/erizo/NicerConnection.cpp
```diff
@@ -26,74 +26,643 @@ extern "C" {
...
+                                                  flags,
+                                                  const_cast<char*>(ufrag_.c_str()),
+                                                  const_cast<char*>(upass_.c_str()), &ctx_);
+  if (r) {
+    ELOG_WARN("%s message: Couldn't create ICE ctx", toLog());
+    start_promise_.set_value();
+    return;
+  }
+
+  r = nicer_->IceContextSetTrickleCallback(ctx_, &NicerConnection::trickle_callback, this);
```
jcague: no, we don't. This is the only callback we have to get info from the local candidates. It's similar to `cb_new_candidate` in LibNice.

erizo_controller/erizoJS/erizoJS.js
jcague: I really like that idea, will change it.

kekkokk:
is nicer ready?

jcague:
only code completed, but not ready for production, we'll test it to look for bugs from now on during the next days/weeks

kekkokk:
Are there new infos about nicer? 
I tried to use it in local environment to test it and seems that firefox is more likely to work better with this but I encountered tons of erizojs timeouts.
also few crash in async_wait.c : 329 in nrappkit.
Then tried to increase IOworkers. this caused my code to not run at all.
one last thing. It seems that nrappkit is not ready for 64bit environment.

So my question is: why nicer? also, why the need of implementing a new ice library inside licode? there were bugs with libnicer we should be aware of? this maybe will help to isolate some strange behaviour.
Also, why not a new nicer version?

Thanks guys!

jcague:
we've been using it in production for the last 4 weeks with no remarkable issues in millions of subscribing minutes. Actually, we're really happy by having switched to nicer. We had only few crashes (less than 3), and we're running it in a 64bit environment using docker. Nicer apparently reduced the amount of connectivity issues we previously had (people suddenly disconnected, etc.).

there's no new nicer version because we haven't found any big issue with the current implementation.

btw, we can't use more than one IOWorker yet.

please, report any issue you find with nicer, because we want to make it the default ICE library in Licode. It would be good to fix them asap.

kekkokk:
I'm doing simple test with chrome 61 <-> firefox 56
one basicexample page opened in each browsers. The test consist in reloading firefox page.
with libnice:
 - sometimes firefox stuck with the spinning icon and no video is shown. it always publish btw.
 - never had a crash

with nicer:
- noticeably faster connection
- firefox always works.
- erizojs crashes a lot. I mean, it closes and send a stream-fail event. but it closes due to timeouts. Log attached:

[erizo_js_timeout.txt](https://github.com/lynckia/licode/files/1382254/erizo_js_timeout.txt)

I can reproduce by simply continue reloading firefox page. You will get it in under one minute of trying. this doesn't seems to happen with libnice tho.

Is there something that prevent to upgrade to latest libnice 0.1.14? seems more "maintained"
Also I saw that the "newest" piece of code inside nicer is belonging to 2013. but there are lots of rfcs since that day.

kekkokk:
We are currently testing libnice 0.1.14 and seems to solve all previous libnice problems. also the speed is improved a lot

PR #889, opened 2017-05-10
External input improvements at erizo controller level

zevarito:
Improvements to external input functionality.

**8da578b log specific error on external input error**
Output the specific error when adding external input fails.

**baf793d avoid race condition when removing publisher**
Wait for AMQP call to complete before force deletion of Erizo that causes the process not finish properly.

**9ea7fe4 add optional custom timeout for callRpc**
add optional Timeout param to `callRpc` function.

**2ceb3d5 increase timeout for ExternalInput callRpc call**
Start to consume a RTSP source could take more than the default 5000 timeout so this commit increase the wait to complete to 20000 for adding an external input call.

Changes in Server public API, see 9ea7fe4.


aalonsog:
Hi @zevarito, many thanks for your contribution. 

I agree with the part related with the External Output (logs, timeout, etc). However, I see a problem with the removePublisher modifications. With your changes, you are waiting for a callback to remove the publisher in roomController records. And, as you can see in erizoJSController, that callback is never called.

If you remove the removePublisher changes from this PR we could approve it. And then we could start discussing possible improvements in the callbacks management between roomController and erizoJSController. Probably that would be a more complex stuff that also involves other parts of erizoController.

zevarito:
Hi @aalonsog,

What do you mean is that `amqper.callRpc(erizoId, 'removePublisher', [publisherId], {callback: function ()}` callback might not be called in some scenarios so the publisher might not be released? I didn't want to touch `callRpc` to much except for the timeout thing but I strongly believe that it should offer the `callbackError` as-well so you can handle both scenarios and got your block executed always, but also know if the operation fails or not.

Anyway I agree that it should be separated in two PR so let me finish some stuff and I'll re-submit in the incoming days, thanks.

zevarito:
@aalonsog meanwhile, closing this PR and open this one: https://github.com/lynckia/licode/pull/990

PR #894, opened 2017-05-12
Add support for Electron screensharing

lodoyun:
**Description**
This PR adds support for Electron screensharing. Electron does not have to use an extension like Chrome does.

No changes in the API, it should work transparently.

### DIFF
--- a/erizo_controller/erizoClient/src/Connection.js
+++ b/erizo_controller/erizoClient/src/Connection.js
@@ -20,7 +20,7 @@ Erizo.Connection = function (spec) {
     } else if (that.browser === 'bowser'){
         L.Logger.debug('Bowser Stack');
         that = Erizo.BowserStack(spec);
-    } else if (that.browser === 'chrome-stable') {
+    } else if (that.browser === 'chrome-stable' || that.browser === 'electron') {
         L.Logger.debug('Chrome Stable Stack');
         that = Erizo.ChromeStableStack(spec);
     } else {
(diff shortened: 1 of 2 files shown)


erizo_controller/erizoClient/src/Connection.js
```diff
@@ -49,18 +49,22 @@ Erizo.getBrowser = function () {
...
         browser = 'bowser';
     } else if (window.navigator.userAgent.match('AppleWebKit') !== null) {
         browser = 'bowser';
     }
-    return browser;
+return browser;
 };
 
-
+var isElectron = function() {
```
jcague: I think we're not using this

erizo_controller/erizoClient/src/Connection.js
```diff
@@ -71,6 +75,15 @@ Erizo.GetUserMedia = function (config, callback, error) {
     if (config.screen) {
         L.Logger.debug('Screen access requested');
         switch (Erizo.getBrowser()) {
+            case 'electron' :
+                 L.Logger.debug('Screen sharing in Electron');
+                var screenConfig = {};
+                screenConfig.video = config.video || {};
+                screenConfig.video.mandatory = config.video.mandatory || {};
+                screenConfig.video.mandatory.chromeMediaSource = 'screen';
+                var theConfig = {video: { mandatory: { chromeMediaSource: 'screen' } } };
```
jcague: are we using this var? seems like we're building the same object twice but we just use one.

erizo_controller/erizoClient/src/Connection.js
```diff
@@ -49,18 +49,22 @@ Erizo.getBrowser = function () {
     } else if (window.navigator.userAgent.match('Bowser') !== null){
         browser = 'bowser';
     } else if (window.navigator.userAgent.match('Chrome') !== null) {
-        if (window.navigator.appVersion.match(/Chrome\/([\w\W]*?)\./)[1] >= 26) {
-            browser = 'chrome-stable';
+        browser = 'chrome-stable';
+        if (window.navigator.userAgent.match('Electron') !== null) {
```
lodoyun: Yep, it contains both Electron and Chrome.
Actually the original idea was to add it as another `else if` but that does not work because of that.

PR #899, opened 2017-05-24
stream-failed event on ice failed p2p

kekkokk:
**Description**
Throws an stream-failed event when ice fails during a p2p call.
It was unmanaged before.


kekkokk:
temp close

Issue #555, opened 2016-10-14
openssl DTLS  Handshake failed


xuhongming251:
hi , I use erizo.so + node js .

nodejs + node-gyp addon + erizo.so

I compile node js 0.10.47 shared link to openssl. and erizo.so use the same version openssl .

I have tested the openssl all the version 1.0.\* ，openssl-1.0.2a 、openssl-1.0.2e、openssl-1.0.2k、openssl-1.0.2l、openssl-1.0.2i、openssl-1.0.2j、openssl-1.0.2u。

but they do not work very well.
1、openssl-1.0.2u、openssl-1.0.2l：
 sometimes，it occurs SSL3_GET_CERT_VERIFY:bad ecdsa signature， so Handshake failed。
2、
openssl-1.0.2a 、openssl-1.0.2e
Handshake failed all the time.

erizo run env：
ubuntu 14.04 LTS

brower env：

google chrome 53.0.2785.116 (64-bit)

xuhongming251:
Handshake failed:
1、erizo create offer ，modify sdp a=sendrecv instead of a=recvonly.
2、exchange signnaling message

but this is ok:
1、browser create offer ，modify sdp a=sendrecv instead of a=sendonly.
2、exchange signnaling message

I donot why erizo crate offer  often a problem,but other peer offer first is ok.

jcague:
I'd recommend node v0.10.37 and openssl 1.0.1l. They work well in our machines.

xuhongming251:
I modify the licode offer sdp , and add a=setup:actpass. now all is ok . have resolved  dtls handshake failed  error , this error have not appear again , thx~!

kekkokk:
@jcague I am getting this error quite often with a clean install from actual master branch
`[erizo-636dc816-7fc7-a590-39a4-994bde409c8f] 2017-02-16 11:19:00,882  - WARN [0x7f91deffd700] DtlsTransport - id: iPrlVcOP8IScztlbk1dv_465085895674250300, type: subscriber,  message: Handshake failed, transportName:video, openSSLerror: Dtls Timeout on Resender`

this is happening especially in a many to many scenario, when the subscriber has lots of video to subscribe.
I reproduce the issue with a remote ec2 instance, ubuntu 14.04.5 LTS.
[...]

PR #901, opened 2017-05-25
Dtls handshake temp fix

kekkokk:
**Description**

This PR aims to resolve a strange problem that occurs when an user try to subscribe to lots of streams and the network is highly congested.
This cause the last flight of the DTLS handshake to be lost and the resender maybe don't handle it correctly. 
This cause a spinning icon to be visible for about 2 minutes before a stream failed is fired.

With this method, the maximum amount of time before a stream-failed is fired is ~5 seconds.

This is a TEMPORARY fix, as we might dig on why the resender of the last flight packet doesn't work at all as explained here: https://github.com/lynckia/licode/issues/555 


kekkokk:
As you wish! I'll give it a try anyway, 'cause since this mod we had nearly 0% failed streams in our many 2 many scenarios.
Or maybe add a tag and leave it open for those who wants to experiment.
Unfortunately I am not able to go deeper, I'm learning C with this project so complete refactoring a so delicate section is not for me. but studying the process It felt logical to make this changes.
4 months in production right now, no more customers : ('I can't see him').
switching to BoringSSL seems to be the preferred option anyway, seems browsers did the switch few time ago.

kekkokk:
temp close

PR #903, opened 2017-05-26
Disable force_layer and re-enable auto padding

kekkokk:
**Description**
It disable the previously forced layers of subscribed simulcast and re-enable the auto padding algo.

if subscribedStream._setQualityLayer('auto') is passed, it switch to auto.


kekkokk:
Ok, I like your observation. 
I split in two separate functions as you suggested but I didn't change nothing in the doc since is neither _setQualityLayer was mentioned. I think you would like to keep it private since it is well tested.

kekkokk:
It also would be nice to have a preference to set the preferred switch. 
For example I'd prefer to have always an higher temporal layer instead of spatial.
Id prefer to see smooth than neat.
You think it would be possible with low effort?

jcague:
agree, but that's a different API, that would be also nice to have limits on the max resolution and/or framerates each subscriber can receive.

kekkokk:
updated with new Stream.js

kekkokk:
!up please merge

PR #904, opened 2017-05-26
Fix when trying to publish an HD simulcast with only 2 layers set.

kekkokk:
When trying to publish an HD stream with only 2 layers set, due to a chrome implementation [Chromium Source](https://chromium.googlesource.com/external/webrtc/+/branch-heads/51/webrtc/media/engine/simulcast.cc#42), nothing is sent and results in a empty video on the subscriber side.

It alerts the publisher and automatically set layers to 3 (which is the minimum for an HD stream).


kekkokk:
updated with new Room.js

kekkokk:
!up please merge or close

erizo_controller/erizoClient/src/Room.js
```diff
@@ -587,6 +587,14 @@ Erizo.Room = (specInput) => {
 
     options.simulcast = options.simulcast || false;
 
+    if (stream.videoSize && stream.videoSize[0] > 640 && options.simulcast &&
+      options.simulcast.numSpatialLayers <= 2) {
+      L.Logger.warning('Trying to publish an HD stream with only 2 quality layers. Due to a Chrome ' +
```
jcague: a warning might be too high cause it's a bug in Chrome, that might be fixed at some point. I agree with the log, but I'd rather use L.logger.debug

jcague: Please, move the code to ChromeStableStack since it only affects Chrome.

jcague:
@kekkokk did you have some time to see the comments? what do you think?

kekkokk:
yeah sorry, I'll fix it asap

kekkokk:
!

erizo_controller/erizoClient/src/webrtc-stacks/ChromeStableStack.js
```diff
@@ -3,9 +3,18 @@ import SdpHelpers from './../utils/SdpHelpers';
 import Logger from '../utils/Logger';
 
 const ChromeStableStack = (specInput) => {
+  const newSpec = specInput;
   Logger.info('Starting Chrome stable stack', specInput);
-  const spec = specInput;
-  const that = BaseStack(specInput);
+  if (newSpec.videoSize[0] > 640 && newSpec.simulcast &&
```
jcague: what happens if newSpec.videoSize is undefined?

erizo_controller/erizoClient/src/webrtc-stacks/ChromeStableStack.js
```diff
@@ -3,9 +3,18 @@ import SdpHelpers from './../utils/SdpHelpers';
 import Logger from '../utils/Logger';
 
 const ChromeStableStack = (specInput) => {
+  const newSpec = specInput;
```
jcague: why not changing spec instead of creating newSpec and reassigning it?

PR #923, opened 2017-06-16
Fix erizoJS start and timeouts

jcague:
**Description**

Fix a timeout that happens sometimes when reloading a publisher.

Not needed.

### DIFF
--- a/erizo/src/erizo/DtlsTransport.cpp
+++ b/erizo/src/erizo/DtlsTransport.cpp
@@ -321,6 +321,15 @@ std::string DtlsTransport::getMyFingerprint() {
 }
 
 void DtlsTransport::updateIceState(IceState state, IceConnection *conn) {
+  std::weak_ptr<Transport> weak_transport = Transport::shared_from_this();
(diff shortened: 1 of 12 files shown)


kekkokk:
I noticed this behavior lots of time, I'll test and report.
thanks for this important fix

kekkokk:
hi guys, i'm experiencing lots of

> ErizoJS timed out will be removed, erizoId: 4c556da2-b216-c20d-5344-50d1fc4c456a, publishersAffected: 1

It happens especially when an user try to publish immediately after an unpublish (i made in basicexample a simple function "republish" that unpublish a stream and publish another one immediately after) and get an erizoJS timeout:

screen shot 2017-07-14 at 12 50 06 pm

it works, but when I click "republish" again this often happens:

screen shot 2017-07-14 at 12 47 30 pm

Like some structures aren't correctly cleared

jcague:
please don't use an old PR to submit an issue, just create a bug with your version, info, etc. Thanks!

PR #926, opened 2017-06-21
Fix dtls fragmented certificate

lodoyun:
**Description**

OpenSSL was using the wrong MTU producing fragmented certificates that we put together in a single packet anyway. This PR addresses this. 

Also, the logs have been updated to better reflect the progress in the DTLS handshake. Most _errors_ were not really _errors_.


erizo/src/erizo/dtls/DtlsClient.cpp
```diff
@@ -61,7 +61,7 @@ void SSLInfoCallback(const SSL* s, int where, int ret) {
     if (ret == 0) {
       ELOG_WARN2(sslLogger, "failed in %s", SSL_state_string_long(s));
     } else if (ret < 0) {
-      ELOG_WARN2(sslLogger, "error in %s", SSL_state_string_long(s));
+      ELOG_INFO2(sslLogger, "callback for %s", SSL_state_string_long(s));
```
jcague: nice!

PR #939, opened 2017-06-29
Refactor Room.js in Erizo Client

jcague:
**Description**

Refactor lots of lines within Room.js in Erizo Client to:
- Improve code readability.
- Reduce bugs.
- Make it easier to implement Single Peer Connection in a future.


mkhahani:
Nice work!
As a tip, using local variables is very useful for better source code minifications and also future refactoring:
```
const socket = that.socket;
socket.on(...);
...
```

PR #942, opened 2017-06-30
Fix wrong REMB detection with Chrome 60

### LINKED ISSUE #938: Simulcast doesn't work in Chrome >= 60
**Description**

Simulcast doesn't work on Chrome >= 60.

**Steps to reproduce the issue:**

1. Install vanilla Licode
2. open basicexample in whatever chrome version and publish a stream with ?simulcast=true
3. open basicexample in another computer with chrome >= 60 (from current beta) with ?onlysubscribe=true

**Describe the results you received:**

opening basicexample/quality_layers.html we can see that estimated bandwidth is 0kb

**Describe the results you expected:**

we should see the black bar of the estimated bandwidth growing.

**Additional information you deem important (e.g. issue happens only occasionally):**

seems that Licode doesn't receive any REMB. They are correctly negotiated according to sdps but no REMB is received (or maybe we don't parse them correctly? (I don't think so)). 

from SenderBandwidthEstimationHandler: 
` // TODO(pedro) Implement alternative when there are no REMBs `

**Licode commit/release where the issue is happening**

78599cd53f8e90a2314b12f40344e513f49f914d

**Additional environment details (Local, AWS, Docker, etc.):**

AWS EC2

jcague:
**Description**

Fixes #938 

Thanks to @kekkokk for the heads up. Turned out to be that Chrome is sending REMB packets within compound audio RR. And they can sometimes have a SSRC different from video's. [1]

Not needed.

[1] https://bugs.chromium.org/p/webrtc/issues/detail?id=7860

### DIFF
--- a/erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
+++ b/erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
@@ -57,7 +57,7 @@ void SenderBandwidthEstimationHandler::notifyUpdate() {
 
 void SenderBandwidthEstimationHandler::read(Context *ctx, std::shared_ptr<dataPacket> packet) {
   RtcpHeader *chead = reinterpret_cast<RtcpHeader*>(packet->data);
-  if (chead->isFeedback() && chead->getSourceSSRC() == connection_->getVideoSinkSSRC()) {
+  if (chead->isFeedback()) {
     char* packet_pointer = packet->data;
     int rtcp_length = 0;
     int total_length = 0;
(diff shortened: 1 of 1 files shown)


kekkokk:
WHOA!
thanks! At this purpose,  may I ask you:
I'm trying to make edge works, and with bowserStack it connect to Licode and do his things but he ignores REMBS, so the bw is always low and the video stuck at certain point. Is there any chance that Licode is sending him REMBS in a "wrong" or "not standard" way? 
I'm talking about a standard scenario, not simulcast.
Because I think we can make edge work with very low effort and we'll be the first in webrtc sfu scenario I think. tried other "competitors" last week and no one worked.

thanks for the fix anyways!

PR #945, opened 2017-07-03
Sender Bandwidth Estimation with no REMBS

kekkokk:
**Description**

This PR aims to manage the quality switch even when no rembs are received.
In particular:
This uses the same calls at the native webrtc/senderbwestimation on every RR received.
I used the publisher stats to caps the bw estimation, to avoid reaching skylimit that prevented the quick readjusting on packet loss or bandwidth fluctuation.

Nothing changes if used in conjunction with rembs, but I added an option to ignores them.
why? because I think that is more efficient in this way than with rembs.
I explain myself: rembs are generated client side based on what is received. but I noticed that there's a problem with the padding generator and high bandwith video:

SEE: https://github.com/lynckia/licode/issues/947

With rembs I was 90% 320x240 and sometimes it switched to 640x480. 

Testing with packet loss and limited bandwidth it performed waaay better.

I also tested the startbitrate set @ 1mbit. it doens't affect poor connections at all but help a lot those connection with no problem (it switch to high quality lot faster).

Please take some time to test it IRL and let me know what you think! :)

[x] It needs and DONT'T includes Unit Tests
Need help with unit tests


erizo/src/erizo/rtp/SenderBandwidthEstimationHandler.h
```diff
@@ -25,7 +25,7 @@ class SenderBandwidthEstimationHandler : public Handler,
 
  public:
   static const uint16_t kMaxSrListSize = 20;
-  static const uint32_t kStartSendBitrate = 300000;
+  static const uint32_t kStartSendBitrate = 1000000;
```
jcague: too high, I think to remember 300k is the same value Chrome uses inside by default, so there's no need to use a different default value. We might add a config value to change it.

erizo/src/erizo/rtp/SenderBandwidthEstimationHandler.h
```diff
@@ -57,6 +57,8 @@ class SenderBandwidthEstimationHandler : public Handler,
   bool initialized_;
   bool enabled_;
   bool received_remb_;
+  bool ignore_rembs_;
```
jcague: this is always false in the PR, am I right?

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
```diff
@@ -96,29 +99,54 @@ void SenderBandwidthEstimationHandler::read(Context *ctx, std::shared_ptr<dataPa
...
               if (!strncmp(uniqueId, "REMB", 4)) {
                 received_remb_ = true;
                 int64_t now_ms = ClockUtils::timePointToMs(clock_->now());
                 uint64_t bitrate = chead->getBrMantis() << chead->getBrExp();
                 ELOG_DEBUG("%s message: Updating Estimate with REMB, bitrate %lu", connection_->toLog(),
                     bitrate);
+                if (bitrate > current_max_bw_set_) {
+                  ELOG_DEBUG("%s message: Updating max bw based on REMB bitrate : %lu", connection_->toLog(), bitrate);
+                  current_max_bw_set_ = bitrate;
+                  sender_bwe_->SetMinMaxBitrate(0, bitrate);
```
jcague: what's the reasoning behind setting the max bitrate to the remb value?

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
```diff
@@ -96,29 +99,54 @@ void SenderBandwidthEstimationHandler::read(Context *ctx, std::shared_ptr<dataPa
...
                     connection_->toLog(), chead->getFractionLost(), delay, period_packets_sent_);
                 sender_bwe_->UpdateReceiverBlock(chead->getFractionLost(),
                     delay, period_packets_sent_, now_ms);
                 period_packets_sent_ = 0;
-                updateEstimate();
+
+                uint64_t max_BW = 0;
+                for (auto &spatial_layer_node : stats_->getNode()["qualityLayers"].getMap()) {
+                  for (auto &temporal_layer_node :
+                      stats_->getNode()["qualityLayers"][spatial_layer_node.first.c_str()].getMap()) {
```
jcague: IMO we're hacking the congestion control here... we're also capping the max bitrate value and it will go from the max layer bitrate to the remb when we receive it... I'm not sure if that's intended in webrtc's code.

erizo/src/erizo/rtp/SenderBandwidthEstimationHandler.h
kekkokk: yeah exactly

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
kekkokk: Yeah you are totally right, this is an absolute hack. 
I set the max value because if I don't, the bw estimation continues growing up to 1gbit, and at that point, if there's a fluctuation due to packet loss for example, it takes tons of seconds to go down to the accurate level.
I wasn't able to make the google code work correclty.
But in real life testing I think it performed perfectly. When rembs are present btw, the algo don't work and everything's work as you wrote it. 
It's a "temporary fix" for the calls in where rembs are not present. 
If you refuse the PR i totally understand, it's maybe not implemented in the correct way.
Again, if I don't set the max value (which I take on the publisher side), and call 
`
sender_bwe_->UpdateReceiverBlock(chead->getFractionLost(),
  | delay, period_packets_sent_, now_ms);
`
the estimation don't stop to grow

kekkokk:
Yeah you are totally right, this is an absolute hack.
I set the max value because if I don't, the bw estimation continues growing up to 1gbit, and at that point, if there's a fluctuation due to packet loss for example, it takes tons of seconds to go down to the accurate level.
I wasn't able to make the google code work correclty.
But in real life testing I think it performed perfectly. When rembs are present btw, the algo don't work and everything's work as you wrote it.
It's a "temporary fix" for the calls in where rembs are not present.
If you refuse the PR i totally understand, it's maybe not implemented in the correct way.
Again, if I don't set the max value (which I take on the publisher side), and call
`sender_bwe_->UpdateReceiverBlock(chead->getFractionLost(),   | delay, period_packets_sent_, now_ms);`
the estimation don't stop to grow

erizo/src/erizo/rtp/SenderBandwidthEstimationHandler.h
kekkokk: it takes less time to grow, but doesn't affect a poor bandwidth accordingly to my tests.

PR #949, opened 2017-07-05
Added location curl option

kiyoaki:
**Description**

Following command in [installUbuntuDeps.sh](https://github.com/lynckia/licode/blob/master/scripts/installUbuntuDeps.sh#L137-L138) returns "gzip: stdin: not in gzip format" because opus-1.1.tar.gz has been redirected to other location.

```
curl -O http://downloads.xiph.org/releases/opus/opus-1.1.tar.gz
tar -zxvf opus-1.1.tar.gz
```

**Changes**

I changed "curl -O http://xxx" to "curl -OL http://xxx"  in all installXXX.sh files.


jcague:
should fix #951

PR #973, opened 2017-07-19
Fix adaptation schemes

lodoyun:
**Description**
Adaptation Schemes were not working since we changed the Stats API. 
This PR changes adaptations schemes so they use the new promise API for `wrtc.getStats()` and the sender bitrate estimation instead of the raw REMB value.


erizo_controller/erizoJS/adapt_schemes/notify-break.js
```diff
@@ -1,111 +1,111 @@
...
+          break;
+          default:
+          log.error('Unknown BW status, id: ' + wrtc.wrtcId);
+        }
+        lastAverage = average;
+      }).catch((reason) => {
+        clearInterval(wrtc.monitorInterval);
+        delete wrtc.monitorInterval;
+        log.error('error getting stats: ' + reason);
+        return;
```
jcague: you don't need a `return` here, right? there are similar returns in the other schemes.

erizo_controller/erizoJS/adapt_schemes/notify-break.js
```diff
@@ -1,111 +1,111 @@
...
+              message: 'insufficient',
+              bandwidth: average});
             }
-            lastAverage = average;
-        }, INTERVAL_STATS);
-    };
+          }
+          break;
+          case BW_WONTRECOVER:
+          log.info('message: Switched to audio-only, ' +
```
jcague: wrong indentation in this switch clause

erizo_controller/erizoJS/adapt_schemes/notify-break.js
```diff
@@ -1,111 +1,111 @@
...
+          retries = 0;
+          average = 0;
+          lastAverage = 0;
+          wrtc.minVideoBW = false;
+          wrtc.setFeedbackReports (false, 1);
+          callback('callback', {type: 'bandwidthAlert',
+          message: 'audio-only',
+          bandwidth: average});
+          clearInterval(wrtc.monitorInterval);
+          delete wrtc.monitorInterval;
```
jcague: IMO we don't need to delete wrtc.monitorInterval every time we clear it, calling clearInterval twice with the same var is harmless.

PR #989, opened 2017-07-31
changing the option for generating source map to 'cheap-module-eval-s…

arsenicraghav:
Let's use this: 
'cheap-module-eval-source-map'
Reasons: 
File name displayed correctly
Line number displayed correctly 
The original source code is displayed

### DIFF
--- a/erizo_controller/erizoClient/webpack.config.erizo.js
+++ b/erizo_controller/erizoClient/webpack.config.erizo.js
@@ -9,5 +9,5 @@ module.exports = {
     library: 'Erizo',
     libraryTarget: 'var',
   },
-  devtool: 'source-map', // Default development sourcemap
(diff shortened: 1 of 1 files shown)


lodoyun:
`cheap-module-eval-source-map` embeds the sourcemap in the bundle file (`erizo.js`). We don't want that since we're running the bundled file through google ClosureCompiler for the production build.

You already have everything with `source-map` (line numbers, original, source, etc.), just make sure you are copying the `debug` version and not the `production` one. If you run `gulp` you will find the bundled Debug file and its source map in `erizo_controller/erizoClient/dist/debug/erizo`.

arsenicraghav:
@lodoyun when we use `source-map` it does not correctly display the file referenced by the error but when we use `cheap-module-eval-source-map` original source code is displayed. more here: 

http://cheng.logdown.com/posts/2016/03/25/679045

PR #990, opened 2017-07-31
External input improvements

zevarito:
Improvements to external input functionality.

**8da578b log specific error on external input error**
Output the specific error when adding external input fails.

**9ea7fe4 add optional custom timeout for callRpc**
add optional Timeout param to `callRpc` function.

**2ceb3d5 increase timeout for ExternalInput callRpc call**
Start to consume a RTSP source could take more than the default 5000 timeout so this commit increase the wait to complete to 20000 for adding an external input call.

Changes in Server public API, see 9ea7fe4.


zevarito:
@aalonsog totally forgot that 😄

PR #1003, opened 2017-08-10
Move IncomingStatsHandler to parse Incoming RTCP packets from subscriber

lodoyun:
**Description**
The `IncomingStatsHandler` was positioned in a way that all feedback RTCP traffic in a subscriber was already filtered (in simulcast). We're losing all the RTCP information in that case. 
By moving it below the `QualityFilterHandler` we get that information... There are, however other problems since we've already filtered out all the NACKs that we've successfully recovered in the `RtpRetransmissionHandler`. 

We probably need to rethink this but this is the minimal change that will get us some RTCP stats to display.

Please note that most of the stats gathered by both the `StatsHandlers` are purely informative except `totalCalculatedBitrate` which is not affected by this PR.

### DIFF
--- a/erizo/src/erizo/WebRtcConnection.cpp
+++ b/erizo/src/erizo/WebRtcConnection.cpp
@@ -276,10 +276,10 @@ void WebRtcConnection::initializePipeline() {
 
   pipeline_->addFront(LayerDetectorHandler());
   pipeline_->addFront(RtcpProcessorHandler());
-  pipeline_->addFront(IncomingStatsHandler());
   pipeline_->addFront(FecReceiverHandler());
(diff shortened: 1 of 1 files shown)


lodoyun:
This PR moves a Handler within our packet pipeline. It's aimed at solving a problem we have when accounting for stats in RTCP packets.

kekkokk:
I made some modifications to NACK generator handler. this PR could cause problems with that module?

lodoyun:
@kekkokk It should not affect it. This handlers does not affect packets in any way, it just gathers stats.

kekkokk:
thanks!!

PR #1020, opened 2017-08-30
Fix Spine to use different rooms

jcague:
**Description**

Fix the way we configure the room which is used by Spine to connect to publishers and subscribers.

Not needed.


aalonsog:
:+1:

PR #1026, opened 2017-09-14
LayerDetectorHandler reads also resolution and framerate

jcague:
**Description**

LayerDetectorHandler now reads frame width, height and rate for each Spatial and Temporal layer (VP9 and VP8). In another PR I will send this information to each subscriber to let them decide which layer they'll filter based on client's capabilities.

Not needed.


kekkokk:
very nice

erizo/src/erizo/rtp/LayerDetectorHandler.cpp
```diff
@@ -35,6 +46,28 @@ void LayerDetectorHandler::read(Context *ctx, std::shared_ptr<DataPacket> packet
...
+              temporal_layer, video_frame_rate_list_[temporal_layer].value());
+  }
+
+  // TODO(javier): send an event to subscribers with the new info
+
+  last_event_sent_ = clock_->now();
+}
+
+void LayerDetectorHandler::notifyLayerInfoChangedEventMaybe() {
+  if (clock_->now() - last_event_sent_ > std::chrono::seconds(5)) {
```
lodoyun: Consider making the 5 seconds a constant `kMinNotifyLayerInfoInterval` maybe?

PR #1028, opened 2017-09-14
Add options to erizo build process

lodoyun:
**Description**

This PR adds "Release" and "Debug" configurations for both erizo and erizoAPI. The main difference is that Erizo is now built with `-O3`. Both are built by default, while "Release" is the one that's normally used by ErizoJS.
In order to run the debug version of both erizo and erizoAPI, you can start erizoAgent with `initErizoAgent.sh -d` or, if you are running the docker image, you can add `--erizoDebug`.

Also, we're now using `rpath` when building erizoAPI, meaning that we no longer need `LD_LIBRARY_PATH` for Licode.

**Warning for Spine users**: You do need `LD_LIBRARY_PATH` and note that `liberizo` is now under `erizo/build/release/erizo` or `erizo/build/debug/erizo` .

Finally, in the Docker image we're keeping both builds but removing the object files in `CMakeFiles`to save some space.

### DIFF
--- a/erizo/src/erizo/media/ExternalInput.cpp
+++ b/erizo/src/erizo/media/ExternalInput.cpp
@@ -59,7 +59,8 @@ int ExternalInput::init() {
 
   // VideoCodecInfo info;
   MediaInfo om;
-  AVStream *st, *audio_st;
+  AVStream *st = nullptr;
+  AVStream *audio_st = nullptr;
 
   int streamNo = av_find_best_stream(context_, AVMEDIA_TYPE_VIDEO, -1, -1, NULL, 0);
   if (streamNo < 0) {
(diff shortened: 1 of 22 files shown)


extras/docker/initDockerLicode.sh
```diff
@@ -115,7 +119,11 @@ run_erizoController() {
 run_erizoAgent() {
   echo "Starting erizoAgent"
   cd $ROOT/erizo_controller/erizoAgent
-  node erizoAgent.js &
+  if [ "$ERIZODEBUG" = "true" ]; then
```
jcague: `==` or `-eq` ?

erizoAPI/binding.gyp
```diff
@@ -1,10 +1,38 @@
...
+              'GCC_ENABLE_CPP_RTTI': 'YES',              # -fno-rtti
+              'MACOSX_DEPLOYMENT_TARGET' : '10.11',      #from MAC OS 10.7
+              'OTHER_CFLAGS': [
+              '-g -O3 -stdlib=libc++ -std=c++11',
+            ]
+          },
+        }, { # OS!="mac"
+          'cflags!' : ['-fno-exceptions'],
+          'cflags' : ['-D__STDC_CONSTANT_MACROS'],
+          'cflags_cc' : ['-Wall', '-O3', '-g' , '-std=c++11', '-fexceptions'],
```
jcague: :+1:

erizo/src/erizo/media/ExternalInput.cpp
```diff
@@ -59,7 +59,8 @@ int ExternalInput::init() {
 
   // VideoCodecInfo info;
   MediaInfo om;
-  AVStream *st, *audio_st;
+  AVStream *st = nullptr;
```
jcague: good catch!

scripts/installErizo.sh
```diff
@@ -53,6 +55,9 @@ install_erizo(){
   cd $ROOT/erizo
   ./generateProject.sh
   ./buildProject.sh $FAST_MAKE
+  if [ $DELETE_OBJECT_FILES == 'true' ]; then
```
jcague: in other places we do `if [ "$DELETE_OBJECT_FILES" == "true" ];`, I didn't know it works like this too, great!

extras/docker/initDockerLicode.sh
lodoyun: We're using `=` throughout the whole file but `==` is more POSIX compliant so I'll switch all of them 👍

scripts/installErizo.sh
lodoyun: Even if it works, I'll switch to `"` for consistency 🙂

PR #1029, opened 2017-09-15
Send async events from pub to subs

jcague:
**Description**

Adds a way to send async events from pub to subs. That way we can safely notify about different events and they will be synced with the media data.

Not needed.

### DIFF
--- a/erizo/src/erizo/MediaDefinitions.h
+++ b/erizo/src/erizo/MediaDefinitions.h
@@ -77,6 +77,7 @@ class Monitor {
 class MediaEvent {
  public:
   MediaEvent() = default;
+  virtual ~MediaEvent() {}
   virtual std::string getType() const {
     return "event";
   }
--- a/erizo/src/erizo/MediaDefinitions.h
+++ b/erizo/src/erizo/MediaDefinitions.h
@@ -74,6 +74,16 @@ class Monitor {
     boost::mutex monitor_mutex_;
 };
 
+class MediaEvent {
+ public:
+  MediaEvent() = default;
+  virtual std::string getType() const {
+    return "event";
+  }
+};
+
+using MediaEventPtr = std::shared_ptr<MediaEvent>;
+
 class FeedbackSink {
  public:
     virtual ~FeedbackSink() {}
@@ -83,7 +93,6 @@ class FeedbackSink {
  private:
     virtual int deliverFeedback_(std::shared_ptr<DataPacket> data_packet) = 0;
 };
-
 
 class FeedbackSource {
  protected:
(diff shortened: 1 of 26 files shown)


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -441,6 +441,17 @@ int WebRtcConnection::deliverFeedback_(std::shared_ptr<DataPacket> fb_packet) {
   return fb_packet->length;
 }
 
+int WebRtcConnection::deliverEvent_(MediaEventPtr event) {
+  if (!sending_ || getCurrentState() != CONN_READY) {
```
lodoyun: I'm wondering if it would be useful to have events that are propagated to the pipeline even before the connection is ready

erizo/src/erizo/rtp/LayerDetectorHandler.h
```diff
@@ -16,6 +16,22 @@
 
 namespace erizo {
 
+class LayerInfoChangedEvent : public MediaEvent {
```
lodoyun: Maybe we should have all the MediaEvent definitions in the same place? Like `MediaEvents.h`. I see the scenario where we use 15 of these and have to scan through every handler header file to find the definition. It's just an opinion though, feel free to leave it as it is.

erizo/src/erizo/MediaDefinitions.h
```diff
@@ -74,6 +74,17 @@ class Monitor {
     boost::mutex monitor_mutex_;
 };
 
+class MediaEvent {
```
lodoyun: 👍  to inheritance for this

erizo/src/erizo/rtp/LayerDetectorHandler.cpp
```diff
@@ -53,13 +53,18 @@ void LayerDetectorHandler::notifyLayerInfoChangedEvent() {
...
   for (uint32_t temporal_layer = 0; temporal_layer < video_frame_rate_list_.size(); temporal_layer++) {
+    video_frame_rate_list.push_back(video_frame_rate_list_[temporal_layer].value());
     ELOG_DEBUG(" TEMPORAL LAYER (%u): %lu",
               temporal_layer, video_frame_rate_list_[temporal_layer].value());
   }
 
-  // TODO(javier): send an event to subscribers with the new info
-
+  if (connection_) {
+    connection_->notifyToEventSink(
```
lodoyun: 👍

erizo/src/erizo/WebRtcConnection.cpp
jcague: yep, makes sense

erizo/src/erizo/rtp/LayerDetectorHandler.h
jcague: I'd rather tied them to the classes that make use of them, but we can change it in the future if it's difficult to find them

PR #1033, opened 2017-09-18
Add erizoJS debug option in configuration

lodoyun:
**Description**
Added the option to load the debug ErizoAPI addon in the ErizoAgent section of `licode_config`


lodoyun:
Yes, the command line parameter overrides the configuration. My reasoning to keep both options is that it can be useful for one-time debug runs without running the risk of forgetting changing the configuration back.

PR #1040, opened 2017-09-21
adds stream attributes to streamEvent

akshayKMR:
**Description**
Updating stream attributes object of a stream, set its attributes to undefined instead.
This is because when creating the streamEvent Object, it doesn't pull the `attrs` from the spec property passed to it. Thus it gets set to undefined.


jcague:
Already merged in #1038

PR #1043, opened 2017-09-28
Add an option to force the client to use only TURN candidates

jcague:
**Description**

Add an option to force the client to use only TURN candidates

We can now force the client to use a TURN server when publishing/subscribing by setting the next parameter:
```js
room.publish(localStream, {forceTurn: true});
```

### DIFF
--- a/doc/client_api.md
+++ b/doc/client_api.md
@@ -389,6 +389,12 @@ room.addEventListener("stream-added", function(event) {
 room.publish(localStream, {maxVideoBW:300});
 ```
 
+We can also force the client to use a TURN server when publishing by setting the next parameter:
+
+```
+room.publish(localStream, {forceTurn: true});
+```
+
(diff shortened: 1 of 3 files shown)


kekkokk:
I suggest to automatically activate this feature after 2 subsequent publish/subscribe fail as it would be probably the only option left to make the call works.

jcague:
I don't think it's necessary, but we can consider it in the future if we detect multiple cases.

kekkokk:
I experienced an issue with 2 different customers where their firewall blocked the stun binding request if were made too closely (a sort of ddos protection I think).
I was stuck in a loop where a stream-failed event were fired and tried to re-subscribe. The only option for me was to force that mechanism :(

kekkokk:
Also, another interesting feature would be dynamically generate ice Servers without breaking the async structure of the process.
I mean, 99% of us is using a third party library to get turn credentials such as xirsys or twilio.
so everytime an user try to publish/subscribe a stream it would be nice to have a place where to put inside an async call to the preferred service which returns an ice config and make the process of publish/subscribe start

PR #1051, opened 2017-10-11
Add support to configure codecs per room

jcague:
**Description**

This PR adds support to configure codecs per room, instead of having the same codec configuration for all rooms in a Licode instance. The list of available configurations will be setup in the `rtp_media_config.js` file, and we can add/remove the proposed configurations in `rtp_media_config_default.js`.

The proposed configurations are:
```js
mediaConfig.codecConfigurations = {
  default: { rtpMappings: { vp8, opus }, extMappings },
  VP8_AND_OPUS: { rtpMappings: { vp8, opus }, extMappings },
  VP9_AND_OPUS: { rtpMappings: { vp9, opus }, extMappings },
  H264_AND_OPUS: { rtpMappings: { h264, opus }, extMappings },
};
```

But we could also add other examples like:
```js
  VP8_AND_PCMU: { rtpMappings: { vp8, pcmu }, extMappings },
```

Now we can specify which media configuration we want to use in a Room.

```js
N.API.createRoom(roomName, function(room) {
  console.log('Room created with id: ', room._id);
}, errorCallback, {mediaConfiguration: 'VP8_AND_OPUS'});
```

NOTE: If no mediaConfiguration is passed or it doesn't exist in the config file we will use the default one.

### DIFF
--- a/doc/server_api.md
+++ b/doc/server_api.md
@@ -32,6 +32,7 @@ A Room object has the following properties:
 - `Room.name`: the name of the room.
 - `Room._id`: a unique identifier for the room.
 - `Room.p2p` (optional): boolean that indicates if the room is a peer - to - peer room. In p2p rooms server side is only used for signalling.
+- `Room.mediaConfiguration` (optional): a string with the media configuration used for this room.
 - `Room.data` (optional): additional metadata for the room.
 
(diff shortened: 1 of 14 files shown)


mkhahani:
Great feature! Thank you.

PR #1056, opened 2017-10-17
Fix Erizo SDP parser

jcague:
**Description**

There was an issue parsing the SDP issued by old versions of Chrome.

Not needed.


jcague:
I included here the fix in #1023

PR #1058, opened 2017-10-20
Update libnice

Trisfald:
**Description**

This's a PR to update libnice to the revision: fbdccf0c2787ebdc65fe13ac64bd25c829ea7972 (from libnice mirror github), which is their current head.

We previously tried to link erizo against libnice 0.1.14 tag, but the latter works only in localhost. It seems there was a bug with srflx candidates that was discovered and fixed a year after the release.

On Ubuntu, glib is now built from source together with its dependency (libffi) and the necessary system-wide libraries are installed with apt. On Mac, brew is used to install system-wide dependencies.

It would be a great help to us if you could try this version of libnice with a couple different networks to see if there are issues.

### DIFF
--- a/erizo/src/CMakeLists.txt
+++ b/erizo/src/CMakeLists.txt
@@ -61,16 +61,22 @@ set(THIRD_PARTY_LIB "${CMAKE_CURRENT_SOURCE_DIR}/../../build/libdeps/build/lib/"
 ## Depencencies
 
 # GLIB
-find_package(Glib REQUIRED)
-include_directories(${GLIB_INCLUDE_DIRS})
+if (APPLE)
+    message("APPLE")
+    find_package(Glib REQUIRED)
+    set (GLIB2_LIBRARIES ${GLIB_LIBRARIES} gio-2.0 gobject-2.0 gthread-2.0)
+    include_directories(${GLIB_INCLUDE_DIRS})
+elseif(UNIX)
+    message("UNIX")
+    set(ENV{PKG_CONFIG_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/../../build/libdeps/build/lib/pkgconfig")
+    pkg_check_modules(GLIB2 glib-2.0>=2.44)
+    include_directories(${GLIB2_INCLUDE_DIRS})
+    set (GLIB2_LIBRARIES ${GLIB2_LIBRARIES} gio-2.0 gobject-2.0 gthread-2.0)
+endif()
 
 # BOOST
 set (BOOST_LIBS thread regex system)
 find_package(Boost COMPONENTS ${BOOST_LIBS} REQUIRED)
-
-# GTHREAD
-find_library(GTHREAD gthread-2.0 HINTS "${THIRD_PARTY_LIB}")
-test_lib(${GTHREAD})
 
 # SRTP
 find_library(SRTP srtp2 HINTS "${THIRD_PARTY_LIB}")
--- a/erizo/src/erizo/CMakeLists.txt
+++ b/erizo/src/erizo/CMakeLists.txt
@@ -21,4 +21,4 @@ add_library(erizo SHARED ${ERIZO_SOURCES})
 
 #message("Libs ${SRTP} ${NICE} ${GTHREAD} ${SSL} ${CRYPTO} ${LIBS} ${LOG}")
 
-target_link_libraries(erizo ${GLIB_LIBRARIES} ${Boost_LIBRARIES} ${SRTP} ${NICE} ${GTHREAD} ${SSL} ${CRYPTO} ${LIBS} ${LOG} webrtc nicer nrappkit)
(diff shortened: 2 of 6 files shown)


kekkokk:
2 weeks using this in production, works like a charm. 
please merge

jcague:
@Trisfald we're now more focused on using nICEr, actually we've been using it for the past 2 months without any big issues (there's currently a know bug with a crash that is happening not often in our servers, but nothing that is affecting the traffic a lot), and it decreased a lot the number of ICE disconnections during the sessions. It might scale better in the future due to not having any static mutex inside. Have you tried using nICEr instead of updating libnice in your case? we'd be happy to hear from your case (you can enable it inside licode_config.js

Trisfald:
We made some tests with nICEr and it worked fairly well: maybe better than libnice 0.1.4, but surely worse than newer libnice versions. The problems with nICEr are that it crashes with more than one IO worker and it doesn't support ICE restarts (libnice does have this functionality). 
Also, it seems like nICEr is an abandoned project. We couldn't find any repo with new patches/feature. On the other hand, libnice is well maintained and its owners work on adding new features from the latest ICE RFC.
In summary we feel libnice 0.1.14 runs a little better and more importantly it has an active community.

jcague:
libnice can use more than one thread, but the truth is that it actually has some static mutex that makes it much less scalable. Anyway, 1 IO Worker in nICEr performs equally than libnice (it doesn't make use of our IOWorkers).

nICEr is part of Firefox and is maintained by people in Mozilla, it  works really well and doesn't seem to be changed in a while because of that. Eric Rescorla implemented it, and added it to Firefox once he started to work for Mozilla, and he's also a big contributor in the WebRTC community, coauthoring RFC drafts like ICE Trickle. So we're pretty confident about the library's health. I haven't checked recently but I would say that nICEr also supports ICE restarts since Firefox does it too. The problem is that Licode doesn't use ICE restart yet (neither with libnice or nicer). Regarding IOWorkers, Firefox uses more than nICEr with more than one thread, so that's something we want to improve in Licode in the future.

Another reason that made us go for nICEr was to reduce the number and size of dependencies (such as glib, for instance).

Did you have any issue other than the number of IOWorkers? Or could you show some numbers or comparison? again, it's just because we wonder why there's such a difference between prod environments as it helped us reducing the issues we used to have. It works like a charm for us and if there's anything wrong for others that's something we definitively need to change. I'd fix those issues rather than updating libnice version.

Trisfald:
Actually we implemented ICE restarts in our codebase leveraging libnice's latest additions. nICEr IceConnection version of ICE restart is a no-op right now as we don't use it much. We'll be glad to PR this feature but libnice is a must have to make restart works.
As for library dependencies, libnice is indeed more heavyweight due to glib (and it get worse for 0.1.14 because it depends also on gnutls). Anyway, glib is already a dependency of the project, the only change is that we build the latest version instead of getting it from apt. Also, we don't see any cons updating a library in the project that is years old and is currently used as default.
In our tests libnice resulted way more robust and we haven't experienced a crash since new version is up!

Actually we can make nICEr crash easily opening two basicexample pages: one with publisher and one with ?onlySubscribe=true. If you reload the page on the subscriber only, probably it will crash on the second/third time you refresh the page; this cause also the publisher to timeout and the whole js closes itself.
We didn't analyze the crash yet (on osx it doesn't generate a core dump btw) because we are focusing on other things.

jcague:
in your test, are you setting `config.erizo.numIOWorkers = 1;`?

Trisfald:
Yes we are setting config.erizo.numIOWorkers = 1. We don't have a stack trace because erizojs is closed without a core dump; after reloading the page for a few times the call doesn't work correctly (the subscriber can't connect to the room) and then erizojs exits.

jcague:
ok, I thought Erizojs crashed, you might need to change the config file then since it seems like clients can't connect to Erizo.

kekkokk:
@jcague actually erizojs crashes (in fact also the publisher get dropped), is not a config problem.
Only problem is that no core dump are generated.

actually is easy to reproduce: 

> Actually we can make nICEr crash easily opening two basicexample pages: one with publisher and one with ?onlySubscribe=true. If you reload the page on the subscriber only, probably it will crash on the second/third time you refresh the page; this cause also the publisher to timeout and the whole js closes itself.

kekkokk:
Is there any way we can have this PR merged? 
The current Libnice implementation lacks of keepalive-conncheck which will be very useful (was introduced in the 0.1.8)
Also we are using this version since this PR was made and never encountered a problem.
Also the current libnice in licode ( 0.1.4 ) is from 2013. I think that if we want to keep the support to libnice, at least we should update it. I cannot see why we can't

jcague:
@kekkokk I are thinking of dropping support for libnice in the future, since nicer has been working really well in our production environment for more than 1 year now, and we would like to reduce dependencies (glib, for instance) in the future. I agree that we could merge this PR in the meantime but I encourage you to start using nicer, and keep sending us report of bugs if you still see any.

PR #1064, opened 2017-11-06
Fix nuve createroom

kekkokk:
**Description**

This PR fixes with the help of a queue the createroom method in nuve.
The bug manifest itself when creating 2 or more different rooms at the same time (only the last were added to services/room in mongodb.

Step to reproduce the issue without this fix:
open basicServer.js and call somewhere these 2 functions
`          
function testCreateRoom() {
  N.API.createRoom('test1', function(roomID) {},function error(){});
  N.API.createRoom('test2', function(roomID) {},function error(){});
}
`
go to your mongo instance and check rooms collection: OK, there are both.
now check your room object in your service object: FAIL, only test2 is present.

Why? because the service object is attached to request when doing the authenticate method.
so, both calls arrives to the server, the server authenticate then, attach to request the currentservice object and then both call createroom. they add their own room to the previosly obtained service object then call updateservice (which rewrite the service in mongo instead of taking in account differences from before and after.

with this method we simply update rooms inside without the risk of losing it

### DIFF
--- a/nuve/nuveAPI/mdb/serviceRegistry.js
+++ b/nuve/nuveAPI/mdb/serviceRegistry.js
@@ -55,9 +55,10 @@ exports.addService = function (service, callback) {
 /*
  * Updates a determined service in the data base.
  */
-exports.updateService = function (service) {
+exports.updateService = function (service, callback) {
     db.services.save(service, function (error) {
         if (error) log.info('message: updateService error, ' + logger.objectToLog(error));
+        if (callback) callback();
     });
 };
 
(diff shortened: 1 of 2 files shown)


jcague:
@kekkokk good catch! can you also change the line in https://github.com/lynckia/licode/pull/1064/files#diff-ecbfea703c30d24e338fd996e7eab8b8R37 please?

kekkokk:
need help with units tests

kekkokk:
!up

PR #1066, opened 2017-11-08
Gulp watch 60x faster

kekkokk:
Simply create new task in webpack which skips the google closure compiler when launching gulp watch.
Useful in development since every change we made it only takes couple of seconds to clean / bundle / dist
It also copy the debug version of licode client to basicexample instead of the production one.


kekkokk:
make sense, can you implement this mechanism in the way you prefer? also I noticed that when an error occurs in bundle_erizo task for example, the pipe break and this task is never done again on file change so the "watch" task becomes not so useful

kekkokk:
apparently this is needed:

>         .on('error', function(error) {
>             plugins.util.log(plugins.util.colors.red(error.message));
>             this.emit('end');
>         })

PR #1068, opened 2017-11-14
[FEATURE-REVIEW] Updated Single PeerConnection 

lodoyun:
**Description**
Updates the feature review for Single PeerConnection with some of the new ideas for the design.

### DIFF
--- a/feature-review/single-peer-connection.md
+++ b/feature-review/single-peer-connection.md
@@ -4,37 +4,37 @@ Single Peer Connection means that we can send/receive all streams in the same Pe
 ![comparison](https://docs.google.com/drawings/d/1EMW43_6BX0mDJ2v4kLWrzDobz6dRjjqhctSkLM869io/pub?w=960&h=721)
 
 ## Architecture
-Single Peer Connection will carry changes in both client and server sides.
+Single Peer Connection involves changes in both client and server sides.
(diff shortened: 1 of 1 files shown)


kekkokk:
It's not very clear to me how the single peerconnection is managed in terms of scalability. I mean, if there are 2 streams, on 2 different erizoJS, and I want to subscribe both in a single peerconnection, how this would work? 
Isn't something as erizo tree a prerequisite for this implementation? (also useful to maximize the scalability of licode)

lodoyun:
@kekkokk In short, you wouldn't be able to subscribe to both with the same PeerConnection. 
The idea here is that the amount of ErizoJSs available for each room would be controlled by the policy. There's a tradeoff there that you could tweak for your application. If you absolutely need everyone with single PeerConnection, you will restrict them to the same process and server. That may work perfectly, depending on your needs.
Erizo trees would allow another level of control but I don't see it a prerequisite.

kekkokk:
ok, makes sense

notedit:
it will reduce lots of port use. look forward this feature,  is there any timeline for this?

arsenicraghav:
@lodoyun Thank you updating the feature review document. Is this already implemented ? If not any timelines ?

PR #1069, opened 2017-11-17
Fixed screen sharing when no video param is included in config

aalonsog:
**Description**

This PR fixes screen sharing streams initialization when no video parameter is included.


yasskant:
Hi all,

I'm facing the same problem and I have the updated connection.js with the changes you did in your last commit but I still have the error "Cannot read property 'mandatory' of undefined"
Can you provide me with some further information?

Thanks in advance

PR #1075, opened 2017-11-27
Send SDP as JSON instead of String between clients and Erizo

jcague:
**Description**

Send SDP as JSON instead of String between clients and Erizo.
This PR aims to start getting rid of SDP mangling using strings. We'll start using a JSON object representing the SDP, which makes the SDP updates much easier.

Not needed.

### DIFF
--- a/erizoAPI/binding.gyp
+++ b/erizoAPI/binding.gyp
@@ -8,16 +8,14 @@
     'target_name': 'addon',
       'sources': ['<@(common_sources)'],
       'include_dirs' : ['<@(common_include_dirs)'],
-      'libraries': ['-L$(ERIZO_HOME)/build/release/erizo -lerizo -Wl,-rpath,./../../erizo/build/release/erizo'],
+      'libraries': ['-L$(ERIZO_HOME)/build/release/erizo -lerizo -Wl,-rpath,<(module_root_dir)/../erizo/build/release/erizo'],
       'conditions': [
         [ 'OS=="mac"', {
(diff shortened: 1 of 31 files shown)


erizo_controller/common/semanticSdp/SDPInfo.js
```diff
@@ -0,0 +1,823 @@
...
+        rids: [],
+      };
+
+      md.direction = Direction.toString(media.getDirection());
+
+      md.rtcpMux = 'rtcp-mux';
+
+      md.connection = media.getConnection();
+
+      md.xGoogleFlag = media.getXGoogleFlag('conference');
```
lodoyun: `getXGoogleFlag`  in MediaInfo does not take a parameter. If we need to check for the `conference` flag in particular, I'd implement that in this PR

erizoAPI/binding.gyp
```diff
@@ -8,16 +8,14 @@
     'target_name': 'addon',
       'sources': ['<@(common_sources)'],
       'include_dirs' : ['<@(common_include_dirs)'],
-      'libraries': ['-L$(ERIZO_HOME)/build/release/erizo -lerizo -Wl,-rpath,./../../erizo/build/release/erizo'],
+      'libraries': ['-L$(ERIZO_HOME)/build/release/erizo -lerizo -Wl,-rpath,<(module_root_dir)/../erizo/build/release/erizo'],
```
lodoyun: 👍  Good catch!

Issue #1076, opened 2017-11-30
SdpInfo.js wrong behavior in simulcast


kekkokk:
Hi, 
I'm facing some issue with firefox Simulcast.
I think that there's something handled in the wrong way in the new SDPInfo.js
https://github.com/lynckia/licode/pull/1075

Before this push, everything worked fine. We made few adjustments in c++ to make firefox simulcast works.
(Client side) I'm calling the .enableSimulcast() function in the SdpHelpers BEFORE creating the offer. Firefox then generates an offer with:

> a=simulcast: send rid=spam;egg
> a=ssrc:1099767458 cname:{299dbffc-d388-8644-b3d2-8974094ff9df}
> a=ssrc:1589187499 cname:{299dbffc-d388-8644-b3d2-8974094ff9df}

then setLocalDescForOffer is called and then the sdps flows into SemanticSdp.SDPInfo.processString:

    that.localSdp = SemanticSdp.SDPInfo.processString(that.localDesc.sdp);

The problem is that the 
> a=simulcast: send rid=spam;egg

directive disappear. 

[THIS](https://www.diffchecker.com/v3C92ZpJ) is the difference between the generated sdps from the browser (untouched) and the processed sdps from SdpInfo.
As you can see the a=simulcast get lost and there's also duplicate

> a=ssrc:1589187499 cname:{299dbffc-d388-8644-b3d2-8974094ff9df}
> a=ssrc:1589187499 cname:{299dbffc-d388-8644-b3d2-8974094ff9df}

jcague:
sorry, we don't support Firefox simulcast currently in the repository, so I can't help. But you have two options: you can find the bug and submit a patch to the SdpInfo or, even better ;), submit the code that makes Firefox simulcast work!

kekkokk:
I can push the code tomorrow with no problem, I submit everything I do (see the update to latest libnice).
But It will not work cause of the issue. If you can point me in the direction to comprehend why the a=simulcast line gets dropped, I can solve then PR.

jcague:
sure! probably I can fix it in the same Firefox Simulcast PR with the rest of the code if you prefer.

kekkokk:
Still, I see the duplicate 

> a=ssrc:1589187499 cname:{299dbffc-d388-8644-b3d2-8974094ff9df}
> a=ssrc:1589187499 cname:{299dbffc-d388-8644-b3d2-8974094ff9df}

cname as a bug

PR #1078, opened 2017-11-30
Firefox simulcast

kekkokk:
This PR intrduces the possibility to do simulcast also from Frirefox.
There is an unresolved bug that still have to be managed. 

Refer to https://github.com/lynckia/licode/issues/1076

It worked well before the Sdp JSON refactor

@jcague wait for your help!


kekkokk:
At this point firefox simulcast is working but there's a bug that inverts the spatial layer order, so che base (0) layer is 640x480 and the layer (1) us 320x240.
I think that maybe is due to a wrong sdp generation. In fact when SdpInfo process the offer created by browser, the result is a duplicate entry (not a bad thing consider the RFC allowing it)

> a=ssrc:3872236851 cname:{750bc0de-2f2f-454b-8bad-47c4654dc244}
> a=ssrc:2229183214 cname:{750bc0de-2f2f-454b-8bad-47c4654dc244}
> a=ssrc:2229183214 cname:{750bc0de-2f2f-454b-8bad-47c4654dc244}
> a=ssrc:2229183214 msid:{87566d06-d440-6c41-9ca6-82b7dcd20d01} {83919928-7494-d549-80d2-6cf793d632da}

I made it working commenting two lines in SDPInfo.js but I think this is not the right solution.

>     if (!track) {
>       track = new TrackInfo(media, trackId);
>       // track.setMediaId(md.mid);
>       // track.setEncodings(encodings);
>       stream.addTrack(track);
>     }

Interestingly, commenting these 2 lines the output now includes another msid mapping for the other simulcast track that previously was not present.

> a=ssrc:2729708755 cname:{2908d91d-ac91-584b-a736-bbb5e2674d66}
> a=ssrc:2729708755 msid:{1b6a9807-366f-f84c-a5dc-85cee9be48ee} {66011479-01dc-de4a-a37a-fd35435d187b}
> a=ssrc:3385189830 cname:{2908d91d-ac91-584b-a736-bbb5e2674d66}
> a=ssrc:3385189830 msid:{1b6a9807-366f-f84c-a5dc-85cee9be48ee} {66011479-01dc-de4a-a37a-fd35435d187b}
> a=ssrc:3385189830 cname:{2908d91d-ac91-584b-a736-bbb5e2674d66}
> a=ssrc:3385189830 msid:{1b6a9807-366f-f84c-a5dc-85cee9be48ee} {66011479-01dc-de4a-a37a-fd35435d187b}

Hope this helps the debug process, I'm stuck.
But why this library was introduced? I think it may be another point of failure debugging new browsers such as edge and safari since sdps are in a way "manipulated"

jcague:
@kekkokk I updated this PR, can you check whether it works or not for you now please?

kekkokk:
Simulcast and layers seems to work finally!
But I noticed some strange behaviour due to p(probably) the padding generation algorythm.
It is extremely slow to reach the upper layer (also in localhost) and sometimes it never happens.
This is the chrome receiving side:

screen shot 2017-12-12 at 9 54 26 am

Seems to work better with firefox as subscriber. A bit slow to reach the upper layer but is more consistent

erizo_controller/erizoClient/src/webrtc-stacks/FirefoxStack.js
```diff
@@ -1,36 +1,54 @@
 import Logger from '../utils/Logger';
 import BaseStack from './BaseStack';
 
+const possibleLayers = [
```
jcague: @kekkokk I've changed the previous parameters to have these ones, which is quite similar to what we get from Chrome's Simulcast

kekkokk: seems much better

jcague: who can't reach the resolution, the sender side publishing 640x480 or the receiver subscribing to it?

kekkokk: the receiver subscribing it will never receive the full 640x480 res because by default we set 2 layers and we push into the array the scaleresolution by 2 and 3.

jcague: that's probably because there's no bw limit now, but you can still set a global max bw with `maxVideoBW` parameter when publishing.

jcague: sorry, I read it wrong, it might be a bug in the way I add it to the parameters.

jcague:
@kekkokk I updated the code and it now uses the higher layers by default (640x480 por higher spatial layer and 320x240 for lower spatial layer).

PR #1083, opened 2017-12-06
fix temporary mutex and data race in NicerConnection

Trisfald:
**Description**

There are some mutex in ICE implementation classes that probably aren't doing their intended task, because they are instantly constructed and destructed (by the way, it seems to me these locks aren't really needed, maybe its better to remove them?).
This PR fixes this oversight and fixes also a bug in NicerConnection: there's a race between the asynchronous closeSync call and the synchronous one that is done if the former times out.


jcague:
wow! good catch! I don't see why we need `closeSync_mutex_` though, shouldn't we use just `close_mutex_` instead?

erizo/src/erizo/NicerConnection.h
```diff
@@ -105,6 +105,8 @@ class NicerConnection : public IceConnection, public std::enable_shared_from_thi
   nr_ice_handler* ice_handler_;
   std::promise<void> close_promise_;
   std::promise<void> start_promise_;
+  boost::mutex close_mutex_;
+  boost::mutex closeSync_mutex_;
```
jcague: please consider using snake_case here: `close_sync_mutex_`

Trisfald:
Glad to be of help!

erizo/src/erizo/NicerConnection.h
jcague: @Trisfald The only thing we need to merge this PR is this change, could you apply it please? thanks a lot!

Trisfald: Done! By bad, I don't know how I missed your prior comment

Issue #1079, opened 2017-12-01
Simulcast no longer works


kekkokk:
After PR https://github.com/lynckia/licode/pull/1075 simulcast in chrome no longer works.

**Steps to reproduce the issue:**
1. join with Chrome tab in http://localhost:3001/?simulcast=true
2. join with another Chrome tab in http://localhost:3001/?onlySubscribe=true
3. open http://localhost:3001/quality_layers.html 

**Describe the results you received:**
The quality remain in the low layer

![screen shot 2017-12-01 at 7 00 06 pm](https://user-images.githubusercontent.com/15072723/33496273-e1a0240c-d6c9-11e7-8c35-d17483329f7f.png)

**Describe the results you expected:**
Works as expected

**Additional information you deem important (e.g. issue happens only occasionally):**
Tried to also force the quality layer with remotestram._setStaticQualityLayer(1,2)

the command is ignored, still receiving low spatial layer.

**Licode commit/release where the issue is happening**
e6cd88b32e008ec1342cfd6024b83a506a2caebe

**Additional environment details (Local, AWS, Docker, etc.):**
tried in localhost (OSX).
the commit before, works as expected

lodoyun:
Thanks for the details and for your time testing different commits.
We're aware of the issue, it does look like there is a problem with the new generated SDP and it does not play nicely with our simulcast implementation.
We should have a solution soon.

ChukwudiUzoma:
@lodoyun Please how do I modify the spatial and temporal layer in Chrome developer console?
room.remoteStreams.get(id)._setStaticQualityLayer(spatialLayer, temporalLayer)  doesnt work. Thanks

lodoyun:
@ChukwudiUzoma It should work. Check that simulcast is working and you're not using either v4 or the latest code

ChukwudiUzoma:
@lodoyun Thanks mate, fixed it. It was just a browser problem.

PR #1089, opened 2017-12-11
Update the media port in the SDP parser

jcague:
**Description**

There is something wrong in the current SDP parser that is affecting Simulcast and some feedback mechanisms. I've compared the SDPs generated by the previous solution and the current one and this is the only thing that is different, and that could be causing those issues. Not sure 100% if there is any other cause.

There was another issue caused by sending a=ssrc and a=ssrc-group in a different order compared with the previous SDPs. SdpInfo.cpp is not perfect and expected to receive a=ssrc-group first, which is not happening now. I added a temporal fix given that we'll eventually get rid of SdpInfo.cpp with the new parser.

Not needed.


jcague:
this might fix #1079

kekkokk:
confirm it fixes #1079

Issue #1091, opened 2017-12-12
Dangling pointer in logging pattern


Trisfald:
There's a potential issue with a pattern commonly used to print logs. For example in MediaStream.h:

```
inline const char* toLog() {
  return ("id: " + stream_id_ + ", " + printLogContext()).c_str();
}
```

This creates a temporary std::string and returns the pointer to its internal buffer. The pointer is later used by the logging macro, but it outlives the lifetime of the string leading to undefined behavior.

I suggest changing the function to:

```
inline std::string toLog() {
  return ("id: " + stream_id_ + ", " + printLogContext());
}
```

and usage to:

```
const auto log_str = toLog();
ELOG_XXX("%s", log_str.c_str());
```

in MediaStream.h, IceConnection.h, Transport.h, WebrtcConnection.h

If you also think this's a problem, I can do a PR to fix!

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Licode commit/release where the issue is happening**
Master

**Additional environment details (Local, AWS, Docker, etc.):**

lodoyun:
Agreed, that looks like a more sound solution 👍 
Please go ahead and do the PR if you have the time for it.

Thanks!

PR #1093, opened 2017-12-12
Fix for dangling pointer in logging

Trisfald:
**Description**
This PR addresses the issue [1091](https://github.com/lynckia/licode/issues/1091).
I changed the return value of toLog() from const char* to std::string.

No changes


Trisfald:
I understand your point - with 2 lines code can get messy pretty fast.
I came up with a possible solution, here there's an [example](https://gist.github.com/Trisfald/0fdc1ecc881f7730428a00c4fd0d35e0).  You can run it with an online compiler (I don't know any fiddle equivalent for c++)

It's about passing strings as they are and using a templated helper to extract the c_str. Tell me if this can be a move in the right direction :)

lodoyun:
Yep, that definitely looks like the best solution.
Do you anticipate any significant impact in performance? I don't particularly see any problem but just in case.

Trisfald:
Shouldn't be any performance implication since the bulk of the logic is executed at compile time.

Trisfald:
Got a working solution that doesn't need any change on the codebase. Log macros can now take std::string as inputs.
The nice things is that by further specializing LogElementForwarder, one can log any custom type with '%s'.

Trisfald:
Yes, that is a good idea! LogElementForwarder should always be optimized away by the compiler, but it's better to have smarter code nonetheless

erizo/src/erizo/logger.h
```diff
@@ -69,68 +70,108 @@ log4cxx::LoggerPtr namespace::logger = log4cxx::Logger::getLogger(logName);
...
 
+DEFINE_ELOG_T(ELOG_TRACET, ELOG_TRACE2)
+DEFINE_ELOG_T(ELOG_DEBUGT, ELOG_DEBUG2)
+DEFINE_ELOG_T(ELOG_INFOT, ELOG_INFO2)
+DEFINE_ELOG_T(ELOG_WARNT, ELOG_WARN2)
+DEFINE_ELOG_T(ELOG_ERRORT, ELOG_ERROR2)
+DEFINE_ELOG_T(ELOG_FATALT, ELOG_FATAL2)
 
+// older versions of log4cxx don't support tracing
+#ifdef LOG4CXX_TRACE
```
jcague: good catch!

erizo/src/erizo/MediaStream.cpp
```diff
@@ -252,7 +252,7 @@ void MediaStream::read(std::shared_ptr<DataPacket> packet) {
   // PROCESS RTCP
   RtpHeader *head = reinterpret_cast<RtpHeader*> (buf);
   RtcpHeader *chead = reinterpret_cast<RtcpHeader*> (buf);
-  uint32_t recvSSRC;
+  uint32_t recvSSRC = 0;
```
jcague: :+1:

PR #1094, opened 2017-12-14
Depacketizer component

Trisfald:
**Description**
This is a component receiving as input raw video packets and outputting complete frames. 
Vp8 implementation uses the same logic as before, while the H264 one is new. Vp9 it not yet done.

We have used the depacketizer to better handle the recording and to add some custom features in a generic way. I think Licode could use this to decouple the frame building logic from ExternalOutput and support different codecs.

Note that the depacketizer can throw; a trivial change is required if one wants to avoid exceptions.

[X] It needs and includes Unit Tests (needs tests for H264)


erizo/src/test/utils/Tools.h
```diff
@@ -70,7 +74,7 @@ class PacketTools {
     sender_report->setPacketType(RTCP_Sender_PT);
     sender_report->setBlockCount(1);
     sender_report->setSSRC(ssrc);
-    sender_report->setLength(4);
+    sender_report->setLength(6);
```
lodoyun: Nice catch

PR #1095, opened 2017-12-15
Stop compiling client's code when running gulp watch

jcague:
**Description**

It stops compiling Erizo Client JS files when running `gulp watch`.
This also fixes #1066 

Not needed.


jcague:
@kekkokk this PR is similar to #1066, the main difference in functionality I think is that it works for both erizo and erizofc and that we still run the `lint`task

PR #1099, opened 2017-12-19
Fix missing generated nack

Trisfald:
**Description**
I run into this bug while looking at the nack generator (related to pr #1096 ). Something was off in the sequence of nacks generated and then I found that the first nack of the second blp was always missing.

The for inside the test case works but maybe the form can be improved

### DIFF
--- a/erizo/src/erizo/rtp/RtcpNackGenerator.cpp
+++ b/erizo/src/erizo/rtp/RtcpNackGenerator.cpp
@@ -93,6 +93,7 @@ bool RtcpNackGenerator::addNackPacketToRr(std::shared_ptr<DataPacket> rr_packet)
       uint16_t distance = blp_nack_info.seq_num - pid -1;
       if (distance <= 15) {
         if (!isTimeToRetransmit(blp_nack_info, now_ms)) {
+          index++;
           continue;
         }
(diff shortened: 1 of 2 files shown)


erizo/src/erizo/rtp/RtcpNackGenerator.cpp
```diff
@@ -89,8 +89,7 @@ bool RtcpNackGenerator::addNackPacketToRr(std::shared_ptr<DataPacket> rr_packet)
     base_nack_info.sent_time = now_ms;
     base_nack_info.retransmits++;
     while (index + 1u < nack_info_list_.size()) {
-      index++;
-      NackInfo& blp_nack_info = nack_info_list_[index];
+      NackInfo& blp_nack_info = nack_info_list_[index + 1];
       uint16_t distance = blp_nack_info.seq_num - pid -1;
       if (distance <= 15) {
         if (!isTimeToRetransmit(blp_nack_info, now_ms)) {
```
lodoyun: We would need another `index++` here. Otherwise we'd be stuck testing for the same nack_info until the time is right to send it.

Trisfald: Didn't see that at all, thanks!

PR #1098, opened 2017-12-19
Use strings to send the SDP within signalling messages.

jcague:
**Description**

For v5 we'll still send SDPs as strings instead of format them in JSON objects. 

[] It needs and includes Unit Tests

**Changes in Client or Server public APIs**

Not needed.

[] It includes documentation for these changes in `/doc`.

PR #1100, opened 2017-12-19
Firefox Simulcast parser

kekkokk:
Implemented a proper parser for the old implementation of Simulcast (such the one in Firefox).
Simulcast on Firefox was working only thanks to a coincidence.

Fix simulcast not working on https://github.com/lynckia/licode/pull/1098


erizo_controller/common/semanticSdp/SDPInfo.js
```diff
@@ -411,6 +413,12 @@ class SDPInfo {
         }
       }
 
+      if (simulcast03) {
```
jcague: good catch!

PR #1105, opened 2017-12-22
 Generate SDPInfo in ErizoJS

jcague:
**Description**

This is the first step to remove the SDP parser in Erizo (C++ side) and do it in ErizoJS.

Not needed.


arsenicraghav:
Great initiative! it would be even better we can have something like this https://github.com/clux/sdp-transform in Licode.

kekkokk:
actually there is exactly that library @arsenicraghav

jcague:
@lodoyun I merge it now, but you can review it later and I'll apply any change in future PRs.

PR #1107, opened 2017-12-29
MediaStreams now use their own worker

lodoyun:
**Description**
Before this PR MediaStreams shared a worker with the corresponding WebRTCConnection. This PR removes that limitation, instantiating MediaStreams with a different working and properly passing the packet path tasks between the workers.


Trisfald:
Hello Pedro! 
In my opinion the problem here is in the order of the pipeline components. LayerDetectorHandler sets the is_keyframe value but, during pipeline->read(), PliPacerHandler is visited before LayerDetectorHandler, so is_keyframe is always false. Maybe the solution is to move up PliPacerHandler? You surely know better if there are cons in doing that.
It has worked until now because before you didn't make a copy of the incoming packet; PliPacerHandler was reading the keyframe value of the packet received before the current one.

lodoyun:
@Trisfald You were right. There was an issue in DtlsTransport that could cause the pipeline to work with outdated packets, that is what was hiding the pipeline order problem.
Both should be fixed with the new commit. Thanks!

erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -178,9 +178,7 @@ void DtlsTransport::onIceData(packetPtr packet) {
     }
     return;
   } else if (this->getTransportState() == TRANSPORT_READY) {
-    unprotect_packet_->length = len;
```
jcague: could unprotect_packet_ be local now?

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -68,6 +68,13 @@ WebRtcConnection::~WebRtcConnection() {
   ELOG_DEBUG("%s message: Destructor ended", toLog());
 }
 
+void WebRtcConnection::close() {
```
jcague: why do you need to change the location? it might conflict with an ongoing PR ;)

erizo/src/erizo/DtlsTransport.cpp
lodoyun: 👍

erizo/src/erizo/WebRtcConnection.cpp
lodoyun: No particular reason, it liked it better that way. I switched it back.

PR #1114, opened 2018-01-08
Fix async function in NicerConnection

jcague:
**Description**

We found a bunch of issues in NicerConnection that could be caused by a wrong smart pointer handling in NicerConnection::async function.

Not needed.

### DIFF
--- a/erizo/src/erizo/NicerConnection.cpp
+++ b/erizo/src/erizo/NicerConnection.cpp
@@ -153,19 +153,18 @@ NicerConnection::~NicerConnection() {
   close();
 }
 
-void NicerConnection::async(function<void()> f) {
+void NicerConnection::async(function<void(std::shared_ptr<NicerConnection>)> f) {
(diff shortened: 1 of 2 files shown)


kekkokk:
I need to tell you that even with this update I find Nicer very unstable. As I said in another issue, to verify my theory simply open one publisher and one page with ?onlySubscribe=true.

cmd+r (reloading) the subscriber page makes erizoJs to crash very easily (sometimes even at first try)

I'll consider reviewing our libnice update PR since is way more reliable and could make ICE restart mechanism possible.

jcague:
we keep investigating the issues you mention, but it doesn't happen to us. We have been using nicer with millions of minutes and saw the same amount of issues we had before with libnice, but with much more stable connections instead (not having unexpected ICE fails). As I said before, ICE restart is also possible with nICEr.
I tried to reproduce it several times with ?onlySubscribe=true and it didn't crash. But I'm curious about any possible difference between our setups that could make it crash in your case. Please, send me an email and we may meet to check possible reasons.

PR #1118, opened 2018-01-10
Fix a crash that still happens when closing NicerConnection

jcague:
**Description**

There's still some cases where NicerConnection makes Erizo crash when closing a connection. This PR aims to solve them by destroying things in the correct threads.

Not needed.

### DIFF
--- a/erizo/src/erizo/DtlsTransport.cpp
+++ b/erizo/src/erizo/DtlsTransport.cpp
@@ -110,9 +110,9 @@ DtlsTransport::DtlsTransport(MediaType med, const std::string &transport_name, c
     iceConfig_.username = username;
     iceConfig_.password = password;
     if (iceConfig_.use_nicer) {
-      ice_ = NicerConnection::create(io_worker_, this, iceConfig_);
+      ice_ = NicerConnection::create(io_worker_, iceConfig_);
     } else {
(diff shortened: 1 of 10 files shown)


erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -128,6 +128,7 @@ DtlsTransport::~DtlsTransport() {
 }
 
 void DtlsTransport::start() {
+  ice_->setIceListener(shared_from_this());
```
lodoyun: 👍

kekkokk:
I don't want to bother you since I'm using happily libnice but I tried this pr and still don't solve the problem I'm facing with the subscriber reload.
I've emailed you at the email found on your github profile @minervaproject.
If you have another preferred mail to reach you @jcague please contact me at francesco.durighetto@bandyer.com

jcague:
thanks for the info @kekkokk ! I don't think this PR will solve the issue you mention because yours doesn't generate any core dump, we're pursuing some new crashes we're seeing in production now. I sent you an answer just a couple of minutes ago.

kekkokk:
[UPDATE]
My problem seems appear only on osx host. In my ubuntu 14.04 LTS machine everything's fine.
To make sure I cloned the project from 0 and changed nothing but still crashes on osx.

PR #1124, opened 2018-01-12
Integrate depacketizer in ExternalOuput adding h264

lodoyun:
**Description**
This PR integrates the new `Depacketizer` into `ExternalOutput` replacing all the old vp8 depacketizer login. As part of the process of integrating it, this PR updates the `Depacketizer` to better suit the rest of Erizo naming conventions and removing the `throw` code.  
This PR also adds support for H264 in recording new that we have a compatible depacketizer, that required a minor change in `LayerDetectorHandler` to start marking the spatial layer of h264 packets (for now it's always 0).

### DIFF
--- a/erizo/src/erizo/media/Depacketizer.cpp
+++ b/erizo/src/erizo/media/Depacketizer.cpp
@@ -2,31 +2,34 @@
 #include <cstring>
 
 namespace erizo {
-void Depacketizer::reset_buffer() {
+DEFINE_LOGGER(Depacketizer, "media.Depacketizer");
+
+void Depacketizer::resetBuffer() {
   buffer_ptr_ = buffer_;
 }
 
 void Depacketizer::reset() {
-  reset_impl();
-  reset_buffer();
+  resetImpl();
+  resetBuffer();
 }
 
-void Depacketizer::buffer_check(const int len) {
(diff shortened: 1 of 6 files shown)


nvazquezg:
@lodoyun This PR allows the creation of recordings using H264 codec? Or the output is still VP8 only?

lodoyun:
@nvazquezg Yes, H.264 recording is supported for H.264 rooms, meaning Licode won't transcode when recording, it will record in the codec that the room is using.
Also, keep in mind the result is still a `.mkv` file so you will have to postprocess it to use any other format.

nvazquezg:
@lodoyun Understood, thanks. To configure a room as H.264 I've set:
 `{ "mediaConfiguration": "H264_AND_OPUS" }`.
Is there anything else I have to do? Because if I try to record in a room configured this way, I get the following error:

`ERROR [0x7fdd31ffb700] media.ExternalOutput - Could not find video codec`

I traced it to here: https://github.com/lynckia/licode/blob/bc4e67ca0ab5230ae27d8311633f43e9e9995fbb/erizo/src/erizo/media/ExternalOutput.cpp#L368

Please, tell me if I am doing something wrong or if I need to file an issue.

nvazquezg:
@yannistseng that did the trick. Thank you very much! :+1:

PR #1127, opened 2018-01-15
Add full simulcast support to recordings

jcague:
**Description**

It adds a new worker + pipeline to ExternalOutput, so that we can handle simulcast layers and filter them. Now we record the higher available layer, but we might want to create an API in the future to record the lower or any intermediate layers instead.

Not needed.

### DIFF
--- a/erizo/src/erizo/media/ExternalOutput.h
+++ b/erizo/src/erizo/media/ExternalOutput.h
@@ -59,12 +59,14 @@ class ExternalOutput : public MediaSink, public RawDataReceiver, public Feedback
 
   void notifyUpdateToHandlers() override;
 
+  bool isRecording() { return recording_; }
+
  private:
   std::shared_ptr<Worker> worker_;
   Pipeline::Ptr pipeline_;
   std::unique_ptr<webrtc::UlpfecReceiver> fec_receiver_;
   RtpPacketQueue audio_queue_, video_queue_;
-  bool recording_, inited_;
+  std::atomic<bool> recording_, inited_;
   boost::mutex mtx_;  // a mutex we use to signal our writer thread that data is waiting.
   boost::thread thread_;
   boost::condition_variable cond_;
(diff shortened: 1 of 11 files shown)


erizoAPI/ExternalOutput.cc
```diff
@@ -22,6 +23,9 @@ class AsyncCloser : public Nan::AsyncWorker {
     ~AsyncCloser() {}
     void Execute() {
       external_output_->close();
+      while (external_output_->isRecording()) {
```
lodoyun: `isRecording` being false does not imply the `ExternalOutput` has finished closing. Consider moving the switch of `recording_` to false at the end of `syncClose`

erizo/src/erizo/pipeline/HandlerManager.h
```diff
@@ -0,0 +1,39 @@
+/*
+ *  Copyright (c) 2016, Facebook, Inc.
```
lodoyun: Correct me if I'm wrong, but didn't you come up with this class? We don't need to put the license here in that case

erizo/src/erizo/media/ExternalOutput.h
```diff
@@ -93,7 +109,7 @@ class ExternalOutput : public MediaSink, public RawDataReceiver, public Feedback
   // Note: VP8 purportedly has two packetization schemes; per-frame and per-partition.  A frame is
   // composed of one or more partitions.  However, we don't seem to be sent anything but partition 0
   // so the second scheme seems not applicable.  Too bad.
-  vp8SearchState video_search_state_;
+  // vp8SearchState video_search_state_;
```
lodoyun: Consider just removing this. It's a leftover from a previous PR.

erizo/src/erizo/pipeline/HandlerManager.h
jcague: you got it

erizoAPI/ExternalOutput.cc
jcague: good catch!

PR #1138, opened 2018-01-25
Refactored connection and pass erizoJSId to clients

lodoyun:
**Description**
This PR refactors `Connection` in `ErizoClient` and passes the `erizoJsId` from `erizoController` to `ErizoClient` to associate a given `Connection` to the erizoJs endpoint.

There is a new structure, `ConnectionManager` that is able to create Connections and keeps track of the `Connection`,`ErizoJS` pair.
`Connections` now are a layer above the underlying `Stack`, they now have a `Stream` map allowing the management of more than one stream per `Connection`. Right now it's not really used but will be in the future.
Having `Connection` as a unified layer above the stacks allows for managing events in a more structured way.

This involves no changes in the API - even accessing the underlying PeerConnection is done the same way.


erizo_controller/erizoClient/src/Socket.js
```diff
@@ -178,8 +178,8 @@ const Socket = (newIo) => {
       Logger.error('Trying to send a message over a disconnected Socket');
       return;
     }
-    socket.emit(type, options, sdp, (response, respCallback) => {
-      callback(response, respCallback);
+    socket.emit(type, options, sdp, (...args) => {
```
jcague: :+1:

erizo_controller/erizoClient/src/ConnectionManager.js
```diff
@@ -0,0 +1,161 @@
...
+    that.peerConnection = stack.peerConnection; // For backwards compatibility
+    stack.peerConnection.onaddstream = (stream) => {
+      if (that.onaddstream) {
+        Logger.error('OnAddStream from peerConnection - stream', stream);
+        that.onaddstream(stream);
+      }
+    };
+
+    stack.peerConnection.onremovestream = (stream) => {
+      if (that.onremovestream) {
```
jcague: where can that.onremovestream be rewritten now?

erizo_controller/erizoClient/src/ConnectionManager.js
```diff
@@ -0,0 +1,161 @@
...
+import FirefoxStack from './webrtc-stacks/FirefoxStack';
+import FcStack from './webrtc-stacks/FcStack';
+import Logger from './utils/Logger';
+import ErizoMap from './utils/ErizoMap';
+import ConnectionHelpers from './utils/ConnectionHelpers';
+
+
+let ErizoSessionId = 103;
+
+const ErizoConnection = (specInput, erizoId = undefined) => {
```
jcague: could we make ErizoConnection a ES6 object as we do with Client?

```js
class ErizoConnection {
...
}
```

erizo_controller/erizoClient/src/ConnectionManager.js
```diff
@@ -0,0 +1,161 @@
...
+    stack.peerConnection.oniceconnectionstatechange = (ev) => {
+      if (that.oniceconnectionstatechange) {
+        that.oniceconnectionstatechange(ev.target.iceConnectionState);
+      }
+    };
+  }
+  return that;
+};
+
+const ConnectionManager = () => {
```
jcague: same as above, could we create a `class ConnectionManager {...}`?

erizo_controller/erizoClient/src/Room.js
```diff
@@ -137,7 +146,7 @@ const Room = (altIo, altConnection, specInput) => {
         stream.pc.remove(peerSocket);
       }
     };
-    connection.addStream(stream.stream);
+    connection.addStream(stream);
```
jcague: :+1:

erizo_controller/erizoClient/src/ConnectionManager.js
jcague: also, should we name it `ErizoConnectionManager`?

erizo_controller/erizoClient/src/ConnectionManager.js
```diff
@@ -0,0 +1,161 @@
...
+  that.removeStream = (stream) => {
+    const streamId = stream.getID();
+    if (!streamsMap.has(streamId)) {
+      Logger.warning(`message: Cannot remove stream not in map, streamId: ${streamId}`);
+      return;
+    }
+    streamsMap.remove(streamId);
+  };
+
+  that.processSignalingMessage = (msg) => {
```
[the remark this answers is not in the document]
lodoyun: I find the long syntax more consistent (the number of parameters does not change the way to do it) readable (esp. when it has to be split because lines are too long). I may be missing something but I see no advantages for the short syntax in these cases.

erizo_controller/erizoClient/src/ConnectionManager.js
```diff
@@ -0,0 +1,161 @@
...
+    Logger.debug(`message: getOrBuildErizoConnection, erizoId: ${erizoId}`);
+    let connection = {};
+
+    if (erizoId === undefined) {
+      // we have no erizoJS id - p2p
+      return ErizoConnection(specInput);
+    }
+    connection = ErizoConnection(specInput, erizoId);
+    if (that.ErizoConnectionsMap.has(erizoId)) {
+      that.ErizoConnectionsMap.get(erizoId)[connection.sessionId] = connection;
```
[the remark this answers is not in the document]
lodoyun: sessionId is a globally unique identifier for the Connections that existed before this PR. I decided to use it to organize the Connections attached to a particular erizoJS.

erizo_controller/erizoClient/src/ConnectionManager.js
lodoyun: We're not doing it now, but decided to keep it there since the option existed in the original code.

Issue #1141, opened 2018-01-30
Sdp Issue on Firefox beta 59 - subscriber stopped working


kekkokk:
There's a problem with the new Firefox 59 beta.
Seems like the Sdp generation is a bit different and puts m=video before m=audio.
this is ok for the publisher but breaks the subscriber.

when processing the answer from licode this error is raised:

![screen shot 2018-01-30 at 10 53 02 am](https://user-images.githubusercontent.com/15072723/35559769-d1622160-05ab-11e8-9f72-b2129e8fe7f0.png)

what Licode does is returning an answer with audio before video and firefox seems doesn't like it.
the mappings are ok, only inverted.

Firefox offer: 
`
a=group:BUNDLE sdparta_0 sdparta_1

m=video 9 UDP/TLS/RTP/SAVPF 120 121 126 97
`

Licode answer:
`
a=group:BUNDLE sdparta_1 sdparta_0

m=audio 1 UDP/TLS/RTP/SAVPF 109
`

PR #1142, opened 2018-01-30
Fix Firefox 59

kekkokk:
Ok so,
With Firefox 59 (now in beta) there are lots of problems.

1) It generates offer with m line video before audio. this caused an "error" when generating the answer.
For more information see issue: https://github.com/lynckia/licode/issues/1141

2) Apparently when creating offer, even passing offertoreceiveaudio: false and offertoreceivevideo: false, it creates offer with 2 media in sendrecv. 
I don't know if it's a bug or not but you can follow the discussion from comment 3 [HERE (Bugzilla)](https://bugzilla.mozilla.org/show_bug.cgi?id=1433953)
you can try the example I posted on comment 3 with Firefox 59.
This caused problem with firefox simulcast publisher because the offer generated was audio in sendonly and video in sendrecv. we were generating an answer with both medias in recvonly (the first stream parsed) and when setting answer Firefox throwed an error.
Since licode can't handle multiple direction in one peerconnection my solution was to munge sdp object and set sendonly and recv only in the appropriate places.

Everything's seems to be working fine now, I haven't done any class test so maybe I need help with that.
Also obviously thanks @Trisfald  who helped me with c++ part

ps. there is also a new variable getSessionVersion that we use to increment at every ice restart. maybe you can keep it so when multiple sdp negotiation will be supported, you will increment that.

### DIFF
--- a/erizo/src/erizo/SdpInfo.h
+++ b/erizo/src/erizo/SdpInfo.h
@@ -258,11 +258,11 @@ class SdpInfo {
   */
   const std::vector<Rid>& rids() const { return rids_; }
 
-  int getSessionCounter() const;
+  int getSessionVersion() const;
 
-  void setSessionCounter(int counter);
+  void setSessionVersion(int counter);
 
-  int session_counter;
+  int session_version;
   /**
    * The audio and video SSRCs for this particular SDP.
    */
--- a/erizo/src/erizo/SdpInfo.h
+++ b/erizo/src/erizo/SdpInfo.h
@@ -238,6 +238,10 @@ class SdpInfo {
   bool supportPayloadType(const unsigned int payloadType);
 
   void createOfferSdp(bool videoEnabled, bool audioEnabled, bool bundle);
+
+  void setFirstMediaReceived(MediaType media);
+
+  MediaType getFirstMediaReceived() const;
   /**
    * @brief copies relevant information from the offer sdp for which this will be an answer sdp
    * @param offerSdp The offer SDP as received via signaling and parsed
@@ -322,6 +326,7 @@ class SdpInfo {
   int audioSdpMLine;
   int videoCodecs, audioCodecs;
   unsigned int videoBandwidth;
+  MediaType first_media_received_;
   std::vector<CandidateInfo> candidateVector_;
   std::vector<CryptoInfo> cryptoVector_;
   std::vector<RtpMap> internalPayloadVector_;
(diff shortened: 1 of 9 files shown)


erizo/src/erizo/SdpInfo.h
```diff
@@ -254,6 +258,11 @@ class SdpInfo {
   */
   const std::vector<Rid>& rids() const { return rids_; }
 
+  int getSessionVersion() const;
```
jcague: I think it's always good to add support to SDP fields, but I'd do that in separate PRs to reduce the scope of fixes.

erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
```diff
@@ -92,6 +92,7 @@ const BaseStack = (specInput) => {
       localDesc.sdp = that.enableSimulcast(localDesc.sdp);
     }
     localSdp = SemanticSdp.SDPInfo.processString(localDesc.sdp);
+    SdpHelpers.forceDirection(localSdp, isSubscribe);
```
jcague: I think this code should only be added to the Firefox stack.

erizo_controller/erizoJS/models/SessionDescription.js
```diff
@@ -205,23 +205,34 @@ class SessionDescription {
...
       ipVer: 4,
       address: '127.0.0.1' });
     sdp.name = 'LicodeMCU';
 
     sdp.msidSemantic = { semantic: 'WMS', token: generateRandom(10) };
 
-    if (info.hasAudio()) {
-      const media = getMediaInfoFromDescription(info, sdp, 'audio');
-      sdp.addMedia(media);
+    if (info.getFirstMediaReceived() === 'audio') {
```
jcague: I see the problem, but I prefer a less hacky solution... would it be possible to add medias in the same order they appear in the bundle group field?

kekkokk: Yeah seems a little hacky way also to me but unfortunately the datas are fetched from SdpInfo as separate field (audio and video).
Until there's a refactor in Sdpinfo.cpp that support an ordered array of media (that I tought will be done in the future with the singlepeerconnection feature) I can't think other ways to pass the order than this hacky variable.

erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
kekkokk: Ok, I don't think could harm to force here too but if you prefer I will move to Firefox Stack

erizo/src/erizo/SdpInfo.h
kekkokk: ok I'll move out

erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
kekkokk: Also they replied to me that Firefox now support transceivers, so we have to set direction in videoTransceiver and audioTransceiver. I think it's more an adapter job since they already do this in safari. So this will be moved in firefox stack as a temporary fix until adapter came out with their own solution

kekkokk:
Also read this:
https://github.com/webrtc/adapter/issues/760#issuecomment-361874732

Maybe they'll fix in 59 before the realease.

Still the media order is a todo fix for us

kekkokk:
I made two separates PR. close this

PR #1143, opened 2018-01-31
fix media order

kekkokk:
Since firefox 59 the media order in the sdps has changed. this hacky fix (until a complete refactor of SdpInfo is done) will resolve the issue.

[x] It maybe needs and it does NOT includes Unit Tests


kekkokk:
No problem, I would also prefer to have a medialist ordered. But required a bit of a refactor and I'm not too practical in c++ as you know, so if you would manage this on your own way it's better for everyone!!  this should be done before Firefox 59 anyway. Keep this in mind

PR #1144, opened 2018-01-31
fix firefox sendrecv

kekkokk:
There's an open bug in firefox 59 that had to be landed in 59 but apparently it's still there, that causes the browser to ingore the offertoreceivevideo: false and offertoreceiveaudio: false, resulting in an sdp in sendrecv direction. (this because they now implemented transceivers)
also in simulcast, firefox generates audio in sendonly direction and video in sendrecv. this breaks the logic since licode can handle only one direction for every peerconnection.

this is a temporary fix, since Firefox or adapter have to handle this misbehavior.

you can follow these topics if interested in:
https://github.com/webrtc/adapter/issues/760
https://bugzilla.mozilla.org/show_bug.cgi?id=1433953 (from comment 3)

### DIFF
--- a/erizo_controller/erizoClient/src/utils/SdpHelpers.js
+++ b/erizo_controller/erizoClient/src/utils/SdpHelpers.js
@@ -1,3 +1,5 @@
+import Direction from '../../../common/semanticSdp/Direction';
+
 const SdpHelpers = {};
 
 SdpHelpers.addSim = (spatialLayers) => {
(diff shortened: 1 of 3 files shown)


jcague:
@kekkokk seems like the bug is already fixed in Firefox 59, has it landed in the binaries yet?

kekkokk:
I'll test in the afternoon then I'll close pr if everything's fixed.

mccob:
any news about this PR ? 

seems still not working with FF 59 with this files (run install erizoControler & install basic example) but I haven't merged with commits since this PR. Only tested files here. May merging would help (but there is conflicts).

mccob:
tried to merge this PR with last release of licode

get javascript error : 

ERROR:  message: TypeError: "sdp is undefined"  in baseStack at Create Offer
TypeError
​
columnNumber: 3
fileName: "/erizo.js"
lineNumber: 5275

tried to understand why

kekkokk:
this is an old pr so likely the code diverged and something screwd up. 
you also need to run ./installErizo.sh -c to compile the client.

mccob:
I run  ./installErizo.sh but may be more fast to use -c :) thank you

Will try to find there is an error. If you any idea, don't hesitate ;)

PR #1151, opened 2018-02-09
Add/edge support

kekkokk:
In march will be released this fix:
[Edge Support](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12919339/)

This allows Licode to support it.
Note: the Edge publisher side in VP8 still sucks, even in p2p.
So for an "usable" environment we have to wait that Edge fixes its vp8 encoder


erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
```diff
@@ -17,6 +17,7 @@ const BaseStack = (specInput) => {
 
   that.pcConfig = {
     iceServers: [],
+    bundlePolicy: 'max-bundle',
```
jcague: have you tested this property with Chrome/Firefox? My only concern is not to break things that are working with the browsers that we currently support.

erizo_controller/erizoJS/models/SessionDescription.js
```diff
@@ -68,7 +68,9 @@ function getMediaInfoFromDescription(info, sdp, mediaType) {
 
   let ice = info.getICECredentials(mediaType);
   if (ice) {
-    media.setICE(new ICEInfo(ice[0], ice[1]));
+    const thisIceInfo = new ICEInfo(ice[0], ice[1]);
+    thisIceInfo.setEndOfCandidates('end-of-candidates');
```
jcague: does this break something in Chrome/Firefox?

erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
kekkokk: this breaks p2p firefox, sorry I forgot to update the pr. this must be moved only in EdgeStack, maybe in a separate pr in the future.

erizo_controller/erizoJS/models/SessionDescription.js
kekkokk: this is bulletproof

arsenicraghav: This is breaking things in chrome, local and remote stream turns black after merging this change

PR #1152, opened 2018-02-09
Fix Rtp padding generator for older firefox build.

kekkokk:
Also set marker to false per RFC

Idk if it needs some unit tests


jcague:
does these changes still works for Chrome? I think to remember we had some issues in the past with the marker set to true (but I might be wrong).

erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp
```diff
@@ -100,6 +100,7 @@ void RtpPaddingGeneratorHandler::sendPaddingPacket(std::shared_ptr<DataPacket> p
   RtpHeader *rtp_header = reinterpret_cast<RtpHeader*>(padding_packet->data);
 
   rtp_header->setSeqNumber(sequence_number.output);
+  rtp_header->setMarker(false);
```
jcague: could we set the marker bit inside RtpUtils::makePaddingPacket please?

erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp
```diff
@@ -110,7 +111,9 @@ void RtpPaddingGeneratorHandler::onPacketWithMarkerSet(std::shared_ptr<DataPacke
   for (uint i = 0; i < number_of_full_padding_packets_; i++) {
     sendPaddingPacket(packet, kMaxPaddingSize);
   }
-  sendPaddingPacket(packet, last_padding_packet_size_);
+  // Temporary fix since https://bugzilla.mozilla.org/show_bug.cgi?id=1435025 is fixed
+  // Only send full rtp padding packet as suggested also in webrtc code.
+  // sendPaddingPacket(packet, last_padding_packet_size_);
```
jcague: this could affect the way the congestion control works, have you tested it with Chrome too? did you notice any significant change in the amount of time it takes for simulcast to increase the quality layers?

kekkokk:
I tested with chrome and seems ok! also because chrome itself send the packet with the marker false.
I'll try some more in the afternoon since the tests I made were done without listening to rembs (we found that they're too conservatives and tend to lower a lot in short period of time. but instead, making the senderBWestimator works only with RRs it's way more stable.
I did a pr time ago that was closed in which I used this approach.

If you can also test this will be great, but I don't think this will cause any issue.

erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp
kekkokk: I'm sorry idk why i messed up here

PR #1160, opened 2018-03-01
Fix some crashes when Erizo is closing MediaStreams and WebRtcConnections

jcague:
**Description**

Fix some crashes when Erizo is closing MediaStreams and WebRtcConnections. It doesn't fix all the cases since I still see some cores from time to time, but it should help reducing the issues.


erizoAPI/MediaStream.cc
```diff
@@ -18,6 +18,8 @@ using v8::Exception;
 using v8::Value;
 using json = nlohmann::json;
 
+DEFINE_LOGGER(MediaStream, "ErizoAPI.MediaStream");
```
lodoyun: 👍 to using the logger here

PR #1162, opened 2018-03-02
Initial support for multiple MediaStreams in a single WebRtcConnection

jcague:
**Description**

This PR adds initial support for handling multiple streams in a single WebRtcConnection inside Erizo.

Not needed.

### DIFF
--- a/erizo/src/erizo/WebRtcConnection.cpp
+++ b/erizo/src/erizo/WebRtcConnection.cpp
@@ -402,7 +402,7 @@ void WebRtcConnection::onTransportData(std::shared_ptr<DataPacket> packet, Trans
   RtpHeader *head = reinterpret_cast<RtpHeader*> (buf);
   RtcpHeader *chead = reinterpret_cast<RtcpHeader*> (buf);
   uint32_t ssrc = head->getSSRC();
-  if (chead->isRtcp()) {
+  if (chead->isRtcp() && chead->packettype != RTCP_Sender_PT) {  // Sender Report
     ssrc = chead->getSourceSSRC();
   }
   forEachMediaStream([packet, transport, ssrc] (const std::shared_ptr<MediaStream> &media_stream) {
--- a/erizo/src/erizo/WebRtcConnection.cpp
+++ b/erizo/src/erizo/WebRtcConnection.cpp
@@ -100,14 +100,6 @@ bool WebRtcConnection::init() {
   return true;
 }
 
-bool WebRtcConnection::isSourceSSRC(uint32_t ssrc) {
-  return media_stream_->isVideoSourceSSRC(ssrc) || media_stream_->isAudioSourceSSRC(ssrc);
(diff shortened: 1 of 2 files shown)


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -385,7 +398,18 @@ void WebRtcConnection::onTransportData(std::shared_ptr<DataPacket> packet, Trans
   if (getCurrentState() != CONN_READY) {
     return;
   }
-  media_stream_->onTransportData(packet, transport);
+  char* buf = packet->data;
+  RtpHeader *head = reinterpret_cast<RtpHeader*> (buf);
+  RtcpHeader *chead = reinterpret_cast<RtcpHeader*> (buf);
+  uint32_t ssrc = head->getSSRC();
+  if (chead->isRtcp()) {
+    ssrc = chead->getSourceSSRC();
```
lodoyun: I think this won't work for sender reports. In that case, we need `chead->getSSRC()`

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -385,7 +398,18 @@ void WebRtcConnection::onTransportData(std::shared_ptr<DataPacket> packet, Trans
...
+  }
+  forEachMediaStream([packet, transport, ssrc] (const std::shared_ptr<MediaStream> &media_stream) {
+    if (media_stream->isSourceSSRC(ssrc) || media_stream->isSinkSSRC(ssrc)) {
```
lodoyun: This is ok. `isSourceSSRC` would make sure packets from a publisher are properly routed while `isSink` would work for feedback packets from a subscriber.

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -201,8 +211,10 @@ bool WebRtcConnection::processRemoteSdp() {
   extension_processor_.setSdpInfo(local_sdp_);
   local_sdp_->updateSupportedExtensionMap(extension_processor_.getSupportedExtensionMap());
 
-  local_sdp_->video_ssrc_list.push_back(media_stream_->getVideoSinkSSRC());
-  local_sdp_->audio_ssrc = media_stream_->getAudioSinkSSRC();
+  forEachMediaStream([this] (const std::shared_ptr<MediaStream> &media_stream) {
+    local_sdp_->video_ssrc_list.push_back(media_stream->getVideoSinkSSRC());
+    local_sdp_->audio_ssrc = media_stream->getAudioSinkSSRC();
```
lodoyun: This is not part of this PR but, do you have a plan here? The obvious solution is to have another list for audio ssrcs.

jcague: I'd rather move the ssrc tracking to a new class called stream_sdp or similar.

PR #1164, opened 2018-03-06
 Decouple Connection logic in ErizoJS

jcague:
**Description**

This PR refactors the way we handle messages in ErizoJSController. The main idea is to decouple things between ErizoJSController, Connection, Publisher and Subscriber, while keeping the focus on making things easier for Single Peer Connection.

Not needed.

### DIFF
--- a/erizo_controller/erizoJS/erizoJSController.js
+++ b/erizo_controller/erizoJS/erizoJSController.js
@@ -215,7 +215,7 @@ exports.ErizoJSController = function (threadPool, ioThreadPool) {
      * Removes a publisher from the room. This also deletes the associated OneToManyProcessor.
      */
     that.removePublisher = function (clientId, streamId) {
-      return new Promise(function(resolve, reject) {
+      return new Promise(function(resolve) {
         var publisher = publishers[streamId];
           if (publisher !== undefined) {
               log.info(`message: Removing publisher, id: ${clientId}, streamId: ${streamId}`);
--- a/erizo_controller/erizoJS/erizoJSController.js
+++ b/erizo_controller/erizoJS/erizoJSController.js
@@ -5,14 +5,12 @@ var amqper = require('./../common/amqper');
 var Client = require('./models/Client').Client;
 var Publisher = require('./models/Publisher').Publisher;
 var ExternalInput = require('./models/Publisher').ExternalInput;
-var SessionDescription = require('./models/SessionDescription');
-var SemanticSdp = require('./../common/semanticSdp/SemanticSdp');
 
 // Logger
 var log = logger.getLogger('ErizoJSController');
 
 exports.ErizoJSController = function (threadPool, ioThreadPool) {
-    var that = {},
+    let that = {},
         // {streamId1: Publisher, streamId2: Publisher}
         publishers = {},
         // {clientId: Client}
@@ -20,31 +18,18 @@ exports.ErizoJSController = function (threadPool, ioThreadPool) {
         io = ioThreadPool,
         // {streamId: {timeout: timeout, interval: interval}}
         statsSubscriptions = {},
-        MIN_SLIDESHOW_PERIOD = 2000,
-        MAX_SLIDESHOW_PERIOD = 10000,
-        PLIS_TO_RECOVER = 3,
-        initWebRtcConnection,
-        closeWebRtcConnection,
+        onAdaptSchemeNotify,
+        onPeriodicStats,
+        addListenerToConnection,
+        onConnectionStatusEvent,
+        closeNode,
         getOrCreateClient,
-        getSdp,
-        getRoap,
-
-        CONN_INITIAL        = 101,
-        // CONN_STARTED        = 102,
-        CONN_GATHERED       = 103,
-        CONN_READY          = 104,
-        CONN_FINISHED       = 105,
-        CONN_CANDIDATE      = 201,
-        CONN_SDP            = 202,
-        CONN_FAILED         = 500,
         WARN_NOT_FOUND      = 404,
-        WARN_CONFLICT       = 409,
-        WARN_PRECOND_FAILED = 412,
-        WARN_BAD_CONNECTION = 502;
+        WARN_CONFLICT       = 409;
 
     that.publishers = publishers;
     that.ioThreadPool = io;
-    
+
     getOrCreateClient = (clientId) => {
       log.debug(`getOrCreateClient with id ${clientId}`);
       let client = clients.get(clientId);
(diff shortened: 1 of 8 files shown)


erizo_controller/erizoJS/models/Connection.js
```diff
@@ -52,27 +67,91 @@ class Connection {
...
+        case CONN_INITIAL:
+          this.emit('status_event', {type: 'started'}, newStatus);
+          break;
+
+        case CONN_SDP:
+        case CONN_GATHERED:
+          this.wrtc.localDescription = new SessionDescription(this.wrtc.getLocalDescription());
+          const sdp = this.wrtc.localDescription.getSdp();
+          mess = sdp.toString();
+          mess = mess.replace(this.privateRegexp, this.publicIP);
```
lodoyun: should probably be `this.options.privateRegexp` and `this.options.publicIP`

erizo_controller/erizoJS/models/Connection.js
```diff
@@ -52,27 +67,91 @@ class Connection {
...
+
+          const info = {type: this.options.createOffer ? 'offer' : 'answer', sdp: mess};
+          this.emit('status_event', info, newStatus);
+          break;
+
+        case CONN_CANDIDATE:
+          mess = mess.replace(this.privateRegexp, this.publicIP);
```
lodoyun: same as above

erizo_controller/erizoJS/models/Connection.js
```diff
@@ -52,27 +67,91 @@ class Connection {
...
   }
 
-  _createMediaStream (id) {
+  _createMediaStream (id, mediaStreamOptions) {
     log.debug(`message: _createMediaStream, connectionId: ${this.id}, mediaStreamId: ${id}`);
     const mediaStream = new addon.MediaStream(this.threadPool, this.wrtc, id,
       this._getMediaConfiguration(this.mediaConfiguration));
     mediaStream.id = id;
+    if (mediaStreamOptions) {
+      mediaStream.metadata = mediaStreamOptions;
```
lodoyun: 👍 However, now that we're adding metadata here we have to reconsider what we're doing with metadata at the WebRTCConnection level since we will add there the metadata from the first stream. Does it still make sense to have different metadata at each level?

erizo_controller/erizoJS/models/Connection.js
```diff
@@ -52,27 +67,91 @@ class Connection {
...
       this.mediaStreams.set(id, mediaStream);
     }
   }
+
   removeMediaStream(id) {
-    if(this.mediaStreams.get(id) !== undefined) {
+    if (this.mediaStreams.get(id) !== undefined) {
       this.wrtc.removeMediaStream(id);
       this.mediaStreams.get(id).close();
       this.mediaStreams.delete(id);
```
lodoyun: Maybe use `getNumMediaStreams` instead? Otherwise we don't really need it at the moment

erizo_controller/erizoJS/models/Publisher.js
```diff
@@ -129,24 +125,169 @@ class Source extends Node {
...
+    for (const index in disabledHandlers) {
+      this.mediaStream.disableHandler(disabledHandlers[index]);
+    }
+  }
+
+  _onSchemeSlideShowModeChange(message, clientId) {
+    this.setSlideShow(message.enabled, clientId);
+  }
+
+  onSignalingMessage(msg) {
```
lodoyun: 👍 to having the publisher and the subscriber negotiation separated. We can probably have the common parts in `NodeClass` at some point but just not having them inside an `if` makes it a lot clearer

erizo_controller/erizoJS/models/Publisher.js
```diff
@@ -129,24 +125,169 @@ class Source extends Node {
...
+        this.mediaStream.generatePLIPacket();
+      }, period);
+    } else {
+      for (let pliIndex = 0; pliIndex < PLIS_TO_RECOVER; pliIndex++) {
+        this.mediaStream.generatePLIPacket();
+      }
+
+      subscriber.mediaStream.setSlideShowMode(false);
+      subscriber.mediaStream.slideShowMode = false;
+      if (this.mediaStream.periodicPlis !== undefined) {
```
lodoyun: Can we use `maybeStopSlideshow()` here?

erizo_controller/erizoJS/models/Publisher.js
```diff
@@ -129,24 +125,169 @@ class Source extends Node {
...
+        if (action.enable) {
+          this.enableHandlers(publisherSide ? undefined : clientId, action.handlers);
+        } else {
+          this.disableHandlers(publisherSide ? undefined : clientId, action.handlers);
+        }
+        break;
+    }
+  }
+
+  generateVideoKeyFrame() {
```
lodoyun: Another nitpick. Since we're not actually generating a keyframe but requesting one via PLI. Can we name this `requestVideoKeyFrame`?

erizo_controller/erizoJS/models/Connection.js
lodoyun: I mean in the log message [below](https://github.com/lynckia/licode/pull/1164/files/b2b244865ad336fbcefe14ec01e3acd4eb57e05f#diff-a257b3a662e2e651ddb5db4d1565d303L79)

erizo_controller/erizoJS/models/Connection.js
jcague: only for logging purposes, but you're right that we'll update it in the future, so we need to follow a different approach. Anyway, I'd do it in another PR.

erizo_controller/erizoJS/models/Connection.js
jcague: good catch!

erizo_controller/erizoJS/models/Connection.js
jcague: you got it!

erizo_controller/erizoJS/models/Connection.js
jcague: good catch!

erizo_controller/erizoJS/models/Publisher.js
jcague: you got it

erizo_controller/erizoJS/models/Publisher.js
jcague: great! good catch!

erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -495,27 +259,13 @@ exports.ErizoJSController = function (threadPool, ioThreadPool) {
     that.removeSubscriber = function (clientId, streamId) {
         const publisher = publishers[streamId];
         if (publisher && publisher.hasSubscriber(clientId)) {
-            
             let subscriber = publisher.getSubscriber(clientId);
+            subscriber.close();
```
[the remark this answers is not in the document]
jcague: good catch!

erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -262,142 +123,46 @@ exports.ErizoJSController = function (threadPool, ioThreadPool) {
...
+                   'clientId: ' + clientId + ', ' +
+                   'streamId: ' + streamId + ', ' +
+                   logger.objectToLog(options) + ', ' +
+                   logger.objectToLog(options.metadata));
+          publisher = new Publisher(clientId, streamId, connection, options);
+          publishers[streamId] = publisher;
+          publisher.initMediaStream();
+          publisher.on('callback', onAdaptSchemeNotify.bind(this, callbackRpc));
+          publisher.on('periodic_stats', onPeriodicStats.bind(this, streamId, undefined));
+          addListenerToConnection(connection, callbackRpc, streamId, clientId);
```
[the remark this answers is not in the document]
jcague: sure! you got it

PR #1167, opened 2018-03-15
 Initial support to negotiate SDPs in Single Peer Connection

jcague:
**Description**

More work towards having Single Peer Connection in Lynckia. With this PR we start to have negotiation working in some cases with Chrome. I had to touch multiple files and how the negotiation is performed, so I don't want to add and fix more things to Single Peer Connection in this PR and test that it doesn't break anything.

Nothing changes in the public APIs. 
I added a thing to be considered by Native Clients: a new attribute called `label` that is needed to identify the stream in the web clients.


erizo/src/erizo/MediaStream.cpp
```diff
@@ -127,8 +129,8 @@ bool MediaStream::setRemoteSdp(std::shared_ptr<SdpInfo> sdp) {
 
   bundle_ = remote_sdp_->isBundle;
 
-  setVideoSourceSSRCList(remote_sdp_->video_ssrc_list);
-  setAudioSourceSSRC(remote_sdp_->audio_ssrc);
+  setVideoSourceSSRCList(remote_sdp_->video_ssrc_map[getLabel()]);
```
lodoyun: 👍 I like the ssrc[label] map.

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -183,6 +190,18 @@ bool WebRtcConnection::setRemoteSdpInfo(std::shared_ptr<SdpInfo> sdp) {
 
 std::shared_ptr<SdpInfo> WebRtcConnection::getLocalSdpInfo() {
   ELOG_DEBUG("%s message: getting local SDPInfo", toLog());
+  forEachMediaStream([this] (const std::shared_ptr<MediaStream> &media_stream) {
```
lodoyun: I don't get why this is added here... Aren't we adding the SinkSSRCs to the local_sdp in both `createOffer` and `processRemoteSdp`. So in both cases of the negotiation we should have the local_sdp correctly filled in.

jcague: agree, I'll make sure we don't need this by running some test

erizo_controller/erizoClient/src/ErizoConnectionManager.js
```diff
@@ -160,9 +156,11 @@ class ErizoConnectionManager {
   closeConnection(connection) {
     Logger.debug(`Removing connection ${connection.sessionId}
        with erizoId ${connection.erizoId}`);
-    connection.close();
-    if (this.ErizoConnectionsMap.get(connection.erizoId) !== undefined) {
-      delete this.ErizoConnectionsMap.get(connection.erizoId)[connection.sessionId];
+    if (connection.streamsMap.size() === 0) {
```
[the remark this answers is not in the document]
jcague: 👍

PR #1169, opened 2018-03-19
Recordings supported video-orientation.

yannistseng:
**Description**
- Parse the RTP header extension to get video-orientation tags and add it to recordings at the container level.


kekkokk:
nice

yannistseng:
Hi Jacque,

Thank you for your reply!

I plan to update `RtpExtensionProcessor` to add support to video rotation extension, but I have the following questions.

I noticed that we called the method `processRtpExtensions` in `WebRtcConnection::write()`;
https://github.com/lynckia/licode/blob/a7cc8c402bf65a2f23d020b3b637234abdcd175c/erizo/src/erizo/WebRtcConnection.cpp#L626
I think there is no possible to enter the if statement, because the method `RtpHeader::setExtension()` never gets called, right?
https://github.com/lynckia/licode/blob/ac9b138fbf80a87444365a8f86f6dec0628d28a9/erizo/src/erizo/rtp/RtpExtensionProcessor.cpp#L70


Any comments will be greatly appreciated, thank you!

yannistseng:
Ok, I will fix it!

erizo/src/erizo/media/ExternalOutput.cpp
```diff
@@ -376,8 +386,9 @@ bool ExternalOutput::initContext() {
     video_stream_->codec->width = 640;
     video_stream_->codec->height = 480;
     video_stream_->time_base = (AVRational) { 1, 30 };
+    video_stream_->metadata = genVideoMetadata();
```
jcague: hmmm... this solution will only work if the video is always in the same orientation, am I right?

PR #1171, opened 2018-03-20
Fix/erizo client build errors

lodoyun:
**Description**

This fixes ErizoClient builds by specifying version numbers for all components involved and working around a couple of errors.
It also makes sure we notice when builds fail in CI by capturing gulp errors and terminating the process with an error code. That error code is also now properly propagated to the parent scripts if any (for instance `installErizo.sh`.

### DIFF
--- a/erizo_controller/erizoClient/gulp/erizoFcTasks.js
+++ b/erizo_controller/erizoClient/gulp/erizoFcTasks.js
@@ -14,9 +14,9 @@ const erizoFcTasks = (gulp, plugins, config) => {
   that.bundle = () =>
     gulp.src(erizoFcConfig.entry)
     .pipe(plugins.webpackGulp(erizoFcConfig.webpackConfig, plugins.webpack))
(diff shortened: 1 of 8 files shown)


kekkokk:
how do we invoke gulp watch now that gulp-cli is not global?

lodoyun:
@kekkokk It's in `licode/node_modules/.bin/gulp`. I changed this in order to avoid confusions, we were installing gulp both in the global scope and in the scope of the package. I do prefer to install it in the package to avoid problems if your development machine, like mine, has more than one nvm.sh environment. I have one for licode and another one that is global.

kekkokk:
just a note: I'd prefer to use $LICODE_ROOT/node_modules/.bin/gulp erizo instead of ../../node_modules/.bin/gulp erizo

traw:
I think this need to be merged in master,  since `./scripts/installErizo.sh` couldn't find `gulp` command.

PR #1180, opened 2018-04-03
Fix/lipsync

kekkokk:
Seems that the previous algo when switching quality layers relied on the last timestamp sent +1.
but in certain situations the previous packet sent maybe was sent few seconds before, causing a desync of N seconds. this maybe can happen when simulcasting and the publisher suddently drop his connection to few kbits. chrome doesn't publish the upper spatial layer anymore and only the base quality layer is published.
The quality manager takes few seconds to realize this and this cause a desync in the output.


kekkokk:
I think I'm using c++11 and this syntax is not supported. need help to make it compatible if you can.

kekkokk:
added few improvements. 
Let me know what you think

PR #1193, opened 2018-04-11
Add option to set limit of Erizos per Room

jcague:
**Description**

Add an option in the config file to set a limit of Erizos per room. 

Not needed.

### DIFF
--- a/erizo_controller/erizoController/roomController.js
+++ b/erizo_controller/erizoController/roomController.js
@@ -16,7 +16,8 @@ exports.RoomController = function (spec) {
         //     {publishers: [ids], kaCount: count, agentId: agentId, internalId: internalId}}
         erizos = {},
 
-        maxErizos = spec.maxConnections || global.config.erizoController.maxErizosInRoom,
+        maxErizosUsedByRoom = spec.maxErizosUsedByRoom || 
+                                global.config.erizoController.maxErizosUsedByRoom,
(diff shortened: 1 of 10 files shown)


scripts/licode_default.js
```diff
@@ -82,6 +82,7 @@ config.erizoController.listen_port = 8080; //default value: 8080
 
 config.erizoController.exitOnNuveCheckFail = false;  // default value: false
 config.erizoController.allowSinglePC = false;  // default value: false
+config.erizoController.maxErizosInRoom = 0;  // default value: 0
```
lodoyun: Consider `maxErizosUsedByRoom`

erizo_controller/erizoController/roomController.js
```diff
@@ -11,9 +11,14 @@ exports.RoomController = function (spec) {
         subscribers = {},
         // {id: erizoJS_id}
         publishers = {},
-        // {erizoJS_id: {publishers: [ids], kaCount: count}}
+
+        // {erizoJS_id:
+        //     {publishers: [ids], kaCount: count, agentId: agentId, internalId: internalId}}
         erizos = {},
 
+        maxErizos = spec.maxConnections || global.config.erizoController.maxErizosInRoom,
```
lodoyun: Why use `maxConnections`? It's probably clearer if we use a name closer to the one in the config

Issue #1197, opened 2018-04-16
ErizoList.js cannot get list of Erizos (


khoachung89:
I'm installing Licode on my Ubuntu machine at local following the guide on: [http://lynckia.com/licode/install.html](http://lynckia.com/licode/install.html)
I also install and run the basic example.
As the requirement, the whole system need to be served over ssl, i have installed nginx and configure it as a proxy and redirect all request to basic example (on port 3000). My nginx configuration as below:

`upstream licode {
    server 127.0.0.1:3001;
}

server {
	listen 443;
	server_name 192.168.126.133;
	ssl on;
	ssl_certificate /home/chung/ssl/selfsigned.crt;
	ssl_certificate_key /home/chung/ssl/selfsigned.key;
	ssl_session_timeout 5m;
    	ssl_protocols  SSLv2 SSLv3 TLSv1;
    	ssl_ciphers  HIGH:!aNULL:!MD5;
    	ssl_prefer_server_ciphers   on;
        #ssl_dhparam /etc/ssl/certs/dhparam.pem;
        ssl_stapling on;
        ssl_stapling_verify on;
        add_header Strict-Transport-Security max-age=15768000;

	location ~ ^/(.+\.(?:js|css|png|jp?g|gif|swf|ico|pdf|pdfx|mov|fla|zip|rar))$ {
        	alias /home/chung/licode/extras/basic_example/public;
        	try_files /$1 =404;

        	access_log off;
        	expires max;
	        add_header Pragma public;
        	add_header Cache-Control "public, must-revalidate, proxy-revalidate";
    	}

	location ~ ^/(.*)$ {
        	access_log off;
        	proxy_pass http://licode;
        	proxy_set_header X-Real-IP $remote_addr;
	        proxy_set_header Host $host;
        	proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

	        # WebSocket support (nginx 1.4)
        	proxy_http_version 1.1;
	        proxy_set_header Upgrade $http_upgrade;
	        proxy_set_header Connection "upgrade";

        	# Path rewriting
	        rewrite /licode$ / break;
        	rewrite /licode/(.*) /$1 break;
	        proxy_redirect off;
    	}
}`

Now, i can go to my browser and start testing.
I can see that i have joined the room already (as the test shown in the attached image)
![licodebug](https://user-images.githubusercontent.com/4082657/38791605-f22999fc-4172-11e8-93f8-c02cf7019fff.PNG)
[...]

Issue #1189, opened 2018-04-09
Licode not working with recent Firefox


mccob:
**Description**

Tested Licode with recent Firefox.
Already seen that there is others issues related to recent Firefox new webrtc stack.

**Steps to reproduce the issue:**
1.  install licode, use basic example 
2.  go to connection_test.html with recent Firefox (stable / beta / nightly, same issue)
3. tried with user agent change to use Chrome licode stack, same issue

**Describe the results you received:**

When reloading Chrome test connection on the other computer to etablish a video stream between Firefox PC & Chrome PC, there is this debug or error message (no issue between Chrome and Chrome) : 

**Basic example test connection with Firefox stack :** 

```
- Mic and Cam OK
- Connected to the room OK
```

no more steps

Console message firefox side : 

```
DEBUG:  Event: onRemoveStream erizo.js:27:39
WARNING:  message: Cannot remove stream not in map, streamId: 623165180164625900 erizo.js:27:39
DEBUG:  Trying to remove connection 107
       with erizoId  erizo.js:27:39
DEBUG:  Closing ErizoConnection erizo.js:27:39
DEBUG:  Event: ice-state-change erizo.js:27:39
DEBUG:  Event: icestatechanged erizo.js:27:39
DEBUG:  Event: stream-removed erizo.js:27:39
DEBUG:  Event: onAddStream erizo.js:27:39
DEBUG:  Event: stream-added erizo.js:27:39
INFO:  Subscribing to: 925010444114849800 erizo.js:27:39
INFO:  Subscriber added erizo.js:27:39
DEBUG:  message: getOrBuildErizoConnection, erizoId: erizo.js:27:39
DEBUG:  Building a new Connection erizo.js:27:39
DEBUG:  Firefox Stack erizo.js:27:39
INFO:  Starting Firefox stack erizo.js:27:39
INFO:  Starting Base stack 
Object { callback: callback()
, nop2p: true, audio: true, video: true, maxAudioBW: undefined, maxVideoBW: 100, limitMaxAudioBW: undefined, limitMaxVideoBW: 300, label: "", iceServers: (1) […], … }
erizo.js:27:39
DEBUG:  Creating offer 
Object { offerToReceiveVideo: true, offerToReceiveAudio: true }
 erizo.js:27:39
RTCIceServer.url is deprecated! Use urls instead. erizo.js:42
onaddstream is deprecated! Use peerConnection.ontrack instead. erizo.js:135
[...]

PR #1200, opened 2018-04-17
Fix issues that happened during first trials of Single Peer Connection

jcague:
**Description**

Bugs fixed in this PR:
- Firefox not accepting different orders of video/audio medias in the SDP.
- Wrong SDP directions when using Erizo.
- Simulcast with Single PC did not work sometimes.

Not needed


jcague:
this PR should fix #1197  and #1189

arsenicraghav:
the single peer connection appears to be working however we are not able to get the remote stream, a black screen shows up when running basic example.

PR #1212, opened 2018-05-08
Use a variable instead of event object to concatenate candidate attr.

fantasy4z:
By WebRTC spec, RTCIceCandidate.candidate is a readonly attribute. It makes
more sense to store, check, and concatenate candidate attribute line in another
variable, instead of the event object.

Spec: https://www.w3.org/TR/webrtc/#dom-rtcicecandidate

macOS Safari implements the event behavior by the spec. Include this patch to make Safari work correctly.

**Description**

### DIFF
--- a/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
+++ b/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
@@ -72,15 +72,14 @@ const BaseStack = (specInput) => {
         candidate: 'end',
       };
     } else {
-      if (!candidate.candidate.match(/a=/)) {
-        candidate.candidate = `a=${candidate.candidate}`;
-      }
-
       candidateObject = {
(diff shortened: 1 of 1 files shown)


erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
```diff
@@ -72,15 +72,14 @@ const BaseStack = (specInput) => {
...
-      if (!candidate.candidate.match(/a=/)) {
-        candidate.candidate = `a=${candidate.candidate}`;
-      }
-
       candidateObject = {
         sdpMLineIndex: candidate.sdpMLineIndex,
         sdpMid: candidate.sdpMid,
         candidate: candidate.candidate,
       };
+      if (!candidateObject.candidate.match(/a=/)) {
```
jcague: good catch!

PR #1216, opened 2018-05-11
Support legacy websocket polling

kekkokk:
We experienced some problems with some network configuration since the websockets were dropped by the firewall, but they can connect to our frontend (which is not forced to websockets).

So the idea here, is no permit a first connection via polling, and then if the upgrade to websocket succeded, then we can enable the "reconnect" feature.

Please give it a try since it's copied from my work repo, but should work

### DIFF
--- a/erizo_controller/erizoClient/src/Socket.js
+++ b/erizo_controller/erizoClient/src/Socket.js
@@ -19,6 +19,7 @@ const Socket = (newIo) => {
   const that = EventDispatcher();
   const defaultCallback = () => {};
   const messageBuffer = [];
+  that.reconnectSupported = false;
 
   that.CONNECTED = Symbol('connected');
   that.RECONNECTING = Symbol('reconnecting');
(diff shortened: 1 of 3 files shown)


kekkokk:
take this as a "poc", I mean, it works for me but I don't have much time to implement.
I'd rather continue to look for Edge compatibility.
Feel free to continue my pr

kekkokk:
Not exactly, it's job for your load balancer to assign a sticky_session cookie to the first request of the handshake to make sure consequent request are delivered to the same server.
And if would be a barrier, I will consider adding this option in the future.
I know that maybe this is not your current scenario but I'm having lots of problems in strict firewall environments such as banks

mkhahani:
Okay good.
I'm more concerned of blocked UDP ports in strict networks and I'm curious to know what have you done for that since Licode does not support fallback on TCP.

kekkokk:
I simply use a TURN that support TCP on 443.
You need to ensure that the company doesn't block STUN request at firewall application layer and have at least 443 opened. a better solution would be asking to open the 3748 UDP port.

I am currently facing a strange behaviour in Firefox. most of these companies uses proxy.
They work well with chrome but no way to make them work on Firefox.
I think something is missing in Firefox.

mkhahani:
Thanks for the tips.
AFAIK Chrome uses the system proxy while FF has built-in support for proxy.

PR #1218, opened 2018-05-14
[POC] Replace libav with FFmpeg 3.4.2 *do not merge*

zevarito:
This is a POC about switching to FFmpeg due to the reasons that @jnoring exposed in the past #232 , plus a couple more that I've tested.

* Experimenting with other kind of Outputs I was able to accomplish tasks with FFmpeg that didn't work with the same code compiling against LibAV (even using edge LibAV).
* When you look for LibAV documentation, examples, discussions, you always end on FFmpeg related docs, forums etc.
* Ubuntu (mayor supporter before) dropped support of LibAV in favor of FFmpeg.
* LibAV 12.3 (latest, Licode uses 11.6) is still behind FFmpeg 3.2.4 which is updated here, while FFmpeg 4.0 was recently released. Please take a look at what changed in each version to see that are not trivial changes/fixes. https://www.ffmpeg.org.

I would like to know if you guys are up to make this change, if so, I will commit myself to work on remove deprecation warnings an rewrite decode/encode parts to following the latest encoding pipeline[0].

[0] https://ffmpeg.org/doxygen/3.3/group__lavc__encdec.html

[NO] It needs and includes Unit Tests

### DIFF
--- a/erizo/src/erizo/media/ExternalInput.cpp
+++ b/erizo/src/erizo/media/ExternalInput.cpp
@@ -184,7 +184,7 @@ void ExternalInput::receiveLoop() {
     AVPacket orig_pkt = avpacket_;
     if (needTranscoding_) {
       if (avpacket_.stream_index == video_stream_index_) {  // packet is video
-        inCodec_.decodeVideo(avpacket_.data, avpacket_.size, decodedBuffer_.get(), bufflen_, &gotDecodedFrame);
+        inCodec_.decodeVideoBuffer(avpacket_.data, avpacket_.size, decodedBuffer_.get(), bufflen_, &gotDecodedFrame);
         RawDataPacket packetR;
         if (gotDecodedFrame) {
           packetR.data = decodedBuffer_.get();
--- a/erizo/src/erizo/media/ExternalInput.cpp
+++ b/erizo/src/erizo/media/ExternalInput.cpp
@@ -86,12 +86,12 @@ int ExternalInput::init() {
                audio_stream_index_, audio_st->time_base.num, audio_st->time_base.den);
     audio_time_base_ = audio_st->time_base.den;
     ELOG_DEBUG("Audio Time base %d", audio_time_base_);
-    if (audio_st->codec->codec_id == AV_CODEC_ID_PCM_MULAW) {
+    if (audio_st->codecpar->codec_id == AV_CODEC_ID_PCM_MULAW) {
       ELOG_DEBUG("PCM U8");
       om.audioCodec.sampleRate = 8000;
       om.audioCodec.codec = AUDIO_CODEC_PCM_U8;
       om.rtpAudioInfo.PT = PCMU_8000_PT;
-    } else if (audio_st->codec->codec_id == AV_CODEC_ID_OPUS) {
+    } else if (audio_st->codecpar->codec_id == AV_CODEC_ID_OPUS) {
       ELOG_DEBUG("OPUS");
       om.audioCodec.sampleRate = 48000;
       om.audioCodec.codec = AUDIO_CODEC_OPUS;
@@ -102,7 +102,7 @@ int ExternalInput::init() {
   }
 
 
-  if (st->codec->codec_id == AV_CODEC_ID_VP8 || !om.hasVideo) {
+  if (st->codecpar->codec_id == AV_CODEC_ID_VP8 || !om.hasVideo) {
     ELOG_DEBUG("No need for video transcoding, already VP8");
     video_time_base_ = st->time_base.den;
     needTranscoding_ = false;
@@ -110,12 +110,12 @@ int ExternalInput::init() {
     MediaInfo om;
     om.processorType = PACKAGE_ONLY;
     if (audio_st) {
-      if (audio_st->codec->codec_id == AV_CODEC_ID_PCM_MULAW) {
+      if (audio_st->codecpar->codec_id == AV_CODEC_ID_PCM_MULAW) {
         ELOG_DEBUG("PCM U8");
         om.audioCodec.sampleRate = 8000;
         om.audioCodec.codec = AUDIO_CODEC_PCM_U8;
         om.rtpAudioInfo.PT = PCMU_8000_PT;
-      } else if (audio_st->codec->codec_id == AV_CODEC_ID_OPUS) {
+      } else if (audio_st->codecpar->codec_id == AV_CODEC_ID_OPUS) {
         ELOG_DEBUG("OPUS");
         om.audioCodec.sampleRate = 48000;
         om.audioCodec.codec = AUDIO_CODEC_OPUS;
@@ -126,9 +126,9 @@ int ExternalInput::init() {
     op_->init(om, this);
   } else {
     needTranscoding_ = true;
-    inCodec_.initDecoder(st->codec);
+    inCodec_.initDecoder(&video_codec_ctx_, st->codecpar);
 
-    bufflen_ = st->codec->width*st->codec->height*3/2;
+    bufflen_ = video_codec_ctx_->width*video_codec_ctx_->height*3/2;
     decodedBuffer_.reset((unsigned char*) malloc(bufflen_));
 
 
--- a/erizo/src/erizo/media/ExternalInput.cpp
+++ b/erizo/src/erizo/media/ExternalInput.cpp
@@ -27,7 +27,7 @@ ExternalInput::~ExternalInput() {
   thread_.join();
   if (needTranscoding_)
     encodeThread_.join();
-  av_free_packet(&avpacket_);
+  av_packet_unref(&avpacket_);
   if (context_ != NULL)
     avformat_free_context(context_);
   ELOG_DEBUG("ExternalInput closed");
@@ -221,7 +221,7 @@ void ExternalInput::receiveLoop() {
         }
       }
     }
-    av_free_packet(&orig_pkt);
+    av_packet_unref(&orig_pkt);
   }
   ELOG_DEBUG("Ended stream to play %s", url_.c_str());
   running_ = false;
--- a/erizo/src/erizo/media/ExternalOutput.cpp
+++ b/erizo/src/erizo/media/ExternalOutput.cpp
@@ -27,8 +27,8 @@ ExternalOutput::ExternalOutput(std::shared_ptr<Worker> worker, const std::string
     audio_stream_{nullptr}, video_source_ssrc_{0},
     first_video_timestamp_{-1}, first_audio_timestamp_{-1},
     first_data_received_{}, video_offset_ms_{-1}, audio_offset_ms_{-1},
-    need_to_send_fir_{true}, rtp_mappings_{rtp_mappings}, video_codec_{AV_CODEC_ID_NONE},
-    audio_codec_{AV_CODEC_ID_NONE}, pipeline_initialized_{false}, ext_processor_{ext_mappings} {
+    need_to_send_fir_{true}, rtp_mappings_{rtp_mappings}, video_codec_id_{AV_CODEC_ID_NONE},
+    audio_codec_id_{AV_CODEC_ID_NONE}, pipeline_initialized_{false}, ext_processor_{ext_mappings} {
   ELOG_DEBUG("Creating output to %s", output_url.c_str());
 
   fb_sink_ = nullptr;
@@ -114,12 +114,14 @@ void ExternalOutput::syncClose() {
       av_write_trailer(context_);
   }
 
-  if (video_stream_ && video_stream_->codec != nullptr) {
-      avcodec_close(video_stream_->codec);
+  if (video_ctx_ != nullptr) {
+    avcodec_close(video_ctx_);
+    avcodec_free_context(&video_ctx_);
   }
 
-  if (audio_stream_ && audio_stream_->codec != nullptr) {
-      avcodec_close(audio_stream_->codec);
+  if (audio_ctx_ != nullptr) {
+    avcodec_close(audio_ctx_);
+    avcodec_free_context(&audio_ctx_);
   }
 
   if (context_ != nullptr) {
@@ -192,7 +194,7 @@ void ExternalOutput::writeAudioData(char* buf, int len) {
   }
 
   long long timestamp_to_write = (current_timestamp - first_audio_timestamp_) /  // NOLINT
-                                    (audio_stream_->codec->sample_rate / audio_stream_->time_base.den);
+                                    (audio_ctx_->sample_rate / audio_stream_->time_base.den);
   // generally 48000 / 1000 for the denominator portion, at least for opus
   // Adjust for our start time offset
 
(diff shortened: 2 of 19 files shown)


zevarito:
@kekkok yes, read the description ;)

#232 and the bullet list, what other kind of reasons do you have in mind?

kekkokk:
No one reason currently, I just use the standalone ffmpeg for postprocessing.
I think that actually libav gets it's job done well...

zevarito:
@kekkokk well, in the case of post-processing if you are doing some of that simultaneously on the same server, you can take some advantage of having ffmpeg installed instead libav, since it works better, and you might avoid to spawn ffmpeg process, containers, etc. to pipe what is inside Licode writing specific code for that instead. Also I've had bad experience processing MKV with ffmpeg command line while it is being written, that lead me to replace libav and do some stuff from code.

However there are implications about the design and scope of the project itself, I believe Licode want's to stay as SFU and lightweight as possible, but at the same time it is wrote in a way that is fairly easy extensible and modular, so would be great to have certain dependencies updated so anyone could add component features for different purposes.

You are right about it does his job well, so far.

kekkokk:
I think it's never a good idea to start the transconding in a fine that it's being written.
I use separate machines to do the transcoding and the mosaic.
Still having everything updated it's good as you said.

lodoyun:
We originally used libav because it was the main option in Ubuntu at the time as the original idea was to use the apt package version.
We no longer have any reason to use libav instead of ffmpeg though. From what I've read ffmpeg is usually in a better state but, frankly, I'm not keeping up with the development of either of them. 

In general, I would say I'd rather switch to ffmpeg instead of updating libav when we need something from a newer version so this PR looks good in that sense. @kekkokk is there anything in particular you don't like about ffmpeg?

kekkokk:
nono absolutely not, I really like FFmpeg. Probably I badly explained myself.

zevarito:
@lodoyun that's cool, I've closed this PR since it was working fine on OSX but wasn't working on Ubuntu, I'll send a new PR as soon as I can. Thanks!

zevarito:
@lodoyun I've started to update API calls to eliminate all warnings, it looks like a big change touching several files on the project, I'll try to commit as small parts as I can so it gets easy for you to review and give me some feedback, there is a strange behavior that actually forces me to remove all deprecation warnings.

The thing is that when the project is compiled in OSX, it seems to have -Wall flag, but when the same codebase is compiled in Ubuntu it seems to also get -Werror for Erizo which leads to the compile errors noticed in CI.

I believe it could be due a mismatch version of CMake between OSX (3.x.x) and Ubuntu (2.x.x) or by some build configuration of a dependency of the project that introduces that flag, still not clear since I've tried to override it in Ubuntu without success.

jcague:
@zevarito yes, we're aware of such mismatch in the compiler flags between OSX and Ubuntu. Actually, the main reason is that OSX uses clang while in Linux we use gcc, so even having -Werror (which I think would be a good idea) does not ensure you will not have warnings in Linux. Anyway, I also think it sometimes makes it hard to submit PRs to the repo when we're using OSX for development, we will consider alternatives (like using clang in linux, or maybe providing an easy way to develop using the docker container.

lodoyun:
@zevarito Yes, that's about it. The AudioEncoder/AudioDecoder should have been the audio equivalent of VideoEncoder/VideoDecoder and abstract libav(ffmpeg) at that level to be used in MediaProcessor. I really can't tell why that is not the case. It does not make sense to have that code duplicated and if you could use it in `MediaProcessor` it would be great.

zevarito:
@lodoyun That's cool, I will look upon that and put everything there.

About ExternalOuput I have a wip that might be interesting to merge once FFMpeg is fully working, it allows to output to diferent containers, protocols and codecs.

I didn't look deeply into the test suite but I believe lot of things are mocked so I was willing to have a checklist of manual test-cases that ensure all pieces of code are touched, like a guideline for developers to test that everything works once low level parts of the code are changed. Do you guys have something like that right now?

lodoyun:
The ExternalOutput wip sounds interesting, we will probably want to keep using a combination that works by default and possibly add options in the configuration as we test more.

The tests for erizo are almost exclusively unit tests. For instance for LibNice or Nicer we will test that the interface that we use to call the library is the right one, that we are calling the right things at the right time, etc. They are not testing the library itself but it's generally a good indicator. We haven't covered ExternalOutput, for instance, which would be helpful. 

We do have full integration tests [here](https://github.com/lynckia/tests) that are run daily but they don't cover recording and are focused on testing different browsers.

Both need to cover recording to be helpful in this case but I think that would cover what you are asking, right?

zevarito:
@lodoyun I was thinking checklist QA from devs before PR, but that you've pointed is totally helpful.

zevarito:
@lodoyun @jcague Can you briefly take a look to see how this is going?

6d21295

* Add sample format and channels to MediaInfo struct.
* Move MediaInfo, RTPInfo, ProcessorType structs into a single file.
* Abstract audio/video enc/dec from Input/OutputProcessor.
* Abstract basic enc/dec FFmpeg API into Coder class.
* Rename variables to follow cpp code guideline.

Coder is a class that allocate and open contexts, encode and decode in a general way (AVFrame<->AVPacket) as the latest FFmpeg API. It is being used by VideoEncoder/Decoder and AudioEncoder/Decoder as an aggregate class for now. Those classes provide functions to encode and decode from buffers as example.

It's worth to notice that Encoding operation result is made through a Callback, since the encoder might output multiple packets after receive a single frame. Alternatively to this approach I think a FIFO can be used so that once you send a Frame to encoding you drain the FIFO of result packets.

Are you running `test` project standalone or replicating CircleCI locally?

erizo/src/erizo/media/ExternalOutput.cpp
```diff
@@ -369,62 +409,69 @@ int ExternalOutput::deliverEvent_(MediaEventPtr event) {
...
-    audio_stream_ = avformat_new_stream(context_, audio_codec);
-    audio_stream_->id = 1;
-    audio_stream_->codec->codec_id = audio_codec_;
-    audio_stream_->codec->sample_rate = audio_map_.clock_rate;
-    audio_stream_->time_base = (AVRational) { 1, audio_stream_->codec->sample_rate };
-    audio_stream_->codec->channels = audio_map_.channels;
-    if (context_->oformat->flags & AVFMT_GLOBALHEADER) {
-      audio_stream_->codec->flags |= CODEC_FLAG_GLOBAL_HEADER;
-    }
+      video_decoder_.initDecoder(input_video_codec_id_,
```
lodoyun: We don't need to initialize the decoders unless we're transcoding, right?

scripts/installUbuntuDeps.sh
```diff
@@ -157,16 +157,12 @@ install_mediadeps(){
   sudo apt-get -qq install yasm libvpx. libx264.
   if [ -d $LIB_DIR ]; then
     cd $LIB_DIR
-    if [ ! -f ./v11.1.tar.gz ]; then
```
lodoyun: We need this check also for ffmpeg to make sure we download and compile this when it's already there

scripts/installUbuntuDeps.sh
```diff
@@ -180,16 +176,12 @@ install_mediadeps_nogpl(){
   sudo apt-get -qq install yasm libvpx.
   if [ -d $LIB_DIR ]; then
     cd $LIB_DIR
-    if [ ! -f ./v11.1.tar.gz ]; then
```
lodoyun: Same as above

scripts/travisInstallDeps.sh
```diff
@@ -132,16 +132,12 @@ install_mediadeps(){
   sudo apt-get -qq install yasm libvpx. libx264.
   if [ -d $LIB_DIR ]; then
     cd $LIB_DIR
-    if [ ! -f ./v11.1.tar.gz ]; then
```
lodoyun: Again :)

scripts/travisInstallDeps.sh
```diff
@@ -154,16 +150,12 @@ install_mediadeps_nogpl(){
   sudo apt-get -qq install yasm libvpx.
   if [ -d $LIB_DIR ]; then
     cd $LIB_DIR
-    if [ ! -f ./v11.1.tar.gz ]; then
```
lodoyun: And again

erizo/src/erizo/media/codecs/Coder.cpp
```diff
@@ -0,0 +1,238 @@
...
+      done(av_packet, frame, true);
+    } else {
+      ELOG_ERROR("avcodec_receive_packet failed. %s, %d %s", encode_ctx->codec->name,
+          ret, av_err2str_cpp(ret));
+      done(av_packet, frame, false);
+    }
+  }
+}
+
+void Coder::saveFrameAsJPEG(AVFrame *frame) {
```
lodoyun: Do we really need this?

erizo/src/erizo/media/codecs/Coder.h
```diff
@@ -0,0 +1,96 @@
...
+
+ public:
+  CoderCodec();
+  virtual ~CoderCodec();
+  virtual int closeCodec();
+  void logCodecContext();
+
+ public:
+  bool initialized;
+  AVCodec* av_codec_;
```
lodoyun: I think all of these should be protected. In general I wouldn't want to access the pointers directly from outside of the `CoderEncoder` or `CoderDecoder` if possible.

erizo/src/erizo/media/codecs/Coder.h
```diff
@@ -0,0 +1,96 @@
...
+  DECLARE_LOGGER();
+
+ public:
+  CoderCodec();
+  virtual ~CoderCodec();
+  virtual int closeCodec();
+  void logCodecContext();
+
+ public:
+  bool initialized;
```
lodoyun: This one is a nitpick. Consider making initalized private and adding a getter method instead - `isInitialized()`

erizo/src/erizo/media/ExternalInput.cpp
```diff
@@ -126,9 +126,9 @@ int ExternalInput::init() {
     op_->init(om, this);
   } else {
     needTranscoding_ = true;
-    inCodec_.initDecoder(st->codec);
+    inCodec_.initDecoder(st->codecpar);
 
-    bufflen_ = st->codec->width*st->codec->height*3/2;
+    bufflen_ = inCodec_.codec_context_->width*inCodec_.codec_context_->height*3/2;
```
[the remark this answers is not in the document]
zevarito: Well, ExternalInput is a particular thing, since I've started there before realize the common API (Coder, CoderCodec, etc.) should be used, then I've updated it but not deeply read and refactor the code, so consider it pending for now please. I'll post bellow a list of the things I am considering that needs to be finished.

erizo/src/erizo/media/ExternalOutput.cpp
```diff
@@ -102,36 +103,33 @@ void ExternalOutput::close() {
...
+    audio_encoder_.closeCodec();
   }
-
   if (context_ != nullptr) {
-      avio_close(context_->pb);
-      avformat_free_context(context_);
-      context_ = nullptr;
+    avio_close(context_->pb);
   }
+  avformat_free_context(context_);
```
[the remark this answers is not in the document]
zevarito: This is a mistake 👍

erizo/src/erizo/media/ExternalOutput.cpp
```diff
@@ -102,36 +103,33 @@ void ExternalOutput::close() {
 }
 
 void ExternalOutput::syncClose() {
-  if (!recording_) {
```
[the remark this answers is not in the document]
zevarito: Well, I thought `syncClose` should be idempotent, but honestly didn't repair on the details. I'll put it back as it was. Can you point me out the scenario where `ExternalOuput::close()` is called more than once?

Also, I would like to ask you how we can deal with publisher interruptions like browser refresh where seems that ExternalOutput is not closed properly.

erizo/src/erizo/media/ExternalOutput.cpp
```diff
@@ -369,62 +409,69 @@ int ExternalOutput::deliverEvent_(MediaEventPtr event) {
...
-    // A decent guess here suffices; if processing the file with ffmpeg,
-    // use -vsync 0 to force it not to duplicate frames.
-    video_stream_->codec->pix_fmt = PIX_FMT_YUV420P;
-    if (context_->oformat->flags & AVFMT_GLOBALHEADER) {
-      video_stream_->codec->flags |= CODEC_FLAG_GLOBAL_HEADER;
-    }
-    context_->oformat->flags |= AVFMT_VARIABLE_FPS;
+    if (avio_open(&context_->pb, context_->filename, AVIO_FLAG_WRITE) < 0) {
+      ELOG_ERROR("Error opening output context.");
+      exit(1);
```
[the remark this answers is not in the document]
zevarito: Too extreme right? :)

I'll remove the `exit` for now, let me know your thoughts about how to handle it with ErizoJS.

erizo/src/erizo/media/ExternalOutput.cpp
zevarito: Yes, that's right we should be able to avoid initialize decoders in some cases.
I didn't want to go deeper here until get things done and define surrounding API's, since didn't look as much resource waste, for example, in the case of Output MKV, it seems to only increase ~15mb of memory when you are recording.

erizo/src/erizo/media/codecs/Coder.cpp
zevarito: Well, actually not, I'll remove.

erizo/src/erizo/media/codecs/Coder.h
zevarito: Ok, you got it. I'll move `addOutputStream` logic inside CoderCodec and protect all that things.

zevarito: @lodoyun sadly I'll leave accesible `codec_context_` and `frame_` a little bit more. The thing is that hide those almost require a complete layer of abstraction of ffmpeg, which I don't know if it is the intention or not, I have started this abstraction trying to avoid repetition like open/close codecs and context and off-course decoding/encoding logic but still relaying on certain ffmpeg functions. I am not telling which is the right way to go since I didn't clarify on my mind either, but before go through that path I believe it require an specific discussion about it and how to achieve it.

zevarito:
@lodoyun I've addressed most of your concerns except for the one I've commented about.
I've also committed a change to `erizoController/Client` `startRecording` function to allow to pass a different extension or full url with the ability of templating `recordingId`, this change is backward compatible but I had not documented this since I don't think is a final API, it just a thing to easy test EO.

erizo/src/erizo/media/ExternalOutput.cpp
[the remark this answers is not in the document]
lodoyun: Probably the best way is to do something similar to what we're doing with MediaStream events and then send any fatal error as an event via a callback and handle it in ErizoJS. See [MediaStream.cpp](https://github.com/lynckia/licode/blob/master/erizo/src/erizo/MediaStream.cpp), [MediaStream.cc] (https://github.com/lynckia/licode/blob/master/erizo/src/erizoAPI/MediaStream.cc) and [Connection.js](https://github.com/lynckia/licode/blob/master/erizo_controller/erizoJS/models/Connection.js#L91)
You probably don't need to do that in this PR, it's a bit involved and we would also need a consistent response to this problem.

erizo/src/erizo/media/codecs/Coder.h
lodoyun: Fair enough. We can keep it like this for now.

erizo_controller/erizoClient/src/Room.js
```diff
@@ -682,14 +682,22 @@ const Room = (altIo, altConnectionHelpers, altConnectionManager, specInput) => {
   };
 
   // Returns callback(id, error)
-  that.startRecording = (stream, callback = () => {}) => {
+  that.startRecording = (stream, callbackOrOptions) => {
```
lodoyun: Even if it's an API just to test things, I'd rather have 
```
that.startRecording = (stream, callback, advancedOptions=undefined) => {
```
We can decide later at what point and in what format we can expose that options in the API and update the docs accordingly.

zevarito: @lodoyun yes, totally, I've even forgot you can add arbitrary params to a js fn, I'll update as soon as I can.

peererror:
@zevarito  what is the status of this pull is it going to merge in main repo

zevarito:
> @zevarito what is the status of this pull is it going to merge in main repo

I think it was ready... 10 months ago :D, right now I don't think so.

lodoyun:
I've updated and tested this in [this branch](https://github.com/lodoyun/licode/tree/zevarito-move_to_ffmpeg_3_4). Unfortunately, there are some issues I think related to the rescaling of the timestamps when not transcoding that prevent me from merging it, the result is that no audio codec works out of the box, all need postprocessing. I have other more pressing things to do atm, but I'll get back to this in the near future

zevarito:
Hey @lodoyun looking around to see in what this pr was, after read your last comment I realize that this commit might be related: f6494e1195c8830987dfdb658034151cdf0d4f48.

Also, of 32 commits from the PR I think these are the most relevant:
6d2129523f75fe8ceb66a067d61d8b759cd46690
7700db04319ff74ef4861599189e24d736b96888
595c9003cc17798cc45111abd3d935606671a149

zevarito:
@lodoyun any update on this PR? did you guys have merged/implemented parts of this somewhere?

lodoyun:
Sadly it's not ready. I have a branch locally that is a lot closer to the current head but I run into some problems with the audio codec, which always needed postprocessing (whereas now, we don't always need it) and it fell through the priority list.
I still want to merge it so thank you for the ping. I'll see if I can allocate some time for this soon.

PR #1222, opened 2018-05-21
Updated erizo.js compilation script to respect ES6 syntax

simonemazzoni:
**Description**

Hello, I updated client erizo.js script to build it as umd module to respect ES6 syntax and make it importable correctly.

Not changed

I also added socket.io-client npm module installation to replace the socket.io-client version manually downloaded. This approach is more mantainable in the future if we want to update the library.

PS: still don't completely understand the logic of the client build script. I see you introduced webpack, but a lot of stuff regarding erizo.js client is done using gulp whereas can be done using webpack directly. I also don't understand the need to use the `google-closeure-compiler` to compile the lib. Could someoune explain that choice to me?

Let me know if the PR is fine and can be merged.

Thanks.
Simone

### DIFF
--- a/erizo_controller/erizoClient/src/Socket.js
+++ b/erizo_controller/erizoClient/src/Socket.js
@@ -1,6 +1,6 @@
 /* global */
 
-import io from 'socket.io-client';
+import io from 'socket.io-client'; // eslint-disable-line
 import Logger from './utils/Logger';
 
 import { EventDispatcher, LicodeEvent } from './Events';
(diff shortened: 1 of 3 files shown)


simonemazzoni:
UPDATE: I noticed circleCI failed due to eslint checks on client not satisfied due to the fact that `socket.io-client` isn't listed in package.json. It should be fixed now.
Again, I don't understand what the npm modules are not installed using a package.json (or multiple package.json if you prefer to split dependencies among licode components)

simonemazzoni:
I close this since the topic is covered by [https://github.com/lynckia/licode/pull/1224](https://github.com/lynckia/licode/pull/1224)

PR #1224, opened 2018-05-22
Refactor client build process

kekkokk:
**Description**

I made some changes based on my colleague suggestion https://github.com/lynckia/licode/pull/1222
- Make a compatible webpack module so we can now include the erizo module in an ES6 project
- Possibility to compile the library without socket.io or adapter bundled in erizo, to significantly save space in the output bundle (there already was the option to use an altIo library, so we doesn't need to always include it).
- Added a script to compile client in various ways (includes my PR https://github.com/lynckia/licode/pull/1221)


erizo_controller/erizoClient/src/Erizo.js
```diff
@@ -3,14 +3,20 @@ import { LicodeEvent, RoomEvent, StreamEvent } from './Events';
 import Stream from './Stream';
 import Logger from './utils/Logger';
 
-// eslint-disable-next-line 
-require('expose-loader?adapter!../lib/adapter.js');
-// eslint-disable-next-line
-require('script-loader!./utils/L.Resizer.js');
+// #if process.env.INCLUDE_ADAPTER == 'TRUE'
```
lodoyun: Why not make it the opposite and remove the import when a variable is present? I'd prefer the default behavior to stay the same

erizo_controller/erizoClient/src/Erizo.js
```diff
@@ -3,14 +3,20 @@ import { LicodeEvent, RoomEvent, StreamEvent } from './Events';
...
+import adapter from 'webrtc-adapter'; // eslint-disable-line
+// #endif
+
+require('script-loader!./utils/L.Resizer.js');  // eslint-disable-line
 
 
 const Erizo = {
+  // #if process.env.INCLUDE_SOCKETIO == 'TRUE'
```
lodoyun: Same as above

erizo_controller/erizoClient/src/Erizo.js
```diff
@@ -3,14 +3,20 @@ import { LicodeEvent, RoomEvent, StreamEvent } from './Events';
...
   Room: Room.bind(null, undefined, undefined, undefined),
+  // #endif
+  // #if process.env.INCLUDE_SOCKETIO != 'TRUE'
```
lodoyun: I assume `#else` would work here. Consider using it

erizo_controller/installErizo_controller.sh
```diff
@@ -21,13 +21,13 @@ check_result() {
 echo [erizo_controller] Installing node_modules for erizo_controller
 
 nvm use
-npm install --loglevel error amqp socket.io@2.0.3 log4js@1.0.1 node-getopt uuid@3.1.0 sdp-transform@2.3.0
+npm install --loglevel error amqp socket.io@2.0.3 log4js@1.0.1 node-getopt uuid@3.1.0 sdp-transform@2.3.0 socket.io-client@2.1.0 webrtc-adapter@6.1.5 webpack-conditional-loader@1.0.12
```
lodoyun: `socket.io-client@2.1.0` -> consider using `socket.io-client@2.0.3` which is the one we're using in spine.

scripts/compileClient.sh
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env bash
```
lodoyun: I'd put this file in `/erizo_controller/erizoClient` and again, the default should be including the needed libraries. That way Licode would work outside of the box, even if you run gulp directly.

erizo_controller/erizoClient/webpack.config.erizo.js
```diff
@@ -7,7 +7,12 @@ module.exports = {
     path: path.resolve(__dirname, 'dist'),
     libraryExport: 'default',
     library: 'Erizo',
-    libraryTarget: 'var',
+    libraryTarget: 'umd',
```
lodoyun: This breaks compatibility in some environments, depending on how you import `erizo.js`.  Using var is the most compatible alternative and that's why it's the option we went with.

erizo_controller/erizoClient/src/Erizo.js
kekkokk: I thought so too, but nope :/

erizo_controller/installErizo_controller.sh
kekkokk: ok np

scripts/compileClient.sh
kekkokk: yeah it makes sense

erizo_controller/erizoClient/webpack.config.erizo.js
kekkokk: I'm not an expert, @simonemazzoni what do you think?

simonemazzoni: @lodoyun I'm aware that export the lib as UMD could break compatibility in some environments, but it enable the usage of new ES6 import/export functionalities which are standard now. I propose to enhance the webpack build script to produce multiple library target files? Maybe using something like `parallel-webpack`? This allows to import erizo.js code in user code, avoiding to expose it as script in the window (possibly masking it's usage and making easier to handle resource caching and update).

erizo_controller/installErizo_controller.sh
simonemazzoni: Why don't we also update spine socket.io version instead? Version 2.1.1 introduced only bug fixes with respect the 2.0.3, isn't it better to keep things updated?

scripts/compileClient.sh
kekkokk: It makes sense for the current structure, but I'd prefer to move ALL the scripts in the script folder.
I think it's way more clear to the people that approach Licode for the first time

kekkokk:
I had to move the compileClient script in erizo_controller 'cause the checkNvm script has the script path hardcoded:

LINKED_DIR=$CHECK_SCRIPT_PATH/../build/libdeps/nvm

erizo_controller/erizoClient/webpack.config.erizo.js
lodoyun: @simonemazzoni I understand but I would prefer gradually transitioning to UMD. I do like the `parallel-webpack` variants option to enable the option but keeping it as `'var'` only by default.

We could, for instance, have a configuration file where you could set up an array with different targets. However if that file is not preset, `var` would be the default. That would allow us to transition to var+umd when we're ready to support both.

simonemazzoni: Yes, that solution sounds good to me. Do you have any idea of where to put that config?

erizo_controller/installErizo_controller.sh
lodoyun: Yes, but in it's own PR, please.

erizo_controller/erizoClient/webpack.config.erizo.js
lodoyun: In erizoClient, where all the build config is. I think is the more sensible place.

kekkokk:
I'm not a big fan to configurations, since the project is already full of extra config options.
To keep things easier, since gulp does these jobs in parallel (so it won't take more time), it compiles the var client AND the umd one.

Tell me if you prefer another way

erizo_controller/erizoClient/webpack.config.erizofc.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     path: path.resolve(__dirname, 'dist'),
     libraryExport: 'default',
     library: 'Erizo',
-    libraryTarget: 'umd',
+    libraryTarget: 'var',
```
lodoyun: erizofc should be umd, it's a different target. This is only used in node, mostly for Spine and umd is fine in this context

kekkokk: ok, tought you'd prefer var

lodoyun:
What jobs does gulp in parallel? Do you mean using`parallel-webpack`?  We are not currently parallelizing any of the gulp tasks. 
Anyway, the reason for not building the umd client by default is not time (although it could be a factor). The reason is that we won't be using that version at this point and there are no integration tests that ensure that the umd version works so we won't be supporting it for now. I'm ok for people to put it in a configuration and use it at their own risk but I don't wan't to give the impression that we will be solving issues caused by transitioning to umd when we won't right now.
We will switch to umd at some point, and it will probably make sense to have both in there as a transition while we deprecate the `var` version.

kekkokk:
ok gotcha.
I'll probably put this configuration in the main licode_config.js since it's in the gitignore and so it won't conflict everytime with the pulls.
Or do you still prefer to have this config inside the client folder?

lodoyun:
I do prefer having it in ErizoClient since it's a compilation option and all of the options in `licode_config.js` are runtime options. 
I still think what I suggested to  simonemazzoni is a good idea:

``` We could, for instance, have a configuration file where you could set up an array with different targets. However if that file is not present, var would be the default. That would allow us to transition to var+umd when we're ready to support both.```

In any case, it should not conflict with the repository. We can also have a default file that is copied when licode is installed and then can be modified freely (just like licode_config).

kekkokk:
now with install_erizo.sh a ignored build client file is copied in the ./compileClient directory and you can choose if the build process should compile erizo.js in var or umd mode.

the problem is that we added lot of extra files just to compile the umd, when at this point was easier to change var to umd in the webpack config. Maybe that webpack config could be added to gitignore. 

Tell me if you like as it is or revert to the original approach. Or feel free to modify my pr

lodoyun:
I don't like having a new file  `erizo_controller/erizoClient/gulp/erizoUmdTasks.js`. Maybe I got confused with the conversation, I thought the proposal was to use `parallel-webpack` to create an optional variant that would compile umd.  That variant could be optionally added via a config file. I don't want all the targets to be an option, just variants of the erizo compilation for the client (never `erizojs` for spine), if possible.

If that's not possible, I would rework this PR to only include the changes to the way adapter.js and socket.io are handled.

kekkokk:
Yeah, so just keep the new way they are included. 
forget about umd right now.
I am very busy these days, so it's all yours!

PR #1226, opened 2018-05-22
Update build script to automatically detect number of CPUs

kekkokk:
Avoiding calling cmake at every ./installErizo.sh allow us to build only the changes that are made between every make and speed up our developing process.

If we want to clear cache with can always run installErizo with -d option

### DIFF
--- a/erizo/generateProject.sh
+++ b/erizo/generateProject.sh
@@ -10,13 +10,11 @@ BASE_BIN_DIR="build"
 generateVersion() {
   echo "generating $1"
   BIN_DIR="$BASE_BIN_DIR/$1"
-  if [ -d $BIN_DIR ]; then
-    cd $BIN_DIR
-  else
+  if [ ! -d $BIN_DIR ]; then
     mkdir -p $BIN_DIR
     cd $BIN_DIR
+    cmake ../../src "-DERIZO_BUILD_TYPE=$1"
   fi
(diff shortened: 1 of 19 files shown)


erizo/generateProject.sh
```diff
@@ -10,13 +10,11 @@ BASE_BIN_DIR="build"
 generateVersion() {
   echo "generating $1"
   BIN_DIR="$BASE_BIN_DIR/$1"
-  if [ -d $BIN_DIR ]; then
-    cd $BIN_DIR
-  else
+  if [ ! -d $BIN_DIR ]; then
```
jcague: what's the issue with the previous way?

kekkokk: in this way cmake is called only the first time that the folder is created after a cleaonOject is called.
I noticed that if cmake is run every time generateProject is called, it build the enteire project every time instead of only the changes made in cpp.

jcague: Just run buildProject.sh if you implement changes inside existing cpp files, or run generateProject.sh if you add/remove files or dependencies. With this change it won't regenerate cmake structure if the directory already exists.

kekkokk:
I left all the things needed to do a "power" compile

jcague:
I don't remember exactly, we probably changed CMakeLists in some way by adding more deps, and it now recreates everything from scratch.

PR #988, opened 2017-07-28
Fix/erizo timeout retry

kekkokk:
**Description**
[ POC ]
REF# https://github.com/lynckia/licode/issues/969

[x] It needs and **DOESN'T INCLUDE** Unit Tests

**Changes in Client or Server public APIs**

switch the logic of retry client side

[] It includes documentation for these changes in `/doc`.

PR #1229, opened 2018-05-28
Avoid server side retry and move logic to client

kekkokk:
**Description**

REF https://github.com/lynckia/licode/pull/988

Move the retry logic of the addPublisher or addSubscriber clientside.
This avoid an inconsistent state on the server if during retry the client disconnects. (the server continues thinking forever that there's a connected client and this prevents the JS to shutdown and restart when empty)

I fixed the current unit tests.


jcague:
I'm not a huge fan of changing this behaviour. It seems to make like it would me safer ifn we add a mechanism to stop retrying when the client disconnects. But I might be wrong.

kekkokk:
I think it would be way more trivial, but in every case this should be fixed IMO.
when lots of clients hit the server is very easy to end up with a "dirty" structure in my experience

PR #1233, opened 2018-05-31
Fix installMacDeps high sierra 10.13.4

kekkokk:
just a simple fix for latest high sierra

### DIFF
--- a/scripts/installMacDeps.sh
+++ b/scripts/installMacDeps.sh
@@ -84,8 +84,7 @@ install_homebrew(){
   if [ "$CACHE" == "true" ]; then
     install_homebrew_from_cache
   fi
-  which -s brew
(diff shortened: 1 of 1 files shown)


scripts/installMacDeps.sh
```diff
@@ -84,8 +84,7 @@ install_homebrew(){
   if [ "$CACHE" == "true" ]; then
     install_homebrew_from_cache
   fi
-  which -s brew
-  if [[ $? != 0 ]] ; then
+  if ! which -s brew; then
```
jcague: can you provide more info about the issue you had please?

kekkokk: in a clean install, sierra 10.13.4, when clicking enter on installing Homebrew nothing happens and exit the script...
It's quite strange but this fixed for me.

PR #1234, opened 2018-05-31
Rework client reconnection

kekkokk:
I didn't get the point of the closeCode hack in the client.
I mean, serverside makes sense, but client doesn't.

The problem is that that workaround doesn't behave equally in all browsers and os.
for example, firefox 60 in windows 10, when whe disconnect the ethernet adapter it immediately send a transport_error and generates a disconnect from the room.
the same test done in chrome, generates a transport_close and it properly tries to re establish the connection.

So in a few words, the events emitted from socket.io are completely different (and in a different order) among browsers.

With this simple rework we should have a consistent result.

Also, if you like the approach, would be great if you help me with the tests

[ X ] it needs and DOES NOT include tests

### DIFF
--- a/erizo_controller/erizoClient/src/Socket.js
+++ b/erizo_controller/erizoClient/src/Socket.js
@@ -24,7 +24,6 @@ const Socket = (newIo) => {
   that.RECONNECTING = Symbol('reconnecting');
   that.DISCONNECTED = Symbol('disconnected');
 
-  const WEBSOCKET_NORMAL_CLOSURE = 1000;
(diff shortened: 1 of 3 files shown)


kekkokk:
what do you mean by normal disconnections?

jcague:
by normal disconnection I mean cases where we intentionally disconnect the websocket either from client or in server side.

kekkokk:
Nono, if socket.disconnect() is called serverside or clientside the retry mechanism will NOT fire.
Just tested to be sure

PR #1237, opened 2018-06-06
unsafe under high concurrency

luoshaohua:
after wrtc.close(), need update publishers immediately

**Description**


luoshaohua:
if removeExternalOutputs error, you wish delete publishers[streamId] ? I'm not sure, it's up to you. But after wrtc.close(), wrtc is 0x00 in c++, any call wrtc.xxx will crash

jcague:
You're right, thanks a lot!

PR #1238, opened 2018-06-06
Added pipeline_ shared_ptr check before dereferencing it.

Equod:
**Made pipeline usage in erizo::MediaStream more safe**

Using shared pointer (pipeline_) directly after std::shared_ptr::reset() (in erizo::MediaStream::syncClose()) may be unsafe. A simple check before dereferencing it makes the task execution safe even after syncClose task.

### DIFF
--- a/erizo/src/erizo/MediaStream.cpp
+++ b/erizo/src/erizo/MediaStream.cpp
@@ -95,8 +95,9 @@ void MediaStream::setMaxVideoBW(uint32_t max_video_bw) {
   asyncTask([max_video_bw] (std::shared_ptr<MediaStream> stream) {
     if (stream->rtcp_processor_) {
       stream->rtcp_processor_->setMaxVideoBW(max_video_bw * 1000);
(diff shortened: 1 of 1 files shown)


erizo/src/erizo/MediaStream.cpp
```diff
@@ -146,8 +147,8 @@ bool MediaStream::setRemoteSdp(std::shared_ptr<SdpInfo> sdp) {
     this->rtcp_processor_->setMaxVideoBW(remote_sdp_->videoBandwidth*1000);
   }
 
-  if (pipeline_initialized_) {
-    pipeline_->notifyUpdate();
+  if (pipeline_initialized_ && pipeline_) {
+      pipeline_->notifyUpdate();
```
jcague: wrong indentation

jcague:
Thanks for the contribution!

I merged this PR despite of having issues in CircleCI, because they're not related are caused by a bug in our CircleCI scripts. I'll try to fix them asap.

PR #1239, opened 2018-06-07
Update adapter.js to v6.2.1

wolmi:
**Description**
As 13 releases of difference on the actual adapter.js imported on the project statically we need to update to fix some issues related to new versions of Firefox and Safari.

### DIFF
--- a/erizo_controller/erizoClient/lib/adapter.js
+++ b/erizo_controller/erizoClient/lib/adapter.js
@@ -11,6 +11,16 @@
 
 var SDPUtils = require('sdp');
 
+function fixStatsType(stat) {
+  return {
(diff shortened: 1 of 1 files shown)


kekkokk:
ok, did you encounter some specific problems?

jcague:
I'm closing this PR because @kekkokk is working on a new way of building the client by downloading adapterjs from NPM, whic is a really good idea (PR #1224). So the way to update adapterjs will be different. Can you check whether there is some important change in adapterjs within the latest minor versions? If so, @kekkokk can you update it in your PR?

kekkokk:
sure! I think I already included the latest version but I'll check

PR #1244, opened 2018-06-11
Added mute stream event

kekkokk:
**Description**
Currently the mute event is not well implemented and there are some differences between p2p and Erizo.
This PR keeps in sync the mute state of a stream, locally and remotely.

This is very important since in the frontend we may need to replace the "black screen" with messages informing the client that the local/remote stream is muted and cannot be displayed.

I also thought to use the current updatespec event to propagate this information but there are lots of complications (for example it is not propagated to the controller if a peerconnection does not exist), so I preferred to make a dedicated event

[X] It includes documentation for these changes in `/doc`.


jcague:
This looks to me like a new feature, so it would be better if you fill a [feature review doc explaining the changes](http://lynckia.com/licode/contribute.html#features). You can find an example we used here: https://github.com/lynckia/licode/tree/master/feature-review

PR #1247, opened 2018-06-12
Small code review like avoiding redundant objects copy.

Equod:
**PARTIAL CODE REVIEW**

Basically the main change is adding Handlers to the pipeline as "make_shared" pointer.

### DIFF
--- a/erizo/src/erizo/DtlsTransport.cpp
+++ b/erizo/src/erizo/DtlsTransport.cpp
@@ -368,8 +368,8 @@ void DtlsTransport::processLocalSdp(SdpInfo *localSdp_) {
   ELOG_DEBUG("%s message: processing local sdp, transportName: %s", toLog(), transport_name.c_str());
   localSdp_->isFingerprint = true;
   localSdp_->fingerprint = getMyFingerprint();
-  std::string username (ice_->getLocalUsername());
-  std::string password (ice_->getLocalPassword());
+  std::string username(ice_->getLocalUsername());
+  std::string password(ice_->getLocalPassword());
   if (bundle_) {
     localSdp_->setCredentials(username, password, VIDEO_TYPE);
     localSdp_->setCredentials(username, password, AUDIO_TYPE);
--- a/erizo/src/erizo/DtlsTransport.cpp
+++ b/erizo/src/erizo/DtlsTransport.cpp
@@ -127,7 +127,7 @@ DtlsTransport::~DtlsTransport() {
 
 void DtlsTransport::start() {
   ice_->setIceListener(shared_from_this());
-  ice_->copyLogContextFrom(this);
+  ice_->copyLogContextFrom(*this);
   ELOG_DEBUG("%s message: starting ice", toLog());
   ice_->start();
 }
@@ -279,10 +279,10 @@ void DtlsTransport::onHandshakeCompleted(DtlsSocketContext *ctx, std::string cli
   boost::mutex::scoped_lock lock(sessionMutex_);
   std::string temp;
 
-  if (rtp_timeout_checker_.get() != NULL) {
+  if (rtp_timeout_checker_) {
     rtp_timeout_checker_->cancel();
   }
-  if (rtcp_timeout_checker_.get() != NULL) {
+  if (rtcp_timeout_checker_) {
     rtcp_timeout_checker_->cancel();
   }
 
@@ -316,14 +316,14 @@ void DtlsTransport::onHandshakeCompleted(DtlsSocketContext *ctx, std::string cli
   }
 }
 
-void DtlsTransport::onHandshakeFailed(DtlsSocketContext *ctx, const std::string error) {
+void DtlsTransport::onHandshakeFailed(DtlsSocketContext *ctx, const std::string& error) {
   ELOG_WARN("%s message: Handshake failed, transportName:%s, openSSLerror: %s",
             toLog(), transport_name.c_str(), error.c_str());
   running_ = false;
   updateTransportState(TRANSPORT_FAILED);
 }
 
-std::string DtlsTransport::getMyFingerprint() {
+std::string DtlsTransport::getMyFingerprint() const {
   return dtlsRtp->getFingerprint();
 }
 
@@ -368,8 +368,8 @@ void DtlsTransport::processLocalSdp(SdpInfo *localSdp_) {
   ELOG_DEBUG("%s message: processing local sdp, transportName: %s", toLog(), transport_name.c_str());
   localSdp_->isFingerprint = true;
   localSdp_->fingerprint = getMyFingerprint();
-  std::string username = ice_->getLocalUsername();
-  std::string password = ice_->getLocalPassword();
+  std::string username (ice_->getLocalUsername());
+  std::string password (ice_->getLocalPassword());
   if (bundle_) {
     localSdp_->setCredentials(username, password, VIDEO_TYPE);
     localSdp_->setCredentials(username, password, AUDIO_TYPE);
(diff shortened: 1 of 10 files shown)


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -737,7 +737,7 @@ void WebRtcConnection::syncWrite(std::shared_ptr<DataPacket> packet) {
 }
 
 void WebRtcConnection::setTransport(std::shared_ptr<Transport> transport) {  // Only for Testing purposes
-  video_transport_ = transport;
+  video_transport_ = std::move(transport);
```
jcague: In general I'm a bit reluctant to use std::move() in places that aren't hot paths because it's too easy to introduce bugs if we try to use the pointer later in the code, am I right?

erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -279,10 +279,10 @@ void DtlsTransport::onHandshakeCompleted(DtlsSocketContext *ctx, std::string cli
   boost::mutex::scoped_lock lock(sessionMutex_);
   std::string temp;
 
-  if (rtp_timeout_checker_.get() != NULL) {
+  if (rtp_timeout_checker_) {
```
jcague: 👍

erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -316,14 +316,14 @@ void DtlsTransport::onHandshakeCompleted(DtlsSocketContext *ctx, std::string cli
...
-void DtlsTransport::onHandshakeFailed(DtlsSocketContext *ctx, const std::string error) {
+void DtlsTransport::onHandshakeFailed(DtlsSocketContext *ctx, const std::string& error) {
   ELOG_WARN("%s message: Handshake failed, transportName:%s, openSSLerror: %s",
             toLog(), transport_name.c_str(), error.c_str());
   running_ = false;
   updateTransportState(TRANSPORT_FAILED);
 }
 
-std::string DtlsTransport::getMyFingerprint() {
+std::string DtlsTransport::getMyFingerprint() const {
```
jcague: 👍

erizo/src/erizo/DtlsTransport.cpp
```diff
@@ -316,14 +316,14 @@ void DtlsTransport::onHandshakeCompleted(DtlsSocketContext *ctx, std::string cli
   }
 }
 
-void DtlsTransport::onHandshakeFailed(DtlsSocketContext *ctx, const std::string error) {
+void DtlsTransport::onHandshakeFailed(DtlsSocketContext *ctx, const std::string& error) {
```
jcague: 👍

erizo/src/erizo/MediaStream.cpp
```diff
@@ -203,41 +203,41 @@ void MediaStream::initializePipeline() {
...
-  pipeline_->addFront(RtcpFeedbackGenerationHandler());
-  pipeline_->addFront(RtpRetransmissionHandler());
-  pipeline_->addFront(SRPacketHandler());
-  pipeline_->addFront(SenderBandwidthEstimationHandler());
-  pipeline_->addFront(LayerDetectorHandler());
-  pipeline_->addFront(OutgoingStatsHandler());
-  pipeline_->addFront(PacketCodecParser());
-
-  pipeline_->addFront(PacketWriter(this));
+  pipeline_->addFront(std::make_shared<PacketReader>(this));
```
jcague: 👍

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -455,7 +455,7 @@ std::string WebRtcConnection::getJSONCandidate(const std::string& mid, const std
 
   std::ostringstream theString;
   theString << "{";
-  for (std::map<std::string, std::string>::iterator it = object.begin(); it != object.end(); ++it) {
+  for (std::map<std::string, std::string>::const_iterator it = object.begin(); it != object.end(); ++it) {
```
jcague: 👍

erizo/src/erizo/dtls/DtlsClient.cpp
```diff
@@ -309,10 +309,10 @@ int createCert(const std::string& pAor, int expireDays, int keyLen, X509*& outCe
     }
 
 
-    std::string DtlsSocketContext::getFingerprint() {
-      char fprint[100];
+    std::string DtlsSocketContext::getFingerprint() const {
+      char fprint[100] = {};
       mSocket->getMyCertFingerprint(fprint);
-      return std::string(fprint, strlen(fprint));
+      return std::string(fprint);
```
jcague: 👍

erizo/src/erizo/logger.h
```diff
@@ -35,21 +35,21 @@ class LogContext {
   LogContext() : context_log_{""} {
   }
 
-  virtual ~LogContext() {}
+  virtual ~LogContext() = default;
```
jcague: 👍

erizo/src/erizo/media/ExternalOutput.cpp
```diff
@@ -316,10 +316,10 @@ void ExternalOutput::initializePipeline() {
   pipeline_->addService(quality_manager_);
   pipeline_->addService(stats_);
 
-  pipeline_->addFront(LayerBitrateCalculationHandler());
-  pipeline_->addFront(QualityFilterHandler());
+  pipeline_->addFront(std::make_shared<LayerBitrateCalculationHandler>());
```
jcague: 👍

jcague:
btw, tests failed but it seems like docker registry could have failed, can you rebuild it to check if tests are green?

erizo/src/erizo/WebRtcConnection.cpp
Equod: Coping the shared pointer basically causes reference counter increment, function move instead doesn't affect reference counter at all because it simply transfers its ownership to another object.
Please consider the following code:

```
#include <iostream>
#include <memory>

class A
{
 public:
  std::string some_usefull_string;
};

void do_stuff_with_copied_shared_ptr(std::shared_ptr<A> ptr)
{
  std::cout << "before coping use_count is " << ptr.use_count() << std::endl;
  std::shared_ptr<A> local_ptr = ptr;
  std::cout << "after coping use_count is " << ptr.use_count() << std::endl;
}

void do_stuff_with_moved_shared_ptr(std::shared_ptr<A> ptr)
{
  std::cout << "before moving use_count is " << ptr.use_count() << std::endl;
  std::shared_ptr<A> local_ptr = std::move(ptr);
  std::cout << "after moving ptr use_count is " << ptr.use_count() << std::endl;
  if(!ptr)
    std::cout << "because ptr is not valid any more" << std::endl;
  if(local_ptr)
    std::cout << "but local_ptr is valid" << std::endl;
  std::cout << "and local_ptr use_count is " << local_ptr.use_count() << std::endl;
}

int main()
{
  std::shared_ptr<A> ptr = std::make_shared<A>();
  std::cout << "after construction use_count is " << ptr.use_count() << std::endl;
  do_stuff_with_copied_shared_ptr(ptr);
  do_stuff_with_moved_shared_ptr(ptr);
  std::cout << "before leaving use_count is " << ptr.use_count() << std::endl;
  return 0;
}

```
Obviously std::move() invalidates the object passed (ptr in this case), so you should be aware of this, but you can use "local_ptr" which is always valid. And this is the best practice in small functions like setters because the reference counter is atomic and changing it requires more time then a for standard variable. It's just a little performance improvement, so fill you free to use it in the way you like. :)
For more info please take a look at [this topic](https://stackoverflow.com/questions/41871115/why-would-i-stdmove-an-stdshared-ptr
) on stackoverflow.

jcague: Yes, I know how it works, my point is whether the possibility to introduce bugs if we use an invalidated pointer when updating this code by accident worths the code optimization. We always have to consider that specially in OS projects we can receive PRs updating this code and it could be difficult to detect this kind of issues during a review. But I also see that setters are a reasonable place where we can use std::move.

Equod: I perfectly agree. So if it's OK for you, in my PRs I'll move(shared_ptr) only in one line setters.

PR #1250, opened 2018-06-28
When closing streams, also unbind from all events.

kekkokk:
**Description**

When closing a stream we are currently NOT unbinding from this:

https://github.com/lynckia/licode/blob/18c5562a312b080468f6fdc5f8a0ccec1abbcf54/erizo_controller/erizoClient/src/Room.js#L126

Steps to reproduce:
tab 1) Publisher 
tab 2) subscriber. Subscribe to remote stream, then unsubscribe, then subscribe again. 2 stream added events are fired

[] It needs and includes Unit Testsì

### DIFF
--- a/erizo_controller/erizoClient/src/Events.js
+++ b/erizo_controller/erizoClient/src/Events.js
@@ -49,6 +49,7 @@ const EventDispatcher = () => {
   that.on = that.addEventListener;
   that.off = that.removeEventListener;
   that.emit = that.dispatchEvent;
+  that.unbindEvents = () => { dispatcher.eventListeners = {}; };
 
   return that;
 };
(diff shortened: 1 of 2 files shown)


erizo_controller/erizoClient/src/Events.js
```diff
@@ -49,6 +49,7 @@ const EventDispatcher = () => {
   that.on = that.addEventListener;
   that.off = that.removeEventListener;
   that.emit = that.dispatchEvent;
+  that.unbindEvents = () => { dispatcher.eventListeners = {}; };
```
jcague: I think that a name like `removeAllListeners()` would be more similar to the rest of methods here. wdyt? 

Could you also move it to line ~35, just below `removeEventListener() {...}` block.

PR #1263, opened 2018-07-17
Fix unpublish and unsubscribe for Single PC

jcague:
**Description**

There was some issues in Single Peer Connection when trying to unsubscribe or unpublish streams because we were not negotiating the SDP again. This PR adds such negotiation and it also introduces some minor fixes.

Not needed.

### DIFF
--- a/erizo_controller/erizoClient/src/Room.js
+++ b/erizo_controller/erizoClient/src/Room.js
@@ -45,7 +45,7 @@ const Room = (altIo, altConnectionHelpers, altConnectionManager, specInput) => {
   // Private functions
   const removeStream = (streamInput) => {
     const stream = streamInput;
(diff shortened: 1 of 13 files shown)


erizo_controller/erizoClient/src/Room.js
```diff
@@ -45,7 +45,7 @@ const Room = (altIo, altConnectionHelpers, altConnectionManager, specInput) => {
   // Private functions
   const removeStream = (streamInput) => {
     const stream = streamInput;
-    if (stream.stream) {
+    if (stream.stream && !stream.local) {
```
lodoyun: Not closing the local stream when unpublishing can be risky in some cases. I'd prefer thinking about how we want to manage this in the API and update it accordingly in another PR

PR #1264, opened 2018-07-17
throw error when trying to subscribe when socket is not authenticated

kekkokk:
**Description**

During HEAVY stress test with spine, it happens sometimes that stream-added (of another stream in the room) is fired on client before the client token callback (which sets the socket.state = CONNECTED).

So the client try to subscribe to this stream but the function sendSDP return since the socket is still in != CONNECTED state and no errors are thrown (only a log.error)

### DIFF
--- a/erizo_controller/erizoClient/src/Room.js
+++ b/erizo_controller/erizoClient/src/Room.js
@@ -762,6 +762,10 @@ const Room = (altIo, altConnectionHelpers, altConnectionManager, specInput) => {
 
   // It subscribe to a remote stream and draws it inside the HTML tag given by the ID='elementID'
   that.subscribe = (streamInput, optionsInput = {}, callback = () => {}) => {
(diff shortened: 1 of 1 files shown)


mkhahani:
Regarding stress test, do you happen to have the following error when creating frequent user tokens in high rate and without delay?

`401 Error{"WWW-Authenticate":"MAuth realm=\"http://marte3.dit.upm.es\""}`

For me it randomly fails. For example 12 failures in 50 requests. But if I put a delay of 100ms between requests then it's fine.

kekkokk:
No, I use a customized version of nuve and also spine.
I rewrote Spine from scratch and when I join, I join simultaneously with all clients. No more connectionDelay.
Regarding Nuve I think I changed something in the past but I really don't remember what.
I think is that:

screen shot 2018-07-18 at 15 44 45

in NuveAuthenticator.js

PR #1271, opened 2018-07-31
Check if the handler is initialized before reading or processing

lodoyun:
**Description**
This PR prevents `BandwidthEstimationHandler ` from reading packets or processing data if it hasn't been initialized before.

### DIFF
--- a/erizo/src/erizo/rtp/BandwidthEstimationHandler.cpp
+++ b/erizo/src/erizo/rtp/BandwidthEstimationHandler.cpp
@@ -88,6 +88,9 @@ void BandwidthEstimationHandler::notifyUpdate() {
 }
 
 void BandwidthEstimationHandler::process() {
+  if (!initialized_) {
+    return;
+  }
   rbe_->Process();
   std::weak_ptr<BandwidthEstimationHandler> weak_ptr = shared_from_this();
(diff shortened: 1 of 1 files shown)


erizo/src/erizo/rtp/BandwidthEstimationHandler.cpp
```diff
@@ -143,6 +146,9 @@ void BandwidthEstimationHandler::updateExtensionMap(bool is_video, std::array<RT
 }
 
 void BandwidthEstimationHandler::read(Context *ctx, std::shared_ptr<DataPacket> packet) {
+  if (!initialized_) {
```
jcague: should we forward the packet to the next handler in this case with `ctx->fireRead(std::move(packet));`?

erizo/src/erizo/rtp/BandwidthEstimationHandler.cpp
```diff
@@ -88,6 +88,9 @@ void BandwidthEstimationHandler::notifyUpdate() {
 }
 
 void BandwidthEstimationHandler::process() {
+  if (!initialized_) {
```
jcague: we're calling process() from read() which is safe now, so I think we don't need this line.

Issue #1269, opened 2018-07-30
add vp9 support


lidedongsn:
SFU is support vp9 now，I can see. When to add vp9 support as MCU(VP9 rtpdepacketer and decoder)?Is
 there any plan? I've implemented a version myself simply.

kekkokk:
could you please share your version in a PR?

lidedongsn:
I modify licode  a lot based on 2017's. Added video and audio mixer(blend one or more videos into one stream), so my version is a lot different. I'll see how I can make a PR

jcague:
is this issue solved with your PR @lidedongsn ?

lidedongsn:
yes. I use vp9 data to decode -> raw yuv -> yuv mixer -> re-encode->and so on... ,this PR[#1274](https://github.com/lynckia/licode/pull/1274) is only to get the **right** vp9 data payload and payload len.  So I can decode vp9 corretly.

PR #1274, opened 2018-08-01
RtpVP9Parser vp9 payload len

lidedongsn:
I've post an issue [#1269](https://github.com/lynckia/licode/issues/1269) disscuss “add vp9 support”. In my tests, RtpVP9Parser.cpp vp9 payload len is not correct, it should - vp9 headers lens. This is only a small modification part in my "MIXER" version. I'm going to do what I can  to add vp9 support.


lidedongsn:
why there are two commit，the first is it. The Second is my misoperation maybe

Issue #362, opened 2015-10-19
sending parallel request produces error: 401 Unauthorized


AngelKyriako:
We are creating a virtual class application using licode and I am against an issue during implementing the unit tests for the multimedia routes based on licode.

Assuming I have written a function (createRoomFunc) that creates a licode room through the nuve client library.

The following procedure that sends each request in series will work, since it requests the server again after a request has sent its corresponding response:
```
    describe('licode API create', function () {   
      it('should create a few rooms successfully', function (done) {   
        async.series([   
          function (next) {   
            createRoomFunc('testCreate', next);   
          },   
          function (next) {    
            createRoomFunc('testCreate2', next);   
          },   
          function (next) {   
            createRoomFunc('testCreate3', next);   
          },   
          function (next) {   
            createRoomFunc('testCreate4', next);
          },
          function (next) {
            createRoomFunc('testCreate5', next);
          }
        ], done);
      });
    });
```

However if the same code snippet is used with async.parallel it will crash sending back a 401 error message.

Any feedback on this aspect ?
I will keep the issue updated as I work around it

zoopnin:
how to run the licode server? can u help me?

AngelKyriako:
doesn't that work for u ?

http://lynckia.com/licode/install.html

2015-10-21 6:20 GMT+03:00 zoopnin <notifications@github.com>:

pybean:
AFAIK nuve requires every request for the same service ID has to be sent and received in time order, in order to prevent the race condition.

If you see messages "Invalid timestamp or cnonce" from your nuve log, then that's the case.

My suggestion is to implement a FIFO queue on the server side to digest parallel requests from the client side, so the those parallel requests will be processed one by one.

AngelKyriako:
Yeah I have set it up with a FIFO redis queue for the time being.

Currently I have one seperate queue for each seperate nuve method. 

I guess I should push every nuve method in the same queue to be sure that the 401 error is not thrown. Correct ?

At first I though this error occurs only for create(room, token)  & delete room requests but it seems it is thrown also for retrieve requests. Is there a reason the for retrieve request to be limited by the time order as well ?

Finally, are there any plans to setup a FIFO queue embedded nuve to prevent this ?

Thank you in advance

lodoyun:
@pybean is right, that is a security mechanism to avoid replay attacks. Basically, you should try to send messages in order and avoiding bursts. Implementing a buffer or queue in Nuve would defeat the whole purpose of the mechanism.

[...]

PR #1276, opened 2018-08-02
Add flag to not check request timestamp

bpllana:
This update is to get around async parallel requests
producing 401 Unauthorized errors. For more info see
https://github.com/lynckia/licode/issues/362

**Description**

This update blocks this issue: https://github.com/lynckia/licode/issues/362 from occurring. The default behavior of nuve authentication is still the same. By default the request timestamps are verified. This update checks the value of an environment variable 'NUVE_AUTH_CHECK_TIMESTAMP' and if the env var is set to 'false' then the timestamp from the request is not verified.

[x] It needs and includes Unit Tests

No changes in the public API are needed. The system under which licode runs would need to set the env var as stated above for this code update to have effect.

### DIFF
--- a/nuve/nuveAPI/auth/nuveAuthenticator.js
+++ b/nuve/nuveAPI/auth/nuveAuthenticator.js
@@ -9,6 +9,10 @@ var logger = require('./../logger').logger;
 var log = logger.getLogger('NuveAuthenticator');
 
 var cache = {};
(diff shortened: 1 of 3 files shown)


jcague:
I'd fix the current implementation to cover scenarios rather than making it less secure, even if it's something we can disable by configuration.

PR #1278, opened 2018-08-09
Some small fixes to build erizoClient without errors

pooyarm:
**Description**

Fix dependencies to build erizoClient without pain and blood.
Some dependencies weren't found in package.json so I added them one by one.
Added eslint auto fixer to fix line breaks based on OS. (to avoid LF and CRLF errors in windows)


pooyarm:
this pull request is just form first two commits:
- Add erizoClient dependencies to package.json
- Use eslint autofix

why other commits which pushed after this pull request has been added here? how can I remove them?

Issue #1284, opened 2018-08-30
Firefox sending low rembs on stream with 30+fps


kekkokk:
Hi guys,
so we are struggling with a bug that's driving us nuts.

Environment: OSX chrome (any version) publishing, Firefox (any version) subscribing , NO simulcast.
The webcam is a Logitech c920, which in chrome is publishing at a rate of about 29/30/31 fps (640x480):

![screen shot 2018-08-30 at 12 12 37](https://user-images.githubusercontent.com/15072723/44845508-05d3e700-ac4e-11e8-955a-9c2632af8ea6.png)

To do this we removed from the basicexample the framerate limitation constraint [10, 20] fps.

when the publisher is alone in the server we reach full bandwidth (1000 kbits, set on licode_config.js)

![screen shot 2018-08-30 at 12 17 03](https://user-images.githubusercontent.com/15072723/44845738-a1fdee00-ac4e-11e8-96d4-aa394078eaad.png)

when a Firefox join as a subscriber, REMBS sent from firefox will drastically drop the bandwidth to ~30kbits:

![screen shot 2018-08-30 at 12 22 13](https://user-images.githubusercontent.com/15072723/44846008-64e62b80-ac4f-11e8-9b71-e1f4b6646b78.png)

if the Firefox client disconnects, everything will return back to work.

Observations:
- only reproducible in OSX chrome publisher, on ubuntu 14 client everything works fine.
- tried different computers and OS

So we tried these steps:
- Try Firefox publisher and Firefox subscriber -> works
- Try every combination of OS and chrome and Firefox version, on the same and different machines with the same webcam, (which in every os can reach 30fps).
also tried with the server being in local network or remote AWS (natively, no docker) -> no changes, same strange behaviour.

So we maybe think is a bug of chrome with 30+ fps. then we tried the same call but in p2p: f**k -> just works flawlessly

It must be something in Licode so:
- Tried to completely disable every module in the pipeline, so the server acts only as a "passthru tube" -> bandwidth continues to drops.
- Tried to also commenting the c++ lines that rewrites the abs send time, so they negotiate the abs send time as extension but the server doesn't rewrite nothing. this time the server is only un-encrypting the ingoing packets and re-encrypting the outgoing packets. this is the most "transparent" way that Licode can works ("I think"). -> bandwidth constinue to drops.
[...]

PR #1287, opened 2018-09-05
Extension translation

Equod:
Fix https://github.com/lynckia/licode/issues/1284

As wrote in the issue,
We found that the extension ids between different browsers were different and we were not translating them in the server.

This PR adds a translator for every RTP packet with extensions and strip the unknown or not negotiated extensions of the subscriber adding padding as per RFC to avoid misunderstanding between browsers.

Also there was some minor bugs parsing the extensions:
- missing padding management during parsing
- missing management of extension id = 15 ([rcf5285](https://tools.ietf.org/html/rfc5285#section-4.2))

> The local identifier value 15 is reserved for future extension and MUST NOT be used as an identifier. If the ID value 15 is encountered, its length field should be ignored, processing of the entire extension should terminate at that point, and only the extension elements present prior to the element with ID 15 considered.

We decided to not attach the mapping in the data packets since it was a fast but not clever way to do it.
We used erizoJsController to pass the extensionMap to the subscriber adding an add-on API method.

We also tested with Firefox and now if effectively switches to the ABS send time BWE.

firefox_abs_bwe


erizo/src/erizo/rtp/RtpExtensionProcessor.cpp
```diff
@@ -5,118 +5,94 @@
...
-      while (currentPlace < (totalExtLength*4)) {
-        extByte = (uint8_t)(*extBuffer);
+    char* extBuffer = reinterpret_cast<char*>(&head->extensions);
+    uint8_t extByte = 0;
+    uint16_t currentPlace = 1;
+    uint8_t extId = 0;
+    uint8_t extLength = 0;
+    while (currentPlace < (totalExtLength * 4)) {
+      extByte = (uint8_t) (*extBuffer);
+      if (extByte != RtpExtentionIdTranslator::padding) {
```
jcague: RtpExtentionIdTranslator::unknown might be more descriptive in this case, am I right? When reading the PR I thought you were adding RTP Padding in some case

erizo/src/erizo/rtp/RtpExtensionProcessor.cpp
```diff
@@ -137,8 +113,39 @@ uint32_t RtpExtensionProcessor::processAbsSendTime(char* buf) {
...
+  uint8_t src_id = data[0] >> 4;
+  uint8_t len = data[0] & (uint8_t) 0x0f;
+  auto dest_id = (uint8_t) translator.translateId(src_id);
+  if (dest_id != RtpExtentionIdTranslator::padding) {
+    if (dest_id != src_id) {
+      ELOG_DEBUG("Changing extension id from %d to %d", (int) src_id, (int) dest_id);
+    }
+    data[0] = (dest_id << 4) | len;
+  } else {
+    // clear extension
```
jcague: I guess this solution is good enough as a first step, but we should probably remove the bytes in the header to save bandwidth.

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -741,4 +741,12 @@ void WebRtcConnection::setTransport(std::shared_ptr<Transport> transport) {  //
   bundle_ = true;
 }
 
+std::pair<RTPExtensionsMap, RTPExtensionsMap> WebRtcConnection::getSourceExtensionMap() {
+  return std::make_pair(extension_processor_.getVideoExtensionMap(), extension_processor_.getAudioExtensionMap());
+}
+
+void WebRtcConnection::setSourceExtensionMap(std::shared_ptr<WebRtcConnection> source_wrtc) {
+  extension_processor_.setSourceExtensionMap(source_wrtc->getSourceExtensionMap());
```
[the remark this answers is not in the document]
Equod: Good observation!

erizo/src/erizo/rtp/RtpExtensionProcessor.cpp
Equod: Sure, RtpExtentionIdTranslator::unknown would be more clear in this case, I'll add it as static constexpr unknown = 0x00; (because extension 0 can't be a valid extension).

erizo/src/erizo/rtp/RtpExtensionProcessor.cpp
Equod: In my opinion removing at all bytes from extension will not be a good solution because a single extension is very small (2 or 4 bytes more or less). And because to remove at all an extension you should:
1) change extension length field
2) change packet header length
3) shift following extensions (if needed)
3) add padding (if needed)
4) shift payload
Please consider that to shift payload you can't just use a memcpy(), because on overlapped memory region you have to use memmove() which is slower because internally uses a temporal buffer.
It can be done but in this way we save less then 0.5% of packet length
So I would suggest to use the "padding approach" instead.

jcague: could you at least put a document with the alternative just in case we need to do it in the future please?

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -741,4 +741,14 @@ void WebRtcConnection::setTransport(std::shared_ptr<Transport> transport) {  //
   bundle_ = true;
 }
 
+std::pair<RTPExtensionsMap, RTPExtensionsMap> WebRtcConnection::getSourceExtensionMap() {
+  return std::make_pair(extension_processor_.getVideoExtensionMap(), extension_processor_.getAudioExtensionMap());
```
jcague: as I said before I think getSourceExtensionMap should be also async

Equod: You're right :+1: but I think it would be better to put just a mutex in getter, because two WebRtcConnections may have two different workers and be executed on different threads.

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -741,4 +741,22 @@ void WebRtcConnection::setTransport(std::shared_ptr<Transport> transport) {  //
...
+                               extension_processor_.getAudioExtensionMap());
+  return result;
+}
+
+void WebRtcConnection::setSourceExtensionMap(std::shared_ptr<WebRtcConnection> source_wrtc) {
+  asyncTask([source_wrtc](std::shared_ptr<WebRtcConnection> this_connection){
+    auto source_map_future = source_wrtc->worker_->post_async_task([source_wrtc]{
+      return source_wrtc->getSourceExtensionMap();
+    });
+    auto map = source_map_future.get();
```
jcague: I think we wouldn't need  the usage of the mutex or at least I can't think of a case where we access that info from different threads.
Anyway, you're blocking thread execution here. Another option would be:
```js
std::weak_ptr<WebRtcConnection> weak_this = shared_from_this();
source_wrtc->asyncTask([weak_this](std::shared_ptr<WebRtcConnection> source){
    map = source->getSourceExtensionMap();
    if (auto this_ptr = weak_this.lock()) {
        this_ptr->asyncTask([map](std::shared_ptr<WebRtcConnection> this_connection) {
            this_connection->setSourceExtensionMap(map);
        });
    }
});
```
It's just pseudo code so it will have errors, but it's just to illustrate the idea. By doing this we would get rid of the new post_async_task and the `mutex` (I always really try to remove them from this context).

Equod: Yep, doing "auto map = source_map_future.get()" waits for the result from another thread, so I think your option works better :smile: . And by making getSourceExtensionMap() private, we don't need extension_map_mutex any more.

erizo/src/erizo/LibNiceConnection.cpp
```diff
@@ -26,11 +28,10 @@ DEFINE_LOGGER(LibNiceConnection, "LibNiceConnection")
 
 void cb_nice_recv(NiceAgent* agent, guint stream_id, guint component_id,
     guint len, gchar* buf, gpointer user_data) {
-  if (user_data == NULL || len == 0) {
-    return;
+  if (user_data && len != 0) {
```
jcague: Please move unrelated changes to another PR

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -741,4 +741,20 @@ void WebRtcConnection::setTransport(std::shared_ptr<Transport> transport) {  //
...
+                               extension_processor_.getAudioExtensionMap());
+  return result;
+}
+
+void WebRtcConnection::setSourceExtensionMap(std::shared_ptr<WebRtcConnection> source_wrtc) {
+  std::weak_ptr<WebRtcConnection> this_weak = shared_from_this();
+  source_wrtc->asyncTask([this_weak](std::shared_ptr<WebRtcConnection> source_connection) {
+    auto map = source_connection->getSourceExtensionMap();
+    if (auto this_connection = this_weak.lock()) {
+      this_connection->extension_processor_.setSourceExtensionMap(map);
```
jcague: I think we should call this_connection->asyncTask() here to make it run in the right worker.

PR #1292, opened 2018-09-11
[POC] Add support for ssrc multiplexing (RTX)

kekkokk:
This PR aims to make Licode works when the publisher negotiates RTX with FID fields.

This was not useful until now because all the supported browsers, when we answered with no rtx fmtp, they did the retransmission only in the main ssrc when requesting nack.

But Edge, as you can see in my answer from fippo, only support RTX in separate rtx ssrc channel, so, we implemented a module that translates the rtx packets as it was with a single ssrc, so for Licode it's transparent.
This is a first step to better support other browsers in the future

Note that now also chrome and firefox will send rtx in separate channel.

It just works but I noticed that in SdpInfo.cpp there was already present some logic for rtx but it doesn't work (probably due to the parsing method of ConnectionDescription or some other javascript code).

So I made this PR as a POC because I need to know what do you think @jcague @lodoyun and if we can better adjust the javascript code or if you like what I've done.

Currently no test are provided.

### DIFF
--- a/erizo/src/erizo/MediaStream.cpp
+++ b/erizo/src/erizo/MediaStream.cpp
@@ -365,7 +365,7 @@ void MediaStream::initializePipeline() {
   pipeline_->addFront(std::make_shared<LayerDetectorHandler>());
   pipeline_->addFront(std::make_shared<OutgoingStatsHandler>());
   pipeline_->addFront(std::make_shared<PacketCodecParser>());
-
+  pipeline_->addFront(std::make_shared<RtxPacketTranslator>());
   pipeline_->addFront(std::make_shared<PacketWriter>(this));
   pipeline_->finalize();
   pipeline_initialized_ = true;
--- a/erizo/src/erizo/MediaStream.h
+++ b/erizo/src/erizo/MediaStream.h
@@ -24,6 +24,7 @@
 #include "pipeline/Service.h"
 #include "rtp/QualityManager.h"
 #include "rtp/PacketBufferService.h"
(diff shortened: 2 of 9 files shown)


kekkokk:
The main problem I see here is that rtx are not added as codec, but instead they are added as an extension of a specific codec by this function:
https://github.com/lynckia/licode/blob/79c382c3a9867af7059a17185c6b27b21d78e04d/erizo_controller/common/semanticSdp/SDPInfo.js#L521

```
CodecInfo {
  codec: 'VP8',
  type: 96,
  rate: 90000,
  encoding: undefined,
  params: {},
  feedback: 
   [ { type: 'goog-remb', subtype: undefined },
     { type: 'transport-cc', subtype: undefined },
     { type: 'ccm', subtype: 'fir' },
     { type: 'nack', subtype: undefined },
     { type: 'nack', subtype: 'pli' } ],
  rtx: 97 }
```

This means that https://github.com/lynckia/licode/blob/79c382c3a9867af7059a17185c6b27b21d78e04d/erizo/src/erizo/SdpInfo.cpp#L960

will never be fired since 
`a=rtpmap:99 rtx/90000
`
will never reach SdpInfo.cpp as a proper codec.

This will "break" the rtp_media_config utility, since I can add the rtx in the default mediaconfig:
`  default: { rtpMappings: { vp8, opus, rtx }, extMappings },
`
but with no effect.

So our approach is to set always on this rtx module, otherwise it needs some SdpInfo.js/cpp refactor

I'm not sure if I explained myself very well

PR #1296, opened 2018-09-13
Add a new algorithm to allocate better bandwidth in Single PC

jcague:
**Description**

The way we were distributing available bandwidth among streams in the same single PC was suboptimal for some cases. This PR aims to improve the algorithm when we use Simulcast and Slideshow.

Not needed.


erizo/src/erizo/bandwidth/TargetVideoBWDistributor.cpp
```diff
@@ -0,0 +1,56 @@
...
+  std::for_each(streams.begin(), streams.end(),
+    [&remaining_bitrate, &remaining_streams, transport, ssrc, this](const std::shared_ptr<MediaStream> &stream) {
+      uint32_t max_bitrate = stream->getMaxVideoBW();
+
+      uint32_t target_bitrate = getTargetVideoBW(stream);
+
+      uint32_t remaining_avg_bitrate = remaining_bitrate / remaining_streams;
+      uint32_t bitrate = std::min(target_bitrate, remaining_avg_bitrate);
+      uint32_t remb = std::min(max_bitrate, remaining_avg_bitrate);
+      uint32_t bitrate_to_transfer = std::max(bitrate - target_bitrate * 1.2, 0.0);
```
lodoyun: I don't think we need `bitrate_to_transfer` it will always be 0 unless we have a bug somewhere else. As far as I can tell `bitrate` is enough to update `remaining_bitrate` and distribute the bitrate correctly.

erizo/src/erizo/bandwidth/MaxVideoBWDistributor.cpp
```diff
@@ -0,0 +1,35 @@
+/*
+ * MaxVideoBWDistributor.cpp
```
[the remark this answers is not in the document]
jcague: I completely agree, I will remove it once we're 100% sure the new code works for all cases.

erizo/src/erizo/bandwidth/TargetVideoBWDistributor.cpp
jcague: good catch!

PR #1297, opened 2018-09-18
API to enable slideshow keyframes below a spatial layer

lodoyun:
**Description**

This PR updates #1199, adding the possibility to enable SlideShow when below layer 0, effectively replacing Fallback mode if desired.
This is useful, for example, when activating simulcast with Screen sharing. Instead of getting a frozen frame, users with low bandwidth will get an updated image (of the lowest possible quality) every few seconds.

Additionally, this PR updates the API to better reflect this behaviour and cleans up some of the variable names.

### DIFF
--- a/erizo/src/erizo/rtp/QualityManager.cpp
+++ b/erizo/src/erizo/rtp/QualityManager.cpp
@@ -168,7 +168,7 @@ void QualityManager::selectLayer(bool try_higher_layers) {
     aux_spatial_layer++;
   }
 
-  ELOG_DEBUG("below_min_layer %u, freeze_fallback_active_: %u", below_min_layer, freeze_fallback_active_);
+  ELOG_DEBUG("message: below_min_layer %u, freeze_fallback_active_: %u", below_min_layer, freeze_fallback_active_);
   if (below_min_layer != freeze_fallback_active_) {
     if (below_min_layer || try_higher_layers) {
       freeze_fallback_active_ = below_min_layer;
--- a/erizo/src/erizo/rtp/QualityManager.cpp
+++ b/erizo/src/erizo/rtp/QualityManager.cpp
@@ -15,9 +15,9 @@ constexpr float QualityManager::kIncreaseLayerBitrateThreshold;
 
 QualityManager::QualityManager(std::shared_ptr<Clock> the_clock)
   : initialized_{false}, enabled_{false}, padding_enabled_{false}, forced_layers_{false},
-  slideshow_fallback_active_{false}, spatial_layer_{0},
+  freeze_fallback_active_{false}, spatial_layer_{0},
   temporal_layer_{0}, max_active_spatial_layer_{0},
-  max_active_temporal_layer_{0}, min_desired_spatial_layer_{-1}, max_video_width_{-1},
+  max_active_temporal_layer_{0}, slideshow_below_spatial_layer_{-1}, max_video_width_{-1},
   max_video_height_{-1}, max_video_frame_rate_{-1}, current_estimated_bitrate_{0},
   last_quality_check_{the_clock->now()}, last_activity_check_{the_clock->now()}, clock_{the_clock} {}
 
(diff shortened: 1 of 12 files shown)


erizo/src/erizo/rtp/QualityManager.cpp
```diff
@@ -166,24 +167,27 @@ void QualityManager::selectLayer(bool try_higher_layers) {
...
-  if (below_min_layer != slideshow_fallback_active_) {
+  ELOG_DEBUG("below_min_layer %u, freeze_fallback_active_: %u", below_min_layer, freeze_fallback_active_);
+  if (below_min_layer != freeze_fallback_active_) {
     if (below_min_layer || try_higher_layers) {
-      slideshow_fallback_active_ = below_min_layer;
+      freeze_fallback_active_ = below_min_layer;
       ELOG_DEBUG("message: Setting slideshow fallback, below_min_layer %u, spatial_layer %d,"
-          "next_spatial_layer %d slidehow_fallback_active_: %d",
-          below_min_layer, spatial_layer_, next_spatial_layer, slideshow_fallback_active_);
+          "next_spatial_layer %d slidehow_fallback_active_: %d, min_desired_spatial_layer: %d",
```
jcague: shouldn't this be called  `min_requested_spatial` like you do above?

erizo/src/erizo/rtp/QualityManager.cpp
```diff
@@ -166,24 +167,27 @@ void QualityManager::selectLayer(bool try_higher_layers) {
...
+          below_min_layer, spatial_layer_, next_spatial_layer, freeze_fallback_active_, slideshow_below_spatial_layer_);
```
jcague: same  as above should we use  `min_requested_spatial_layer` here?

erizo/src/erizo/rtp/QualityManager.cpp
```diff
@@ -259,9 +262,10 @@ void QualityManager::forceLayers(int spatial_layer, int temporal_layer) {
   temporal_layer_ = temporal_layer;
 }
 
-void QualityManager::setMinDesiredSpatialLayer(int spatial_layer) {
-  ELOG_DEBUG("message: setting min desired spatial layer, spatial_layer: %d", spatial_layer);
-  min_desired_spatial_layer_ = spatial_layer;
+void QualityManager::enableSlideShowBelowSpatialLayer(int spatial_layer) {
+  ELOG_DEBUG("message: Will activate slideshow when below spatial layer, spatial_layer: %d", spatial_layer);
+  slideshow_below_spatial_layer_ = spatial_layer;
+  freeze_fallback_active_ = false;
```
jcague: is there a way to disable this feature afterwards?

erizo/src/erizo/rtp/QualityManager.cpp
lodoyun: Right, actually I think I want both `min_requested_spatial_layer` and `slideshow_below_spatial_layer_` to debug possible problems

erizo/src/erizo/rtp/QualityManager.cpp
lodoyun: It is possible but it's a little bit awkward, you would have to do `enableSlideShowBelowSpatialLayer(-1)` a possible solution would be to have a `disableSlideShowBelowSpatialLayer` either in the client side and have it be a call to this function and add -1 as a parameter or create the whole call stack from the client to a new `QualityManage::disableSlideShowBelowSpatialLayer()`.
Option 1 adds less noise but Option 2 is clearer. What do you think?
Besides, there's a bug currently that may prevent this from working but I'm on it and should be fixed in whatever option we choose.

lodoyun:
I've changed the API to `enableSlideShowBelowSpatialLayer (bool enabled, int spatialLayer)`.
You can now disable the functionality with `enableSlideShowBelowSpatialLayer (false)`, the spatialLayer will be ignored in that case.
`enableSlideShowBelowSpatialLayer(true)` will default to spatial Layer 0.

erizo/src/erizo/rtp/QualityManager.cpp
lodoyun: Addressed this in the new commit 👍

erizo/src/erizo/rtp/QualityManager.cpp
```diff
@@ -159,31 +161,36 @@ void QualityManager::selectLayer(bool try_higher_layers) {
...
-      ELOG_DEBUG("message: Skipping below min spatial layer, aux_layer: %d, min_valid_spatial_layer",
+      ELOG_DEBUG("message: Skipping below min spatial layer, aux_layer: %d, min_valid_spatial_layer: %d",
           aux_spatial_layer, min_valid_spatial_layer);
     }
     aux_temporal_layer = 0;
     aux_spatial_layer++;
   }
 
-  if (below_min_layer != slideshow_fallback_active_) {
+  ELOG_DEBUG("below_min_layer %u, freeze_fallback_active_: %u", below_min_layer, freeze_fallback_active_);
```
jcague: nick: I know we don't follow a strict patter for DEBUG messages, but all the other messages in this class start with `message:`

kekkokk:
would be very nice to also have an API to force the minimum temporal layer, since, especially on lower spatial layer, since in terms of kbytes the difference is really small.

I think that in the user experience, one may prefer to see a fluid smaller spatial layer instead of one laggy (ex. spatial 1 and temporal 0) but with better resolution.

So the quality manager will switch only on preferred ones

PR #1300, opened 2018-09-20
Update to Ubuntu 16.04

lodoyun:
**Description**

This PR fixes a compatibility issue with Ubuntu 16.04 and sets it as the default for the docker builds.
Once this PR is merged, all our tests will be run with the new docker image so Ubuntu 16.04 will be our supported Ubuntu version - this PR also updates the docs accordingly

 [X] It includes documentation for these changes in `/doc`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
 MAINTAINER Lynckia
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
```
jcague: 💯

PR #1307, opened 2018-09-28
Fix SDP negotiation when a stream is audio only

jcague:
**Description**

There was an issue in Single Peer Connection by which we were not negotiating SDP properly when the first subscribed stream had only audio.

Not needed.

### DIFF
--- a/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
+++ b/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
@@ -43,8 +43,8 @@ const BaseStack = (specInput) => {
   specBase.remoteDescriptionSet = false;
 
   that.mediaConstraints = {
-    offerToReceiveVideo: true,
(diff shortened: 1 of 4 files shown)


erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
```diff
@@ -43,8 +43,8 @@ const BaseStack = (specInput) => {
   specBase.remoteDescriptionSet = false;
 
   that.mediaConstraints = {
-    offerToReceiveVideo: (that.video !== undefined && that.video !== false),
-    offerToReceiveAudio: (that.audio !== undefined && that.audio !== false),
+    offerToReceiveVideo: true,
```
lodoyun: This was problematic at some point with firefox. We should try it and move it to Chrome's stack if that's the case.

jcague: good catch!

PR #1314, opened 2018-10-03
Burst size is now calculated depending on the packets to be sent

lodoyun:
**Description**
This PR makes the burst size for the Token Bucket used in `PaddingGeneratorHandler` variable.
Before this PR, the burst size was a constant but we are trying to send a variable amount of packets depending on the target padding bitrate. It makes sense to also make variable the burst size that the tokenBucket will allow.

The maximum and minimum burst sizes are bounded to avoid bursts that are too big and, at the same time, avoid setting a burst limit too small.

### DIFF
--- a/erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp
+++ b/erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp
@@ -15,7 +15,8 @@ constexpr duration kStatsPeriod = std::chrono::milliseconds(100);
 constexpr uint8_t kMaxPaddingSize = 255;
 constexpr uint64_t kMinMarkerRate = 3;
 constexpr uint64_t kInitialBitrate = 300000;
-constexpr uint64_t kPaddingBurstSize = 255 * 10;
+constexpr uint8_t kMaxBurstPackets = 200;
+constexpr uint8_t kSlideShowBurstPackets = 20;
 
 RtpPaddingGeneratorHandler::RtpPaddingGeneratorHandler(std::shared_ptr<erizo::Clock> the_clock) :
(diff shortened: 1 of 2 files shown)


kekkokk:
Just a tip: consider cloning the actual header of the packet instead of generating new one, so you don't lose the abs_send_time extension. 

https://github.com/lynckia/licode/blob/5eb96dd077e8419f3602ef69f07838724938b97a/erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp#L103

otherwise your padding may be ignored on the subscriber side once you integrate the extensionTranslator

jcague:
@kekkokk I might be getting it wrong, but I think that's what we're doing currently in this code.

kekkokk:
Yeah the problem is that you are copying the packet header with a pointer to the packet. but first you send the packet in the firewrite:

https://github.com/lynckia/licode/blob/5eb96dd077e8419f3602ef69f07838724938b97a/erizo/src/erizo/rtp/RtpPaddingGeneratorHandler.cpp#L85

then you copy the header. what happens is that the extension translator have already translated the abs_send_time id at the end of the pipeline, so if you point at this packet to generate a new header, you will have an already translated ext.id, so when you send the burst, the translator will check the extid mapping, will not find this id since it has already been translated and ends up removing the extension so you will lose the abs send time in the padding packet. copying the header BEFORE sending the packet with marker set in the firewrite will avoid this issue.

The same approach should be used in the rtpbuffer for retransmission.

lodoyun:
@kekkokk @jcague Unless I'm mistaken, this would create issues with the extension translator that will be PR'd. Currently, I don't think it creates any problems, even if the header is copied later, the extensionProcessor will set the abs_send_time properly. In any case, this does not belong to this PR since it's not related to the functionality introduced here.
I'll go ahead and merge it and we can address this issue in the extension Id translator PR.

PR #1315, opened 2018-10-06
fix subscribing to you own streams

cracker0dks:
was changed to ``localStreams.has(arg.id),``  a few commits ago but breaks subscribing to own streams (You can check it in the basic example.


jcague:
This solution breaks the case fixed by the other commit, so I'd rather do a different solution. what's the use case for subscribing to your own streams?

cracker0dks:
first of all the basicExample
I also always show the own remostream of the webcam instead of the local so if you see yourself you know its all working (licode, internet...)

PR #1333, opened 2018-11-13
Use nvm to launch erizojs from agent

kekkokk:
Currently erizo.js is spawned by launch.sh
this exec node, which is the globally installed node and not the one sourced with nvm.

this can cause some conflicts on a machine running node >8 since the APIs are compiled with gyp of node 6 but spawned with node 8.

### DIFF
--- a/erizo_controller/erizoAgent/launch.sh
+++ b/erizo_controller/erizoAgent/launch.sh
@@ -1,3 +1,10 @@
 #!/usr/bin/env bash
+
+SCRIPT=`pwd`
+PATHNAME=`dirname $SCRIPT`
+ROOT=$PATHNAME/..
+SCRIPTS_FOLDER=$ROOT/scripts
+NVM_CHECK="$SCRIPTS_FOLDER"/checkNvm.sh
+
 ulimit -c unlimited
-exec node $ERIZOJS_NODE_OPTIONS $*
+exec $NVM_CHECK node $ERIZOJS_NODE_OPTIONS $*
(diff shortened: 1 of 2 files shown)


erizo_controller/erizoAgent/launch.sh
```diff
@@ -1,3 +1,10 @@
...
+
+SCRIPT=`pwd`
+PATHNAME=`dirname $SCRIPT`
+ROOT=$PATHNAME/..
+SCRIPTS_FOLDER=$ROOT/scripts
+NVM_CHECK="$SCRIPTS_FOLDER"/checkNvm.sh
+
 ulimit -c unlimited
-exec node $ERIZOJS_NODE_OPTIONS $*
+exec $NVM_CHECK node $ERIZOJS_NODE_OPTIONS $*
```
jcague: why don't you apply the same pattern as used in other initXXX.sh scripts? it would be something like:
```
...
NVM_CHECK="$SCRIPTS_FOLDER"/checkNvm.sh

. $NVM_CHECK
ulimit -c unlimited
nvm use
exec node $ERIZOJS_NODE_OPTIONS $*
```

kekkokk: unfortunately checkNvm.sh has hardcoded a path:

`  LINKED_DIR=$CHECK_SCRIPT_PATH/../build/libdeps/nvm`

this works for every file that is calling the script in a subfolder of the main Licode folder, but not in the case of the agent which is 2 subfolder under the root project. 

So I'll propose a change with another commit to make it work wherever it's called from.
note: I had to use "--delete-prefix --silent" after each nvm use to avoid nvm printing lots of warning when executing:

```
Found '/Users/duri/git/Licode_v4/.nvmrc' with version <6.14.4>
nvm is not compatible with the npm config "prefix" option: currently set to "/Users/duri/git/Licode_v4/build/libdeps/nvm/versions/node/v6.14.4"
Run `npm config delete prefix` or `nvm use --delete-prefix v6.14.4` to unset it.
```

PR #1338, opened 2018-11-28
Fix bad arithmetic expression with overflow

jcague:
**Description**

There are spurious crashes in production due to this arithmetic expression. It seems like the issue is caused by an overflow when multiplying a uint64_t by 8.

Not needed.


jcague:
This issue was finally fixed by #1359

PR #1344, opened 2018-12-10
Update tokensResource.js: Fix bug: Can't set headers after they are sent

hocmai-longtt:
Fix bug: Can't set headers after they are sent

**Description**

### DIFF
--- a/nuve/nuveAPI/resource/tokensResource.js
+++ b/nuve/nuveAPI/resource/tokensResource.js
@@ -164,7 +164,9 @@ exports.create = (req, res) => {
       }
       log.info(`message: createToken success, roomId: ${currentRoom._id}, ` +
         `serviceId: ${currentService._id}`);
-      res.send(tokenS);
+      if (!res.headersSent) {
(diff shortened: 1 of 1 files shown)


nuve/nuveAPI/resource/tokensResource.js
```diff
@@ -164,7 +164,9 @@ exports.create = (req, res) => {
       }
       log.info(`message: createToken success, roomId: ${currentRoom._id}, ` +
         `serviceId: ${currentService._id}`);
-      res.send(tokenS);
+      if (!res.headersSent) {
```
jcague: why do we need to do this check here? by looking at the code I didn't find any case where we could have already sent the response.

darkterrorooo: i think resean is some callback call twice, see #1435

darkterrorooo: it is easy to reappear when stress testing. i had fixed it

PR #1354, opened 2019-01-29
[Feature Request] Automatic Subscriptions

jcague:
**Description**

This PR adds a feature request to be discussed and that could be used during its development. The idea is to have a mechanism to subscribe automatically to newly published streams. It could eventually help reducing subscription times to a minimum while reducing complexity in cases with many streams.

[] It needs and includes Unit Tests

**Changes in Client or Server public APIs**

[] It includes documentation for these changes in `/doc`.

PR #1361, opened 2019-02-15
Add auto subscriptions API

jcague:
**Description**

This is a work in progress project. The API is being discussed [here](https://github.com/lynckia/licode/pull/1354)

There will be changes. TBD.

- [ ] It includes documentation for these changes in `/doc`.


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -98,41 +98,63 @@ void WebRtcConnection::close() {
...
+  }
 
-  ELOG_DEBUG("%s message: Creating sdp offer, isBundle: %d", toLog(), bundle_);
+  ELOG_DEBUG("%s message: Creating sdp offer, isBundle: %d, setup: %d",
+    toLog(), bundle_, local_sdp_->internal_dtls_role);
 
-  if (video_enabled_) {
-    forEachMediaStream([this] (const std::shared_ptr<MediaStream> &media_stream) {
+  forEachMediaStream([this] (const std::shared_ptr<MediaStream> &media_stream) {
+    if (!media_stream->isReady() || media_stream->isPublisher()) {
```
lodoyun: I'm not following exactly what we're trying to accomplish with `ready_ ` in `MediaStream`.
It looks like it is always initialized with `forced=true` so it's always ready right now, is this WIP?. Is the idea that there will be mediaStreams that we have to be sure a remote Sdp has been passed and processed? In that case I would change the name of the `force` parameter to something like `doNotWaitForRemoteSdp` or something like that.

Also, what is the effect of the MediaStream not being ready here? It is going to be ignored in the SDP, are we going to handle that differently in the future?

erizo/src/erizo/NicerConnection.cpp
```diff
@@ -162,6 +162,9 @@ void NicerConnection::async(function<void(std::shared_ptr<NicerConnection>)> f)
 }
 
 void NicerConnection::start() {
+  ufrag_ = getNewUfrag();
```
lodoyun: 👍 I assume this is because you found it was possible to request the frag and pass before they were generated. Good catch.

erizo/src/erizo/WebRtcConnection.cpp
jcague: yes, you're tight, the idea is that we won't add streams that are not ready to the SDP. Previously, we added them if they were initialized. They can be ready if we receive a remote offer or if we sent a local offer, so it shouldn't change the way it used to worked for previous cases.
I'll change the parameter name to `doNotWaitForRemoteSdp` as you propose because it's more descriptive.

erizo/src/erizo/NicerConnection.cpp
jcague: yep, that happens in the new negotiations, when Erizo sends the first offer.

erizo/src/erizo/MediaStream.cpp
```diff
@@ -155,28 +160,48 @@ bool MediaStream::isSinkSSRC(uint32_t ssrc) {
...
+
+  if (isPublisher() && !ready_) {
+    bool stream_found = false;
+
+    if (video_ssrc_list_it != remote_sdp->video_ssrc_map.end() ||
+        audio_ssrc_it != remote_sdp->audio_ssrc_map.end()) {
+      stream_found = true;
+    }
+
+    if (!stream_found && isPublisher()) {
```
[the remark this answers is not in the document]
jcague: good catch!

erizo_controller/erizoController/roomController.js
```diff
@@ -360,6 +368,83 @@ exports.RoomController = (spec) => {
...
+
+    streamIds = streamIds.filter(streamId =>
+      publishers[streamId] !== undefined &&
+      subscribers[streamId].indexOf(clientId) === -1);
+
+    if (streamIds.length === 0) {
+      return;
+    }
+
+    const erizoIds = [...new Set(streamIds.map(streamId => getErizoQueue(streamId)))];
```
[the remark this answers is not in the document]
jcague: hmmm.... I like that solution too, will change it

chaitanyaprem:
I am running into the following error while doing fresh install in ubuntu-14.04 LTS desktop. My installUbuntuDeps.sh and installNuve.sh is successful without any errors. installErizo.sh is failing with the error.
[erizoinstall.txt](https://github.com/lynckia/licode/files/2967528/erizoinstall.txt)

chaitanyaprem:
> I am running into the following error while doing fresh install in ubuntu-14.04 LTS desktop. My installUbuntuDeps.sh and installNuve.sh is successful without any errors. installErizo.sh is failing with the error.
> [erizoinstall.txt](https://github.com/lynckia/licode/files/2967528/erizoinstall.txt)

This error got fixed after I installed boost version 1.65.0. Looks like install script needs to be updated to handle this.

PR #1369, opened 2019-02-25
Periodic task that checks uptime in erizoJS

lodoyun:
**Description**

We've detected that there are times where ErizoJS processes won't stop when all publishers leave, causing problems when those processes stay up a long time processing many connections.
We're investigating the issue and will fix the root of the problem. 

This PR adds a protection for this problem by checking erizoJS processes that remain up after a long time since the first time they were active. It also makes sure that no new publisher or subscriber has been added recently. These parameters can be adjusted in `licode_config`.


kekkokk:
I was investigating this a very long time ago and found the issue on this mechanism:
https://github.com/lynckia/licode/pull/1241

It's a very old PR but maybe it can help you.

PR #1371, opened 2019-02-26
Fix negotiations by better protecting access to PeerConnection

jcague:
**Description**

Now that we started to unsubscribe and subscribe to streams in production we found some issues when using Single Peer Connection. This PR aims to fix several of this pending issues in different commits.

Not needed.

### DIFF
--- a/erizo_controller/erizoJS/models/Connection.js
+++ b/erizo_controller/erizoJS/models/Connection.js
@@ -20,6 +20,8 @@ const CONN_SDP_PROCESSED = 203;
 const CONN_FAILED = 500;
 const WARN_BAD_CONNECTION = 502;
 
+const RESEND_LAST_ANSWER_RETRY_TIMEOUT = 50;
+
 class Connection extends events.EventEmitter {
   constructor(id, threadPool, ioThreadPool, options = {}) {
     super();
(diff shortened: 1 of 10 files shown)


erizo_controller/erizoJS/models/Publisher.js
```diff
@@ -453,6 +453,7 @@ class Publisher extends Source {
       clearInterval(this.mediaStream.periodicPlis);
       this.mediaStream.periodicPlis = undefined;
     }
+    return this.connection.removeMediaStream(this.mediaStream.id);
```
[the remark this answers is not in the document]
jcague: an alternative is to store the promise in a variable in line 445 and return it later, but it's up to you.

erizo_controller/erizoJS/models/Connection.js
```diff
@@ -120,19 +122,34 @@ class Connection extends events.EventEmitter {
     this.emit('status_event', info, evt, streamId);
   }
 
+  _retryWithPromise(fn, timeout) {
```
jcague: it might be better to have this function in a helper/util file

jcague: :100:

Issue #1364, opened 2019-02-19
Video problem in chrome v 72


ashar02:
Briefly describe the problem you are having in a few paragraphs.
After release of chrome 72 video is appearing blurry and pixelated.

**Steps to reproduce the issue:**
I joined the room on https://chotis2.dit.upm.es with two participants. After 15 - 30 seconds, remote video becomes blurry and pixelated. Even I installed my own licode server from code. It happens with latest code as well.

**Describe the results you received:**
On investigation using webrtc-internals. I come to know that there is big packet loss which is causing this issue on publisher end. 
<img width="1673" alt="screen shot 2019-02-19 at 4 14 18 pm" src="https://user-images.githubusercontent.com/1763187/53011367-06617380-3462-11e9-9361-94e052bebc62.png">

**Describe the results you expected:**
As I have sufficient bandwidth available, it should not face such number of packet losses.

**Additional information you deem important (e.g. issue happens only occasionally):**
I some how manage to install chrome 68 - 70 and on same condition there is no packet loss and thus video bitrate and quality are good on these.

**Licode commit/release where the issue is happening**
I used 'try room' on licode site. As well as used commit d92f97308b819de38dd7fefbbf939820c7d675d0 for my own installation. 

**Additional environment details (Local, AWS, Docker, etc.):**

ashar02:
@jcague can you please see this issue. This is coming every time with just two participants.

jcague:
Does it happen to you when there is only 1 participant?

Anyway, I tried to reproduce this issue with Chrome v72 and chotis2 but everything worked right for me. I'm using MacOS and tried with 1 and 2 participants, so it's going to be difficult to know the reason for such packet losses. What is your RTT (you can see it in another graph inside webrtc internals).

![image](https://user-images.githubusercontent.com/557698/53013047-0737dc80-3445-11e9-956d-35bf70a6e352.png)

ashar02:
<img width="1669" alt="screen shot 2019-02-19 at 5 04 59 pm" src="https://user-images.githubusercontent.com/1763187/53013862-8e4a7c00-3468-11e9-9b9e-07a328428c2d.png">
Here is the full graph. Upon several tests, I came to know that some times it comes in 30 seconds and sometimes it comes in 1 to mins. You can see whole graphs.

ashar02:
[...]

PR #1374, opened 2019-03-04
[POC] Do not send varying packet losts in RR

jcague:
**Description**

Chrome 72+ has changed the bandwidth estimation behaviour and it's now getting affected in situations where we forward RRs with different packet losts reported. This PR is a PoC that might become in a final solution if we don't find a better one.

The issue is well explained at #1364.

Not needed.


jcague:
this has been already merged in a better PR.

PR #1375, opened 2019-03-05
Remove socket listeners on channel disconnect

lodoyun:
**Description**
We've detected some instances where `Client` would process calls even after `Channel` has disconnected. This PR makes sure `Client` does not receive any events from `socket` when `disconnect` is called or when the `disconnect` event is fired.

### DIFF
--- a/erizo_controller/erizoController/models/Client.js
+++ b/erizo_controller/erizoController/models/Client.js
@@ -26,6 +26,20 @@ function listenToSocketEvents(client) {
   client.channel.on('disconnect', client.onDisconnect.bind(client));
 }
 
+function stopListeningToSocketEvents(client) {
+  log.info('Stop listening to socket events');
+  client.channel.socketRemoveListener('sendDataStream', client.onSendDataStream.bind(client));
+  client.channel.socketRemoveListener('signaling_message', client.onSignalingMessage.bind(client));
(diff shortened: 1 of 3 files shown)


erizo_controller/erizoController/models/Client.js
```diff
@@ -26,6 +26,20 @@ function listenToSocketEvents(client) {
   client.channel.on('disconnect', client.onDisconnect.bind(client));
 }
 
+function stopListeningToSocketEvents(client) {
+  log.info('Stop listening to socket events');
+  client.channel.socketRemoveListener('sendDataStream', client.onSendDataStream.bind(client));
```
paul-maxime: I think the `removeListener` call will not remove anything here.
`bind()` returns a new function every time it is called, so this function will be different from the one that was registered.

paul-maxime: You have to store the generated function somewhere if you want to remove it later:
```js
const handler = instance.onEvent.bind(instance);
on('some event', handler);
removeListener('some event', handler);
```

paul-maxime: Note that we might also just call `removeAllListeners` here instead.

lodoyun: Good catch, thats not good at all. I will work on a fix later today.
Regarding `removeAllListeners` I prefer explicitly removing the listeners that we set up in `listenToSocketEvents`.

PR #1386, opened 2019-03-22
change DTLSv1_2_method to DTLS_method

lidedongsn:
refer #1380 
 `mContext = SSL_CTX_new(DTLSv1_2_method());`
Chrome under M74 （mine 版本 73.0.3683.75（正式版本） （64 位））got “message: Handshake failed, transportName:video, openSSLerror: error:1409210A:SSL routines:ssl3_get_server_hello:wrong ssl version”
ssl/ssl.h
const SSL_METHOD *DTLS_method(void); /* DTLS 1.0 and 1.2 */
should be ok

### DIFF
--- a/erizo/src/erizo/dtls/DtlsClient.cpp
+++ b/erizo/src/erizo/dtls/DtlsClient.cpp
@@ -217,7 +217,7 @@ int createCert(const std::string& pAor, int expireDays, int keyLen, X509*& outCe
 
     ELOG_DEBUG("Creating Dtls factory, Openssl v %s", OPENSSL_VERSION_TEXT);
 
-    mContext = SSL_CTX_new(DTLSv1_2_method());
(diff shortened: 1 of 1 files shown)


jcague:
I don't see that issue with chrome  73.0.3683.86. We actually need this change to keep compatibility with newer versions.

alokp:
@jcague this PR might be a better solution because per [the docs](https://www.openssl.org/docs/man1.1.0/man3/DTLS_method.html) DTLS_method() supports both DTLS v 1.0 and v1.2 but DTLSv1_2_method() will not work with the DTLS v1.0 clients.

jcague:
@lidedongsn sorry I was wrong with this code. I didn't realize there's a generic method to use both versions. Will try it locally to check it in my local env with newer versions of chrome. Thanks for the contribution!

Issue #1393, opened 2019-04-04
std::array<RTPExtensions, 10>  for new chrome not enough


lidedongsn:
chrome: 版本 75.0.3756.0（正式版本）canary （64 位）
```
a=extmap:14 urn:ietf:params:rtp-hdrext:toffset
a=extmap:13 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:12 urn:3gpp:video-orientation
a=extmap:2 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:11 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:6 http://www.webrtc.org/experiments/rtp-hdrext/video-content-type
a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:8 http://tools.ietf.org/html/draft-ietf-avtext-framemarking-07
a=extmap:9 http://www.webrtc.org/experiments/rtp-hdrext/color-space
a=extmap:3 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:4 urn:ietf:params:rtp-hdrext:sdes:rtp-stream-id
a=extmap:5 urn:ietf:params:rtp-hdrext:sdes:repaired-rtp-stream-id
```
exmap id is out of "10" 

in files
```
        modified:   BandwidthEstimationHandler.cpp
        modified:   BandwidthEstimationHandler.h
        modified:   RtpExtensionProcessor.cpp
        modified:   RtpExtensionProcessor.h
```
for all std::array<RTPExtensions, 10>  to std::array<RTPExtensions, 20>?

Issue #1399, opened 2019-04-16
[fatal] WARNNING!!! Licode not work! (chrome between before  and after M74 )


lidedongsn:
try comment file rtp_media_config.js
```
//"urn:3gpp:video-orientation",
```
publish stream in chrome 版本 73.0.3683.103（正式版本） （64 位）
subscribe in chrome 版本 75.0.3766.2（正式版本）canary （64 位）
**no subscribed stream!!!**

I'll explain why comment this for a test later.
I know that something about ExtensionId, I'll try my test, and give a PR soon.

lidedongsn:
版本 73.0.3683.103（正式版本） （64 位）answer sdp video extmap：
```
a=extmap:2 urn:ietf:params:rtp-hdrext:toffset
a=extmap:3 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:4 urn:3gpp:video-orientation
a=extmap:6 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
```
版本 75.0.3766.2（正式版本）canary （64 位） answer sdp video extmap:
```
a=extmap:3 urn:3gpp:video-orientation
a=extmap:5 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:13 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:14 urn:ietf:params:rtp-hdrext:toffset
```
73 publish 75 subscribe,
captrue rtp using wireshark, a vp8 packet as example
before pass licode:
```
Header extensions
    RFC 5285 Header Extension (One-Byte Header)
        Identifier: 3
        Length: 3
        Extension Data: 0ac6a8

```
Identifier: 3 as http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
after pass licode to 75:
```
Header extensions
    RFC 5285 Header Extension (One-Byte Header)
        Identifier: 3
        Length: 3
        Extension Data: 0ac6a8

```
75 receive Identifier: 3 as  urn:3gpp:video-orientation

when comment comment file rtp_media_config.js
```
//"urn:3gpp:video-orientation",
```
75 will not receive Identifier: 3  rtp

It's wrong anyhow!
so need translate ExtensionId as needed
see PR #1400

kekkokk:
this is a topic that was already discussed https://github.com/lynckia/licode/pull/1287 but was never merged.
[...]

PR #1400, opened 2019-04-16
add ExtensionId translator

lidedongsn:
fix issue #1393 #1399 
add ExtensionId translator
Publisher：  translate external ExtensionId into internal ExtensionId 
Subscriber：translate internal ExtensionId into external ExtensionId

### DIFF
--- a/erizo/src/erizo/MediaStream.cpp
+++ b/erizo/src/erizo/MediaStream.cpp
@@ -714,7 +714,15 @@ void MediaStream::changeDeliverExtensionId(DataPacket *dp, packetType type) {
           extId = extByte >> 4;
           extLength = extByte & 0x0F;
           // extId == 0 should never happen, see https://tools.ietf.org/html/rfc5285#section-4.2
+if (extId != 0) {
           for (int i = 1; i < 15; i++) {
+            if (extMap.at(i) == extId) {
+              extBuffer[0] = (extBuffer[0] | 0xF0) & (i << 4 | 0x0F);
+            }
+          }
+         }
+          extBuffer = extBuffer + extLength + 2;
+          currentPlace = currentPlace + extLength + 2;         
             if (extMap.at(i) == extId) {
               extBuffer[0] = (extBuffer[0] | 0xF0) & (i << 4 | 0x0F);
             }
--- a/erizo/src/erizo/MediaStream.cpp
+++ b/erizo/src/erizo/MediaStream.cpp
@@ -703,7 +703,7 @@ void MediaStream::changeDeliverExtensionId(DataPacket *dp, packetType type) {
           break;
       }
       uint16_t totalExtLength = h->getExtLength();
-      if (h->getExtId() == 0xBEDE) { // One-Byte Header
+      if (h->getExtId() == 0xBEDE) {  // One-Byte Header
         char* extBuffer = (char*)&h->extensions;  // NOLINT
         uint8_t extByte = 0;
         uint16_t currentPlace = 1;
@@ -765,7 +765,7 @@ void MediaStream::parseIncomingExtensionId(char *buf, int len, packetType type) 
           break;
       }
       uint16_t totalExtLength = h->getExtLength();
-      if (h->getExtId() == 0xBEDE) { // One-Byte Header
+      if (h->getExtId() == 0xBEDE) {  // One-Byte Header
         char* extBuffer = (char*)&h->extensions;  // NOLINT
         uint8_t extByte = 0;
         uint16_t currentPlace = 1;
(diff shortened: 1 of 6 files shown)


paul-maxime:
Yes, I had the same problem. The extension IDs are not always the same between each clients, especially between firefox and chrome.
Translating them like the payload IDs is a nice solution.

erizo/src/erizo/MediaStream.cpp
```diff
@@ -677,6 +682,51 @@ void MediaStream::getJSONStats(std::function<void(std::string)> callback) {
...
+        char* extBuffer = (char*)&h->extensions;  // NOLINT
+        uint8_t extByte = 0;
+        uint16_t currentPlace = 1;
+        uint8_t extId = 0;
+        uint8_t extLength = 0;
+        while (currentPlace < (totalExtLength*4)) {
+          extByte = (uint8_t)(*extBuffer);
+          extId = extByte >> 4;
+          extLength = extByte & 0x0F;
+          for (int i = 0; i < 20; i++) {
```
lodoyun: We should never translate to or from `0`. Adding a check to ignore `extId == 0` should suffice. This causes issues in my tests.

erizo/src/erizo/rtp/RtpExtensionProcessor.h
```diff
@@ -47,7 +47,7 @@ class RtpExtensionProcessor{
 
  private:
   std::vector<ExtMap> ext_mappings_;
-  std::array<RTPExtensions, 10> ext_map_video_, ext_map_audio_;
+  std::array<RTPExtensions, 20> ext_map_video_, ext_map_audio_;
```
lodoyun: I understand 10 was an arbitrary value but if we go ahead and change it, why not make it actually match the space for the Extension Ids? (1 - 14 according to https://tools.ietf.org/html/rfc5285) so, with `15` we would also cover 0.

erizo/src/erizo/MediaStream.cpp
```diff
@@ -677,6 +682,52 @@ void MediaStream::getJSONStats(std::function<void(std::string)> callback) {
...
+        char* extBuffer = (char*)&h->extensions;  // NOLINT
+        uint8_t extByte = 0;
+        uint16_t currentPlace = 1;
+        uint8_t extId = 0;
+        uint8_t extLength = 0;
+        while (currentPlace < (totalExtLength*4)) {
+          extByte = (uint8_t)(*extBuffer);
+          extId = extByte >> 4;
+          extLength = extByte & 0x0F;
+          // extId == 0 should never happen, see https://tools.ietf.org/html/rfc5285#section-4.2
```
lodoyun: `extId==0` will happen when padding is needed, look at the example in the rfc. we just need to ignore it because the translation does not make sense.

erizo/src/erizo/MediaStream.cpp
```diff
@@ -677,6 +682,52 @@ void MediaStream::getJSONStats(std::function<void(std::string)> callback) {
...
+          for (int i = 1; i < 15; i++) {
```
lodoyun: This is what I'm talking about. 
```suggestion
if (extId != 0) {
          for (int i = 1; i < 15; i++) {
            if (extMap.at(i) == extId) {
              extBuffer[0] = (extBuffer[0] | 0xF0) & (i << 4 | 0x0F);
            }
          }
         }
          extBuffer = extBuffer + extLength + 2;
          currentPlace = currentPlace + extLength + 2;         
```

lidedongsn: @lodoyun would you mind fixing the coding error over building

lidedongsn: @lodoyun I just ignore ```extId==0``` at for (int **i = **1****; i < 15; i++)

lodoyun: The linter is failing because I did not respect the indentation, I probably shouldn't have used the suggestions feature. Just take the idea and use your own code.

zhangshishen:
0                   1                   2                   3
       0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
      |       0xBE    |    0xDE       |           length=3            |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
      |  ID   | L=0   |     data      |  ID   |  L=1  |   data...
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
            ...data   |    0 (pad)    |    0 (pad)    |  ID   | L=3   |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
      |                          data                                 |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

I think extension with 1 byte padding will cause error. According to https://tools.ietf.org/html/rfc5285#section-4.2

PR #1402, opened 2019-04-25
Connections now persist with Clients when in singlePC

lodoyun:
**Description**
This PR allows connections (the PeerConnection-WebRTCConnection pair) to remain open while a client stays in a room in SinglePC. That means when unpublishing or unsubscribing from the last stream, the connection won't be immediately closed, allowing the client to publish/subscribe again without the added cost of creating a new Connection (ice negotiation, etc.)

This also changes the way ErizoJS processes are handled, they are not closed when the last publisher leaves, they are now notified when clients leave ErizoController and the process ends when the last client leaves.

### DIFF
--- a/erizo_controller/erizoController/models/ErizoList.js
+++ b/erizo_controller/erizoController/models/ErizoList.js
@@ -43,9 +43,11 @@ class ErizoList extends EventEmitter {
     return this.erizos[this.getInternalPosition(position)];
   }
 
-  forEachExisting(task) {
+  forEachUniqueErizo(task) {
+    const uniqueIds = new Set();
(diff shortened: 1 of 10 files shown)


erizo_controller/erizoController/models/ErizoList.js
```diff
@@ -43,9 +43,11 @@ class ErizoList extends EventEmitter {
     return this.erizos[this.getInternalPosition(position)];
   }
 
-  forEachExisting(task) {
+  forEachUniqueErizo(task) {
```
jcague: 👍

PR #1405, opened 2019-05-09
Fix h264 keyframe parse error

yuanchao0310:
This PR fix H264 keyframe parse error:

- if the packet is FU-A (fragmented nal) type
     set keyframe flag to true when nal_type==5 and start_bit==1

- if the packet is STAP-A (one packet, multiple nals), sometimes the keyframe size is smaller than MTU, especially low resolution video, so publisher will put sps,pps and keyframe into one packet. need to parse it in such case.
    parse all nals type and  set keyframe flag to true if one nal_type==5

### DIFF
--- a/erizo/src/erizo/rtp/RtpH264Parser.cpp
+++ b/erizo/src/erizo/rtp/RtpH264Parser.cpp
@@ -103,7 +103,7 @@ int RtpH264Parser::parse_packet_fu_a(RTPPayloadH264* h264, unsigned char* buf, i
   buf += 2;
   len -= 2;
 
-  if (nal_type == 5) {
+  if (nal_type == 5 && start_bit == 1) {
     h264->frameType = kH264IFrame;
   }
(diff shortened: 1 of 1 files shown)


jcague:
thanks for the contribution and clarification @yuanchao0310 ! did you test it with real data?

yuanchao0310:
> thanks for the contribution and clarification @yuanchao0310 ! did you test it with real data?

Yes. I test it on Android phone.
I found the licode MediaServer send PLI request to Android phone(China 360 brand), and Android Phone logs show it received the PLI request and generated one keyframe, but MediaServer cannot detect the keyframe and repeat send the PLI request every 200ms.(erizo/src/erizo/rtp/PliPacerHandler.cpp)
From the tcpdump data, I found that the Android phone put sps, pps and keyframe into one RTP packet(STAP-A type, one packet, multiple nals).

PR #1409, opened 2019-05-13
fix dtls multi-thread crash issue

yuanchao0310:
**Description**

fix dtls multi-thread crash issue:
refer to:
https://linux.die.net/man/3/crypto_set_locking_callback
http://curl.haxx.se/libcurl/c/opensslthreadlock.html
http://curl.haxx.se/libcurl/c/threaded-ssl.html

### DIFF
--- a/erizo/src/erizo/MediaStream.cpp
+++ b/erizo/src/erizo/MediaStream.cpp
@@ -497,9 +497,11 @@ void MediaStream::read(std::shared_ptr<DataPacket> packet) {
       // Deliver data
       if (isVideoSourceSSRC(recvSSRC) && video_sink_) {
         parseIncomingPayloadType(buf, len, VIDEO_PACKET);
+        parseIncomingExtensionId(buf, len, VIDEO_PACKET);
         video_sink_->deliverVideoData(std::move(packet));
       } else if (isAudioSourceSSRC(recvSSRC) && audio_sink_) {
         parseIncomingPayloadType(buf, len, AUDIO_PACKET);
+        parseIncomingExtensionId(buf, len, AUDIO_PACKET);
         audio_sink_->deliverAudioData(std::move(packet));
       } else {
         ELOG_DEBUG("%s read video unknownSSRC: %u, localVideoSSRC: %u, localAudioSSRC: %u",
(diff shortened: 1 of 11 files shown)


yuanchao0310:
Sample code to reproduce the crash bug, it similar with the constructor of DtlsTransport.cpp (erizo/src/erizo/DtlsTransport.cpp)

```
#include "dtls/DtlsSocket.h"
#include <thread>

const int thread_num = 100;

void func() {
  boost::scoped_ptr<dtls::DtlsSocketContext> dtlsRtp;
  dtlsRtp.reset(new dtls::DtlsSocketContext());
  dtlsRtp->createClient();
}

int main()
{
  dtls::DtlsSocketContext::Init();
  std::thread th[thread_num];
  for (int i = 0; i < 100; i++ ) {
    for (int j = 0; j < thread_num; j++) {
      th[j] = std::thread(func);
    }
    for (int j = 0; j < thread_num; j++) {
      th[j].join();
    }
  }

  return 0;
}
```

the coredump backtrace looks like:
#0  0x0000000001af7c68 in rsa_pub_cmp ()
#1  0x0000000001a77d81 in X509_check_private_key ()
#2  0x00000000019fb793 in SSL_CTX_use_PrivateKey ()
#3  0x0000000000c0a06a in dtls::DtlsSocketContext::DtlsSocketContext (this=0x7fccccffe000) at /opt/download/licode/erizo/src/erizo/dtls/DtlsClient.cpp

yuanchao0310:
This PR use Linux pthread_mutex_* APIs.
if can be build with c++11 , I think it will be better to use std::mutex, looks like:

yuanchao0310:
Another example code from ffmpeg:
https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/tls_openssl.c

jcague:
@yuanchao0310 I submitted a proposal with changes for this PR, including a unit test: https://github.com/yuanchao0310/licode/pull/1

yuanchao0310:
> @yuanchao0310 I submitted a proposal with changes for this PR, including a unit test: [yuanchao0310#1](https://github.com/yuanchao0310/licode/pull/1)

Merged. Please review again.

PR #1413, opened 2019-05-17
BWE Fixes for publisher estimation in chrome >72

lodoyun:
**Description**

Another possible solution for the bandwidth estimation problems when not using simulcast in Chrome >72.
This PR drops all other RR packets from subscribers and relies on the sender estimation implementation in licode to update REMB from subscribers.

Also, this PR updated RR packets generated in Licode so the packets lost are monotonically increasing. There are different ways to do this but, at this time, we chose to use the packet loss ratio when it's > 0.

If tests are positive, in the future, we will work on separating the logic that decides what RR packets to forward in a new handler.

### DIFF
--- a/erizo/src/erizo/rtp/QualityFilterHandler.cpp
+++ b/erizo/src/erizo/rtp/QualityFilterHandler.cpp
@@ -36,13 +36,16 @@ void QualityFilterHandler::handleFeedbackPackets(const std::shared_ptr<DataPacke
            chead->getBlockCount() == RTCP_FIR_FMT)) {
       sendPLI();
     }
-  });
+    });
 }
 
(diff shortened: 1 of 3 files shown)


kekkokk:
if it helps, we decided long time ago to terminate all RR and REMBS from subscribers.
we only send the RR generated from server and send the REMBs calculated from the lowest between the server calculated and the lowest REMB received from subscribers.

lodoyun:
@kekkokk That approach should also be valid but I prefer Chrome doing the actual selecting from the lowest REMB if I can. Also, we're potentially getting a more accurate estimation since we're running the full sender side algorithm taking into account, delays and packet loss ratio from the subscribers instead of blindly forwarding REMBs, which only paints a part of the picture.

Issue #1356, opened 2019-02-07
VP9 codec


ashar02:
After installation from source code on ubuntu 14.04. I disabled vp8 codec and made vp9 as default video codec in rtp_media_config.js. But when I joined the room video got stuck after few seconds. Am I doing any thing wrong?? Anyone can help?

ashar02:
I verified using webrtc-internals that vp9 is selected as video codec.

ashar02:
@jcague please comment?

ashar02:
@jcague to run it correctly do I have to enable simulcast in my environment?

jcague:
no you don't need to enable simulcast, I'll try to reproduce your issue shortly

ashar02:
@jcague every solution in market is using vp9 to save bandwidth having same quality. In licode it starts properly but then it stops after few seconds. Are you able to produce it? Can you give me hint where to see in code to run it properly?

jcague:
This issue could be related with the things @lodoyun solved in #1413 , have you tested with the last version in the master branch?

ashar02:
Let me check and let you know today!!

ashar02:
I checkout the code from master head. Use ubuntu 16.04. I changed the default: { rtpMappings: { vp8, opus }, extMappings } to default: { rtpMappings: { vp9, opus }, extMappings }, in file rtp_media_config.js

And ran the server and I noticed the problem persist. After few seconds remote video freezes.
<img width="1680" alt="Screen Shot 2019-06-05 at 6 23 21 PM" src="https://user-images.githubusercontent.com/1763187/58960159-ffb3fd80-87bf-11e9-8cb7-5ef99258e50a.png">
<img width="1680" alt="Screen Shot 2019-06-05 at 6 24 55 PM" src="https://user-images.githubusercontent.com/1763187/58960173-0b072900-87c0-11e9-9f11-a949efd3d7a4.png">
<img width="1680" alt="Screen Shot 2019-06-05 at 6 32 08 PM" src="https://user-images.githubusercontent.com/1763187/58960271-4ace1080-87c0-11e9-8c29-714b2dffa6a8.png">

So it seems that publisher is sending the stream correctly but at subscription end after few seconds it stops receiving the video stream in case of vp9.

ashar02:
So how to narrow this issue?

ashar02:
@jcague is there any impact of #1419 on vp9?

jcague:
[...]

PR #1420, opened 2019-06-07
Fix the way we detect layers in VP9 parser

jcague:
**Description**

We were not setting pictureId and right TIDs and SIDs in VP9. There is still some work to do to fix SVC, but it is not a priority since it is not enabled by default in Chrome (we still need a magic argument).

Not needed.


jcague:
This PR should fix #1356

PR #1421, opened 2019-06-07
Add Black keyframes in mute video

lodoyun:
**Description**

This PR solves a race condition when muting video before receiving any keyframes. To solve this, this PR implements transforming all keyframes into black keyframes when muted (and using VP8).
Also, to solve possible problems when unmuting a video, it will wait until a keyframe comes to start forwarding video again.

### DIFF
--- a/erizo/src/erizo/rtp/FakeKeyframeGeneratorHandler.cpp
+++ b/erizo/src/erizo/rtp/FakeKeyframeGeneratorHandler.cpp
@@ -71,6 +71,7 @@ std::shared_ptr<DataPacket> FakeKeyframeGeneratorHandler::transformIntoKeyframeP
   (std::shared_ptr<DataPacket> packet) {
     if (packet->codec == "VP8") {
       auto keyframe_packet = RtpUtils::makeVP8BlackKeyframePacket(packet);
+      packet->is_keyframe = true;
       return keyframe_packet;
     } else {
       ELOG_DEBUG("Generate keyframe packet is not available for codec %s", packet->codec);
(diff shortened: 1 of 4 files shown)


erizo/src/erizo/rtp/FakeKeyframeGeneratorHandler.cpp
```diff
@@ -71,6 +71,7 @@ std::shared_ptr<DataPacket> FakeKeyframeGeneratorHandler::transformIntoKeyframeP
   (std::shared_ptr<DataPacket> packet) {
     if (packet->codec == "VP8") {
       auto keyframe_packet = RtpUtils::makeVP8BlackKeyframePacket(packet);
+      packet->is_keyframe = true;
```
jcague: good catch!

erizo/src/erizo/rtp/RtpTrackMuteHandler.cpp
```diff
@@ -128,4 +147,19 @@ inline void RtpTrackMuteHandler::setPacketSeqNumber(std::shared_ptr<DataPacket>
...
+
+
+std::shared_ptr<DataPacket> RtpTrackMuteHandler::transformIntoBlackKeyframePacket
+  (std::shared_ptr<DataPacket> packet) {
+    if (packet->codec == "VP8") {
+      auto keyframe_packet = RtpUtils::makeVP8BlackKeyframePacket(packet);
+      return keyframe_packet;
+    } else {
+      ELOG_WARN("%s cannot generate keyframe packet is not available for codec %s", stream_->toLog(), packet->codec);
+      return packet;
```
jcague: IMO we shouldn't send packets if video is muted in other codecs.

kekkokk:
the new Webrtc core present in firefox chrome and safari, if the peerconnection is audio and video, expect at least one keyframe to sync a/v and so start playing audio.
we noticed few issue when joining as a subscriber with the video muted option.
This prevents the client to play the audio. 
I will adopt a different mechanism and send a black keyframe every second to be sure the client can sync or at least request a keyframe also when activating the mute and then rewrite this with a black frame. this prevents: a) a freezed image and b) the audio problem

lodoyun:
@kekkokk We don't request them periodically but we automatically transform all the first packets until a keyframe into black keyframes (in another handler) and, when muted, we always let keyframes pass, also transforming them into black keyframes.

PR #1423, opened 2019-06-12
Fix SDP negotiation for resubscriptions

jcague:
**Description**

This PR aims to reapply old commits that solved cases where we want to unsubscribe and subscribe again. It is mainly focused for Single Peer Connection, but it also affects multiple Peer Connections.

The old commits introduced a set of issues like Erizo crashes, negotiation failures, etc. that I have been fixing in this PR with more commits.

Not needed.


erizo_controller/erizoClient/src/Room.js
```diff
@@ -83,6 +83,7 @@ const Room = (altIo, altConnectionHelpers, altConnectionManager, specInput) => {
     if (!that.p2p) {
       stream.pc.addStream(stream);
     }
+    stream.state = 'subscribed';
```
lodoyun: In the future we should think about having a more involved fsm like we're using in other places. Even maybe just having all the states set as a constant somewhere so we can see all the possible states at a glance would be an improvement.

PR #1428, opened 2019-07-08
Use shared_ptr instead of raw pointer for erizo OTM in erizoAPI

lodoyun:
**Description**

We were still using a raw pointer in ErizoAPI with OneToManyProcessor. This could potentially cause crashes with async operations since we had no way to protect the pointer. This PR changes this so the erizo object is managed as a shared_ptr in the same way we do with WebRTCConnection or MediaStream.

### DIFF
--- a/erizoAPI/OneToManyProcessor.cc
+++ b/erizoAPI/OneToManyProcessor.cc
@@ -16,13 +16,13 @@ Nan::Persistent<Function> OneToManyProcessor::constructor;
 // Classes for Async (not in node main thread) operations
 class AsyncDeleter : public Nan::AsyncWorker {
  public:
-    AsyncDeleter(erizo::OneToManyProcessor* otm, Nan::Callback *callback):
+    AsyncDeleter(std::shared_ptr<erizo::OneToManyProcessor> otm, Nan::Callback *callback):
       AsyncWorker(callback), otmToDelete_(otm) {
       }
     ~AsyncDeleter() {}
     void Execute() {
       otmToDelete_->close();
-      delete otmToDelete_;
+      otmToDelete_.reset();
     }
     void HandleOKCallback() {
       Nan::HandleScope scope;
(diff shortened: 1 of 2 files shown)


erizoAPI/OneToManyProcessor.cc
```diff
@@ -102,11 +107,15 @@ NAN_METHOD(OneToManyProcessor::close) {
   }
 
   Nan::AsyncQueueWorker(new  AsyncDeleter(me, callback));
+  obj->me.reset();
```
jcague: 👍

erizoAPI/OneToManyProcessor.cc
```diff
@@ -84,16 +86,19 @@ NAN_MODULE_INIT(OneToManyProcessor::Init) {
 
 NAN_METHOD(OneToManyProcessor::New) {
   OneToManyProcessor* obj = new OneToManyProcessor();
-  obj->me = new erizo::OneToManyProcessor();
-  obj->msink = obj->me;
+  obj->me = std::make_shared<erizo::OneToManyProcessor>();
+  obj->msink = obj->me.get();
```
jcague: we still use raw pointers through MediaSink and MediaSource, that's something we must change in a future PR I think

erizoAPI/OneToManyProcessor.cc
```diff
@@ -36,23 +36,25 @@ class AsyncDeleter : public Nan::AsyncWorker {
...
+      if (auto locked_otm = weak_otm_.lock()) {
+        locked_otm->removeSubscriber(peerId_);
+      }
     }
     void HandleOKCallback() {
       // We're not doing anything here ATM
     }
  private:
-    erizo::OneToManyProcessor* otm_;
+    std::weak_ptr<erizo::OneToManyProcessor> weak_otm_;
```
jcague: 👍

erizoAPI/OneToManyProcessor.cc
lodoyun: Agreed

PR #1429, opened 2019-07-10
Fixed race conditions with promises when closing Nodes

lodoyun:
**Description**

This PR fixes two problems with `closeNode`:
1 - MediaStream close promise was not resolved because the `future_async_` was closed as part of the process, now the async will be closed after it notifies the completion of the task
2 - When WebRTCConnection is closed as part of closeNode because we're closing the last mediaStream in that connection, we would not resolve the `removeStream` promise. Now we make sure the stream is removed and the promise resolved before checking if we should close the connection

### DIFF
--- a/erizoAPI/MediaStream.cc
+++ b/erizoAPI/MediaStream.cc
@@ -53,10 +53,10 @@ Nan::Persistent<Function> MediaStream::constructor;
 MediaStream::MediaStream() : closed_{false}, id_{"undefined"} {
   async_stats_ = new uv_async_t;
   async_event_ = new uv_async_t;
-  future_async_ = new uv_async_t;
+  close_future_async_ = new uv_async_t;
   uv_async_init(uv_default_loop(), async_stats_, &MediaStream::statsCallback);
   uv_async_init(uv_default_loop(), async_event_, &MediaStream::eventCallback);
-  uv_async_init(uv_default_loop(), future_async_, &MediaStream::closePromiseResolver);
+  uv_async_init(uv_default_loop(), close_future_async_, &MediaStream::closePromiseResolver);
 }
 
 MediaStream::~MediaStream() {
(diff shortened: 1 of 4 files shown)


erizoAPI/MediaStream.cc
```diff
@@ -56,7 +56,7 @@ MediaStream::MediaStream() : closed_{false}, id_{"undefined"} {
   future_async_ = new uv_async_t;
   uv_async_init(uv_default_loop(), async_stats_, &MediaStream::statsCallback);
   uv_async_init(uv_default_loop(), async_event_, &MediaStream::eventCallback);
-  uv_async_init(uv_default_loop(), future_async_, &MediaStream::promiseResolver);
+  uv_async_init(uv_default_loop(), future_async_, &MediaStream::closePromiseResolver);
```
jcague: np: shouldn't we call this `close_future_async_`?

erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -100,18 +100,18 @@ exports.ErizoJSController = (threadPool, ioThreadPool) => {
...
-    }
-    return closePromise;
+    return closePromise.then(() => {
+      const client = clients.get(clientId);
+      if (client === undefined) {
+        log.debug('message: trying to close node with no associated client,' +
+          `clientId: ${clientId}, streamId: ${node.streamId}`);
+        return;
+      }
+      const remainingConnections = client.maybeCloseConnection(connection.id);
```
jcague: heads up: this could result in new issues given that maybeCloseConnection is now async

erizoAPI/MediaStream.cc
```diff
@@ -506,21 +507,24 @@ void MediaStream::notifyFuture(Nan::Persistent<v8::Promise::Resolver> *persisten
...
   MediaStream* obj = reinterpret_cast<MediaStream*>(async->data);
-  if (!obj || !obj->me) {
+  if (!obj) {
     return;
   }
   boost::mutex::scoped_lock lock(obj->mutex);
-  ELOG_DEBUG("%s, message: promiseResolver", obj->toLog());
+  ELOG_DEBUG("%s, message: closePromiseResolver", obj->toLog());
   obj->futures_manager_.cleanResolvedFutures();
+  obj->Ref();
```
jcague: :+1:

erizoAPI/MediaStream.cc
lodoyun: Sure 👍

erizo_controller/erizoJS/erizoJSController.js
lodoyun: I agree, and we will have to test and monitor this. However, having the WebRTCConnection object being destroyed because of maybeCloseConnection while removeStream is potentially still running is also not ideal.

PR #1431, opened 2019-07-12
Enforce optimizations in OpenSSL and libSRTP with cflags

jcague:
**Description**

This PR adds compilation flags to openssl and libsrtp builds that would theoretically optimize these libraries and save CPU.

Not needed.


jcague:
this does not make sense anymore as we saw these flags are used by default

PR #1435, opened 2019-07-22
fix bugs when Creates a new token

darkterrorooo:
**Description**


darkterrorooo:
I found this bugs when doing stress testing .
this bug may cause nuve to return respone twice when create token.

jcague:
@darkterrorooo this fix looks promising! It seems like lint is complaining though, could you fix those minor issues please?

PR #1445, opened 2019-08-02
Add a client API to update simulcast layer bitrate

jcague:
**Description**

Adds a client API to update simulcast layer bitrate.

Note: It also fixes an issue with the updateConfiguration when run for publishers.

This is the new API in the cliend side.

```
localStream.updateSimulcastLayersBitrate({0: 80000, 1: 430000});
```

In this example we are configuring 2 spatial layers bitrates, limiting the lower layer to 80 Kbps and the higher to 430 Kbps.

### DIFF
--- a/erizo_controller/erizoJS/erizoJSController.js
+++ b/erizo_controller/erizoJS/erizoJSController.js
@@ -218,7 +218,7 @@ exports.ErizoJSController = (erizoJSId, threadPool, ioThreadPool) => {
       return;
     }
 
-    if (publisher.cliendId === clientId) {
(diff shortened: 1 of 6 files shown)


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -218,7 +218,7 @@ exports.ErizoJSController = (erizoJSId, threadPool, ioThreadPool) => {
       return;
     }
 
-    if (publisher.cliendId === clientId) {
+    if (publisher.clientId === clientId) {
```
lodoyun: Good catch 👍

PR #1447, opened 2019-08-07
Fix crashes when getting local sdp

lodoyun:
**Description**

This PR fixes two possible crashes in the life of SdpInfo in getLocalDescription:
1. The path when we cannot call `getLocalSdpInfoSync` would crash because `this_ptr` would not be present.
2. We made the copy of SdpInfo when the promise is resolved in `ErizoAPI`, we could be potentially modifying that data structure in another thread at the same time.

### DIFF
--- a/erizo/src/erizo/WebRtcConnection.cpp
+++ b/erizo/src/erizo/WebRtcConnection.cpp
@@ -275,9 +275,10 @@ boost::future<std::shared_ptr<SdpInfo>> WebRtcConnection::getLocalSdpInfo() {
   worker_->task([weak_this, task_promise] {
     std::shared_ptr<SdpInfo> info;
     if (auto this_ptr = weak_this.lock()) {
-      info = this_ptr->getLocalSdpInfoSync();
+      auto temp_info = this_ptr->getLocalSdpInfoSync();
+      info = std::make_shared<SdpInfo>(*temp_info);
     } else {
-      ELOG_WARN("%s message: Error trying to getLocalSdpInfo, returning empty", this_ptr->toLog());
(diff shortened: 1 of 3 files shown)


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -331,7 +331,8 @@ std::shared_ptr<SdpInfo> WebRtcConnection::getLocalSdpInfoSync() {
     local_sdp_->videoDirection = erizo::SENDRECV;
   }
 
-  return local_sdp_;
+  auto local_sdp_copy = std::make_shared<SdpInfo>(*local_sdp_.get());
```
jcague: Good catch!

erizoAPI/WebRtcConnection.cc
```diff
@@ -356,9 +356,14 @@ NAN_METHOD(WebRtcConnection::getLocalDescription) {
...
-        std::shared_ptr<erizo::SdpInfo> sdp_info = std::make_shared<erizo::SdpInfo>(*fut.get().get());
-        ResultVariant result = sdp_info;
-        obj->notifyFuture(persistent, result);
+        std::shared_ptr<erizo::SdpInfo> sdp_info = fut.get();
+        if (sdp_info) {
+          std::shared_ptr<erizo::SdpInfo> sdp_info_copy = std::make_shared<erizo::SdpInfo>(*sdp_info.get());
+          ResultVariant result = sdp_info_copy;
+          obj->notifyFuture(persistent, result);
+        } else {
+          obj->notifyFuture(persistent);
```
jcague: we could also reject promises at some point (in future PRs)

PR #1450, opened 2019-08-28
Add a new mechanism to enable fallback mode on high audio losses

jcague:
**Description**

It adds a new mechamism to enable fallback in Simulcast when we are experiencing high audio packet losses in the connection.

It automatically goes to fallback if audio fraction loss is higher than 20% (HIGH_AUDIO_LOSSES level), and it enables normal fallback when audio losses remains higher than 5% (LOW_AUDIO_LOSSES level). It switches off fallback when the quality level improves during 20 seconds or more.


erizo/src/erizo/bandwidth/ConnectionQualityCheck.cpp
```diff
@@ -0,0 +1,74 @@
...
+  if (streams.size() != buffer_.capacity()) {
+    buffer_.set_capacity(streams.size());
+  }
+  int reports_count = 0;
+  RtpUtils::forEachRtcpBlock(packet, [streams, this, &reports_count](RtcpHeader *chead) {
+    reports_count++;
+    uint32_t ssrc = chead->isFeedback() ? chead->getSourceSSRC() : chead->getSSRC();
+    bool is_rr = chead->isReceiverReport();
+    if (is_rr) {
+      ELOG_WARN("  RR Block_COUNT %d", chead->getBlockCount());
```
lodoyun: Just a reminder to switch the log levels to DEBUG or INFO before we merge this.

erizo/src/erizo/bandwidth/ConnectionQualityCheck.cpp
```diff
@@ -0,0 +1,74 @@
...
+    } else {
+      ELOG_WARN("  %d", chead->getPacketType());
+    }
+    uint8_t fraction_lost = is_rr ? chead->getFractionLost() : 0;
+    std::for_each(streams.begin(), streams.end(),
+        [ssrc, fraction_lost, is_rr, this] (const std::shared_ptr<MediaStream> &media_stream) {
+      bool is_audio = media_stream->isAudioSourceSSRC(ssrc) || media_stream->isAudioSinkSSRC(ssrc);
```
lodoyun: While I think using audio loss is a good strategy, this would invalidate this mechanism in all streams with muted audio.
Consider evaluating fraction_loss also for video ssrcs. A possible compromise would be to assign a higher weight to audio loss when computing the mean.

erizo/src/erizo/rtp/RtpHeaders.h
```diff
@@ -411,6 +411,9 @@ class RtcpHeader {
         packettype == RTCP_PS_Feedback_PT ||
         packettype == RTCP_RTP_Feedback_PT);
   }
+  inline bool isReceiverReport() {
```
lodoyun: 👍

PR #1456, opened 2019-09-10
Add a basic pipeline to webrtcconnection

jcague:
**Description**

This PR adds a basic Pipeline to WebRtcConnection. The idea is to provide it with a tool to easily add more and more functionality to the connection/transport level. In the future we might decide for instance to move all the BWE logic to the connection better than having it in the MediaStream.

Not needed.

### DIFF
--- a/erizo/src/erizo/WebRtcConnection.h
+++ b/erizo/src/erizo/WebRtcConnection.h
@@ -22,6 +22,7 @@
 #include "rtp/RtpExtensionProcessor.h"
 #include "lib/Clock.h"
 #include "pipeline/Handler.h"
+#include "pipeline/HandlerManager.h"
 #include "pipeline/Service.h"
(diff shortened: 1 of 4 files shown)


erizo/src/erizo/WebRtcConnection.h
```diff
@@ -208,6 +216,53 @@ class WebRtcConnection: public TransportListener, public LogContext,
...
+
+class ConnectionPacketReader : public InboundHandler {
+ public:
+  explicit ConnectionPacketReader(WebRtcConnection *connection) : connection_{connection} {}
+
+  void enable() override {}
+  void disable() override {}
+
+  std::string getName() override {
+    return "connection-reader";
```
lodoyun: This makes me think we could, at some point, change the mediaStream equivalent's name to "mediastream-reader"

PR #1398, opened 2019-04-10
Adds unified stream management in ErizoController

lodoyun:
**Description**
This PR refactors the way ErizoController keeps control over the streams published and subscribed.
Before this PR we keep data structures in `Room`, `RoomController` and `ErizoList` to account for different parts of a stream's life. Those data structures were updated by `RoomController`, `Client`, mostly.

This PR updates that management to make it clearer and less error-prone. All Published streams in a Room are stored in a `StreamManager` and each `PublishedStream` will have lists with its avsubscriptions, data subscribers and externalOutputs. Streams and subscriptions are always created and updated in `Client` with a few exceptions ( the erizoId associated with a stream is updated in `roomController`). 

- [ ] It needs and includes Unit Tests

**Changes in Client or Server public APIs**

[] It includes documentation for these changes in `/doc`.

PR #1463, opened 2019-09-24
Refactor Stream Management in ErizoController

lodoyun:
**Description**
This PR updates #1398 with clean commits, resolved conflicts and a couple of new fixes.


erizo_controller/erizoController/models/Client.js
```diff
@@ -167,6 +180,9 @@ class Client extends events.EventEmitter {
                          'state: SUBSCRIBER_READY, ' +
                          `streamId: ${signMess.streamId}, ` +
                          `clientId: ${this.id}`);
+        this.roomController
+          .getPublishedStreamById(signMess.streamId)
```
jcague: I'd check if stream is not undefined

erizo_controller/erizoController/models/Client.js
```diff
@@ -176,6 +192,7 @@ class Client extends events.EventEmitter {
         log.error('message: addMultipleSubscribers timeout when contacting ErizoJS, ' +
                           `streamId: ${signMess.streamId}, ` +
                           `clientId: ${this.id}`);
+        this.roomController.getPublishedStreamById(signMess.streamId).removeAvSubscriber(this.id);
```
jcague: I'd check if stream is not undefined

erizo_controller/erizoController/models/Client.js
```diff
@@ -357,6 +382,141 @@ class Client extends events.EventEmitter {
     });
   }
 
+  _publishExternalInput(id, options, sdp, callback) {
```
jcague: I really like all these internal functions!

erizo_controller/erizoController/models/Stream.js
```diff
@@ -2,94 +2,199 @@
 
 /* eslint-disable no-param-reassign */
 
-exports.Stream = (spec) => {
-  const that = {};
-  const dataSubscribers = [];
-
-  that.getID = () => spec.id;
+const StreamStates = Object.freeze({
```
jcague: 💯

erizo_controller/erizoController/roomController.js
```diff
@@ -239,20 +220,17 @@ exports.RoomController = (spec) => {
...
               log.warn('message: addPublisher ErizoJS timeout no retry, ' +
                                  `retries: ${retries}, streamId: ${streamId}, ` +
                                  `erizoId: ${getErizoQueueFromStreamId(streamId)},`,
                                  logger.objectToLog(options.metadata));
-              const erizo = erizos.findById(publishers[streamId]);
-              if (erizo !== undefined) {
-                const index = erizo.publishers.indexOf(streamId);
-                erizo.publishers.splice(index, 1);
-              }
+              streamManager.updateErizoIdForPublishedStream(streamId, null);
```
jcague: Maybe I'm getting this wrong, shouldn't we remove the stream from the list of publishedStreams?

lodoyun: We do that in Client.js. The general rule is that roomController only updates the ErizoId for the published streams. The rest (adding/removing publishers, subscribers and update the state) should be done in the client

PR #1476, opened 2019-10-11
Add new Padding Manager in WebRtcConnection

jcague:
**Description**

I'm moving the logic that decides whether to enable or disable padding to the WebRtcConnection. It will also decide how much bitrate will be used and will distribute it among the subscribed streams in the same single peer connection.

Not needed.

### DIFF
--- a/extras/basic_example/public/script.js
+++ b/extras/basic_example/public/script.js
@@ -101,7 +101,7 @@ const startBasicExample = () => {
     req.send(JSON.stringify(roomData));
   };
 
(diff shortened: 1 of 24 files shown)


extras/basic_example/public/script.js
```diff
@@ -101,7 +101,7 @@ const startBasicExample = () => {
     req.send(JSON.stringify(roomData));
   };
 
-  const roomData = { username: 'user',
+  const roomData = { username: `user ${parseInt(Math.random() * 100, 10)}`,
```
lodoyun: 👍

PR #1487, opened 2019-10-23
Avoid audio drifting by updating audio SRs

jcague:
**Description**

We are having audio drifting because we are not correctly handling timestamps in video. This is a quick hack to avoid sending audio SRs properly and use a best effort approach when syncing audio and video in Chrome. 

Another better solution that updates timestamps in video SRs in Simulcast is coming.

Not needed.


erizo/src/erizo/OneToManyProcessor.cpp
```diff
@@ -33,7 +33,8 @@ namespace erizo {
     RtcpHeader* chead = reinterpret_cast<RtcpHeader*>(audio_packet->data);
     for (it = subscribers.begin(); it != subscribers.end(); ++it) {
       if ((*it).second != nullptr) {
-        if (chead->isRtcp()) {
+        // Hack to avoid audio drifting in Chrome.
```
lodoyun: 👍

PR #1499, opened 2019-11-07
Install JS deps via new package.json files

jcague:
**Description**

We were installing JS deps by using the command line directly. This is an attempt to do it via package.json files, which will help us better understanding which deps we have and use other tools to track secutiry vulnerabilities.

Not needed.


Dockerfile
```diff
@@ -32,6 +31,8 @@ RUN ./installErizo.sh -dfeacs && \
 
 WORKDIR /opt/licode
 
+ARG COMMIT
```
lodoyun: 👍 good catch

nuve/nuveClient/tools/compile.sh
```diff
@@ -5,6 +5,6 @@ set -e
 mkdir ../dist || true
 mkdir ../build || true
 
-google-closure-compiler-js ../src/N.js ../src/N.API.js > ../build/nuve.js
+../../node_modules/google-closure-compiler-js/cmd.js ../src/N.js ../src/N.API.js > ../build/nuve.js
```
lodoyun: 👍

PR #1505, opened 2019-11-27
Remove libnice and glib

lodoyun:
**Description**
We have been using NiceR for 2+ years without problems. This PR removes the libnice alternative, reducing the maintenance work and the testing that is required to maintain 2 libraries for IO. As a bonus, it should reduce image size and build time in some cases.


scripts/installMacDeps.sh
```diff
@@ -54,13 +54,13 @@ check_result() {
 install_homebrew_from_cache(){
   if [ -f cache/homebrew-cache.tar.gz ]; then
     tar xzf cache/homebrew-cache.tar.gz --directory /usr/local/Cellar
-    brew link glib pkg-config boost cmake yasm log4cxx gettext coreutils
+    brew link pkg-config boost cmake yasm log4cxx gettext coreutils
```
jcague: :+1:

erizo/src/erizo/IceConnection.h
```diff
@@ -1,5 +1,5 @@
 /*
- * LibNiceConnection.h
+ * IceConnection.h
```
jcague: :+1:

erizo/src/erizo/dtls/DtlsClient.cpp
```diff
@@ -55,7 +55,7 @@ unsigned long ssl_thread_id() {  // NOLINT
 }
 #else
 void ssl_thread_id(CRYPTO_THREADID *id) {  // NOLINT
-    CRYPTO_THREADID_set_numeric(id, (unsigned long)std::hash<std::thread::id>()(std::this_thread::get_id());  // NOLINT
+    CRYPTO_THREADID_set_numeric(id, (unsigned long)std::hash<std::thread::id>()(std::this_thread::get_id()));  // NOLINT
```
jcague: 🤔 good catch!

erizo/src/erizo/IceConnection.h
```diff
@@ -68,8 +66,8 @@ class IceConfig {
         turn_port{0},
         min_port{0},
         max_port{0},
-        should_trickle{false},
-        use_nicer{false} {
```
jcague: 💯

erizoAPI/WebRtcConnection.cc
```diff
@@ -150,8 +150,7 @@ NAN_METHOD(WebRtcConnection::New) {
     int maxPort = Nan::To<int>(info[6]).FromJust();
     bool trickle = Nan::To<bool>((info[7])).FromJust();
     Nan::Utf8String json_param(Nan::To<v8::String>(info[8]).ToLocalChecked());
-    bool use_nicer = Nan::To<bool>((info[9])).FromJust();
-    bool enable_connection_quality_check = Nan::To<bool>((info[10])).FromJust();
+    bool enable_connection_quality_check = Nan::To<bool>((info[9])).FromJust();
```
jcague: There could be other parameters inside info that are related to the ICE configuration (turnServer, username, etc.)

PR #1507, opened 2019-11-28
Fix RR handling for SenderBWE

jcague:
**Description**

There were still cases where the SenderBandwidthEstimationHandler was not calculating the bandwidth estimation properly. This PR fixes them. The problem was that we were updating the same estimator with RRs received from multiple streams.

Not needed.

### DIFF
--- a/erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
+++ b/erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
@@ -1,6 +1,8 @@
-#include "./MediaDefinitions.h"
 #include "rtp/SenderBandwidthEstimationHandler.h"
 
+#include <utility>
+
+#include "./MediaDefinitions.h"
 #include "rtp/RtpUtils.h"
 #include "./MediaStream.h"
 
@@ -12,7 +14,7 @@ constexpr duration SenderBandwidthEstimationHandler::kMinUpdateEstimateInterval;
 
 SenderBandwidthEstimationHandler::SenderBandwidthEstimationHandler(std::shared_ptr<Clock> the_clock) :
   connection_{nullptr}, bwe_listener_{nullptr}, clock_{the_clock}, initialized_{false}, enabled_{true},
-  received_remb_{false}, period_packets_sent_{0}, estimated_bitrate_{0}, estimated_loss_{0},
+  received_remb_{false}, estimated_bitrate_{0}, estimated_loss_{0},
   estimated_rtt_{0}, last_estimate_update_{clock::now()}, sender_bwe_{new SendSideBandwidthEstimation()},
   max_rr_delay_data_size_{0}, max_sr_delay_data_size_{0} {
     sender_bwe_->SetBitrates(kStartSendBitrate, kMinSendBitrate, kMaxSendBitrate);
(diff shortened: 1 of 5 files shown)


erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
```diff
@@ -128,6 +144,34 @@ void SenderBandwidthEstimationHandler::read(Context *ctx, std::shared_ptr<DataPa
...
+    uint64_t avg_delay = 0;
+    uint32_t rr_delay_data_size = rr_delay_data_.size();
+    int64_t now_ms = ClockUtils::timePointToMs(clock_->now());
+    std::for_each(rr_delay_data_.begin(), rr_delay_data_.end(),
+        [&avg_delay, &total_packets_lost, rr_delay_data_size] (const std::shared_ptr<RrDelayData> &rr_info) {
+          total_packets_lost += rr_info->packets_lost;
+          avg_delay += rr_info->delay / rr_delay_data_size;
+    });
+    if (period_packets_sent_ > 0) {
+      uint32_t fraction_lost = total_packets_lost * 255 / period_packets_sent_;
```
lodoyun: I think you can't calculate the fraction lost in a period from the packets lost you get in a Receiver Report. You would have to keep track of the differences in packet loss between two consecutive Receiver Reports since it indicates the cumulative lost packets.

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
```diff
@@ -50,33 +56,47 @@ void SenderBandwidthEstimationHandler::notifyUpdate() {
...
   stats_ = pipeline->getService<Stats>();
   if (!stats_) {
     return;
   }
   initialized_ = true;
 }
 
+void SenderBandwidthEstimationHandler::updateMaxListSizes() {
+  size_t num_streams = 0;
+  connection_->forEachMediaStream([&num_streams] (std::shared_ptr<MediaStream> media_stream) {
```
lodoyun: Since we are only taking into account video packets for `packets_sent`, shouldn't we check we're only using video ssrcs?
I'm not sure if we did that properly before though, we can discuss this.

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
```diff
@@ -50,33 +56,47 @@ void SenderBandwidthEstimationHandler::notifyUpdate() {
...
+  updateReceiverBlockFromList();
+}
+
 void SenderBandwidthEstimationHandler::read(Context *ctx, std::shared_ptr<DataPacket> packet) {
   RtpUtils::forEachRtcpBlock(packet, [this](RtcpHeader *chead) {
-    ELOG_DEBUG("%s ssrc %u, sourceSSRC %u, PacketType %u", connection_->toLog(),
-        chead->getSSRC(),
-        chead->getSourceSSRC(),
-        chead->getPacketType());
+//    ELOG_DEBUG("%s ssrc %u, sourceSSRC %u, PacketType %u", connection_->toLog(),
```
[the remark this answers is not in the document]
jcague: good catch, that's an annoying log when we set the log level to DEBUG so I'll remove it

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
jcague: good catch!

erizo/src/erizo/rtp/SenderBandwidthEstimantionHandler.cpp
jcague: good catch! I think it makes sense to check video and audio for RTP and RR packets.

PR #1527, opened 2019-12-16
Queue negotiation messages in ErizoJS

jcague:
**Description**

Queue SDP negotiation messages in ErizoJS to avoid race conditions. It handle interlocks between client and ErizoJS queues.

Not needed.

### DIFF
--- a/erizo_controller/erizoJS/models/Connection.js
+++ b/erizo_controller/erizoJS/models/Connection.js
@@ -7,6 +7,7 @@ const addon = require('./../../../erizoAPI/build/Release/addon');
 const logger = require('./../../common/logger').logger;
 const SessionDescription = require('./SessionDescription');
 const SemanticSdp = require('./../../common/semanticSdp/SemanticSdp');
+
 const log = logger.getLogger('Connection');
 
 const CONN_INITIAL = 101;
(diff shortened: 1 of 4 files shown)


erizo_controller/erizoJS/models/Connection.js
```diff
@@ -306,69 +280,143 @@ class Connection extends events.EventEmitter {
...
-        return this.processOffer(msg.sdp).catch(() => {
+        return this.setRemoteDescription(msg.sdp).catch(() => {
           log.error('message: Error processing updatestream in connection, connectionId:', this.id);
         });
       }
+    } else if (msg.type === 'offer-dropped') {
+      log.debug('message: Offer dropped, sending again', this.id);
+      return this.sendOffer();
+    } else if (msg.type === 'answer-dropped') {
+      log.debug('message: Answer dropped, sending again', this.id);
```
lodoyun: Consider logging a warning here. Also, sending an answer back could be risky.

erizo_controller/erizoJS/models/Connection.js
```diff
@@ -306,69 +280,143 @@ class Connection extends events.EventEmitter {
...
+    if (msg.type === 'answer') {
+      if (!this.isNegotiationLocked) {
+        log.warn('message: Received answer and negotiation was not locked, connectionId: ', this.id);
+      }
+      const promise = this._onSignalingMessage(msg);
+      this._unlockNegotiation();
+      return promise;
+    }
+    if (msg.type === 'offer-dropped') {
+      this.isNegotiationLocked = false;
```
lodoyun: If there are cases where we only want to unlock the negotiation without executing the next step, I think we could do it explicitly by having two functions `_unlockNegotiation()` and `_executeNextStep()` just like in the client side

lodoyun: Also, are you sure we don't want to enqueue `offer-drop`?

PR #1534, opened 2020-01-14
Skip ice candidate if peer and stream objects are NULL

vpoddubchak:
This PR is a workaround of crash which sometimes happens when iOS client tries to connect to the room. In some cases peer and stream objects are NULL (possible because of race condition) in moment then method nicer->IcePeerContextParseTrickleCandidate is calling. 
In cases when skipped ice-candidate is host - all will be OK: connection will be established. But I'm afraid that in cases when we skip srflx candidate it can bring to connection problems. Anyway it is better than crash of the room...


vpoddubchak:
Tests passed only when I allow using  **peer_** variable only after **startSync()** has been done (synced by start_promise_).
I think this is a fix.

erizo/src/erizo/NicerConnection.cpp
```diff
@@ -320,6 +320,16 @@ void NicerConnection::startGathering() {
 }
 
 bool NicerConnection::setRemoteCandidates(const std::vector<CandidateInfo> &candidates, bool is_bundle) {
+  // At this moment we should have peer_ initalized. To make sure, we are waiting for start_promise_ there
```
jcague: Shouldn't we make sure we don't call setRemoteCandidates before having started Nicer? I don't like the idea of blocking the thread while we are waiting. Another alternative is to return false and call setRemoteCandidates after a while again.

PR #1547, opened 2020-02-25
Trigger connection-failed events when negotiation times out

jcague:
**Description**

We have a new timeout for negotiations that get locked and we trigger a `connection-failed` when it happens.

I updated the client API with the new event.

### DIFF
--- a/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
+++ b/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
@@ -11,10 +11,12 @@ import SdpHelpers from '../utils/SdpHelpers';
 import Logger from '../utils/Logger';
 import FunctionQueue from '../utils/FunctionQueue';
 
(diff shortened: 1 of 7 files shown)


erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
```diff
@@ -14,7 +14,11 @@ import FunctionQueue from '../utils/FunctionQueue';
 const BaseStack = (specInput) => {
   const that = {};
   const specBase = specInput;
-  const negotiationQueue = new FunctionQueue();
+  const negotiationQueue = new FunctionQueue(30000, () => {
```
lodoyun: Consider using a constant somewhere instead of using the `30000` value so it's easier to see what it's for

jcague: good catch!

PR #1549, opened 2020-03-02
Cleanup of SdpInfo and handle ICE Candidates from JS

jcague:
**Description**

I'm cleaning up SdpInfo as a subproject to better handling new SDPs with Unified Plan inside Erizo.

Not needed.


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -504,16 +492,17 @@ boost::future<void> WebRtcConnection::processRemoteSdp(int received_session_vers
...
-  return asyncTask([mid, mLineIndex, sdp] (std::shared_ptr<WebRtcConnection> connection) {
-    connection->addRemoteCandidateSync(mid, mLineIndex, sdp);
+boost::future<void> WebRtcConnection::addRemoteCandidate(std::string mid, int mLineIndex, CandidateInfo candidate) {
+  return asyncTask([mid, mLineIndex, candidate] (std::shared_ptr<WebRtcConnection> connection) {
+    connection->addRemoteCandidateSync(mid, mLineIndex, candidate);
   });
 }
 
-bool WebRtcConnection::addRemoteCandidateSync(std::string mid, int mLineIndex, std::string sdp) {
+bool WebRtcConnection::addRemoteCandidateSync(std::string mid, int mLineIndex, CandidateInfo candidate) {
```
lodoyun: 💯

erizo/src/test/SdpTest.cpp
```diff
@@ -1,253 +0,0 @@
-#include <gmock/gmock.h>
```
lodoyun: 👍 I'm pretty sure we don't need the sdps that we're storing in `src/test/assets` since we're not going to test the parsing right?

erizo/src/erizo/WebRtcConnection.h
```diff
@@ -114,11 +108,6 @@ class WebRtcConnection: public TransportListener, public LogContext, public Hand
    * Copy some SdpInfo data to local SdpInfo
    */
   void copyDataToLocalSdpIndo(std::shared_ptr<SdpInfo> sdp_info);
```
lodoyun: I know this is not from this PR but it may make sense to change this to `copyDataToLocalSdpInfo` as part of the cleaning

erizo/src/test/SdpTest.cpp
jcague: good catch!

erizo/src/erizo/WebRtcConnection.h
jcague: good catch!

PR #1556, opened 2020-03-17
Avoid using publisher from OTM

jcague:
**Description**

There have been other crashes inside OneToManyProcessor when trying to access publisher's stuff.

Not needed.

### DIFF
--- a/erizo/src/erizo/OneToManyProcessor.cpp
+++ b/erizo/src/erizo/OneToManyProcessor.cpp
@@ -25,8 +25,9 @@ namespace erizo {
       return 0;
 
     boost::unique_lock<boost::mutex> lock(monitor_mutex_);
-    if (subscribers.empty())
+    if (subscribers.empty() || !publisher) {
       return 0;
+    }
 
     std::map<std::string, std::shared_ptr<MediaSink>>::iterator it;
     RtpHeader* head = reinterpret_cast<RtpHeader*>(audio_packet->data);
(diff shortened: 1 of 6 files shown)


erizo/src/erizo/OneToManyProcessor.cpp
```diff
@@ -69,13 +70,14 @@ namespace erizo {
       return 0;
     }
     boost::unique_lock<boost::mutex> lock(monitor_mutex_);
-    if (subscribers.empty())
+    if (subscribers_.empty() || !publisher_) {
```
lodoyun: 👍

erizo/src/erizo/OneToManyProcessor.h
```diff
@@ -50,16 +52,20 @@ class OneToManyProcessor : public MediaSink, public FeedbackSink {
...
   int deliverAudioData_(std::shared_ptr<DataPacket> audio_packet) override;
   int deliverVideoData_(std::shared_ptr<DataPacket> video_packet) override;
   int deliverFeedback_(std::shared_ptr<DataPacket> fb_packet) override;
   int deliverEvent_(MediaEventPtr event) override;
   boost::future<void> closeAll();
   bool isSSRCFromAudio(uint32_t ssrc);
   uint32_t translateAndMaybeAdaptForSimulcast(uint32_t orig_ssrc);
+
+ private:
+  typedef std::shared_ptr<MediaSink> sink_ptr;
```
lodoyun: 💯 to the variable names updates. In that line, shouldn't this be `sink_ptr_` ?

erizoAPI/OneToManyProcessor.cc
```diff
@@ -165,7 +165,7 @@ NAN_METHOD(OneToManyProcessor::getPublisherState) {
     return;
   }
 
-  auto wr = std::dynamic_pointer_cast<erizo::MediaStream>(me->publisher);
+  auto wr = std::dynamic_pointer_cast<erizo::MediaStream>(me->getPublisher());
```
lodoyun: 👍

erizo/src/erizo/OneToManyProcessor.h
jcague: actually we can remove it because we don't use it

jcague: It was a typedef instead, but we don't use to define any variable in this class

PR #1558, opened 2020-03-23
User WeakPtr instead of raw pointers in MediaDef

jcague:
**Description**

Introduce Weak pointers to Media Definitions to avoid having wrong access to memory from OneToManyProcessor.

Not needed.


erizo/src/erizo/MediaStream.cpp
```diff
@@ -152,6 +150,16 @@ boost::future<void> MediaStream::close() {
   });
 }
 
+void MediaStream::configure() {
```
lodoyun: Why can't we do this in `MediaStream.init()`? I think semantically it's the same. For both it's things that we should do before using the media stream

jcague: you're right, I've switched their names and now we only call init() from within MediaStream.cc

PR #1564, opened 2020-04-15
Adapt to new nICEr API

vpoddubchak:
After https://github.com/lynckia/nrappkit/pull/4 and https://github.com/lynckia/nICEr/pull/2 are merged, need to update licode to work with changed nICEr api.

This PR is showing how I did it. Not sure it is 100% correct, but it works in our environment and now is on testing stage.

Note: need to change GIT_TAG for project_nicer in erizo/src/third_party/nicer.cmake to new value.


jcague:
This looks promising! thanks a lot for the contribution, why do you want to update nICEr? are there new features?

vpoddubchak:
Mostly because of stability issues: we often see disconnection with "Ica failed" error. For some clients it happens during almost each session.
On our side we still have ability to switch between libnice and nicer, and when we switch to the latest libnice it works more stable. But you decided to go with nICEr, this is why we tried the latest version of it. For now I can say - it works better.
During merging I saw such new features:
 - changed ice-restart logic
 - mDNS support

jcague:
I merged these commits in another PR to fix a couple of cases, thanks for the contribution

PR #1567, opened 2020-04-22
Adapt to new nicer api (work from another's PR)

jcague:
**Description**

I took commits from @vpoddubchak PR and fixed some minor issues. I usually do a PR to the other branch, but I can't do it in this case for some unknown reason, maybe because his repo is private.

Not needed.


jcague:
@vpoddubchak here I follow your work in #1564 and fixed a couple of things. Let's see if it builds and I'll start testing it with our tools. Thanks a lot for the contribution. I've read the original commits and it looks pretty straightforward to implement ICE refresh now so it looks promising!

QuanjieDeng:
This  helpful for me,I get  crash   amostly  every time  at commit  [8a1c...](https://github.com/lynckia/licode/tree/8a1c0ab054415020804819c1952f3c31ece89122) ,Folow the step  up  modify   local  code ,it  work  and  not  happen again!

Issue #1577, opened 2020-05-09
Firefox - SDP configuration, remote  stream subscription failure 


chesky18:
- If publisher (Chrome) is publishing screen or video, firefox is has issue with SDP response configuration. Issueseems to be BaseStack.js, following code does not work:
let direction = Direction.reverse('sendrecv');
    if (numberOfRemoteMedias > 0 && numberOfLocalMedias > 0) {
      direction = Direction.reverse('sendrecv');
    } else if (numberOfRemoteMedias > 0 && numberOfLocalMedias === 0) {
      direction = Direction.reverse('recvonly');
    } else if (numberOfRemoteMedias === 0 && numberOfLocalMedias > 0) {
      direction = Direction.reverse('sendonly');
    } else {
      direction = Direction.reverse('inactive');
    }

After changing the code to following, it seems to work:
let direction = Direction.reverse(Direction.SENDRECV);
    if (numberOfRemoteMedias > 0 && numberOfLocalMedias > 0) {
      direction = Direction.reverse(Direction.SENDRECV);
    } else if (numberOfRemoteMedias > 0 && numberOfLocalMedias === 0) {
      direction = Direction.reverse(Direction.SENDONLY);
    } else if (numberOfRemoteMedias === 0 && numberOfLocalMedias > 0) {
      direction = Direction.reverse(Direction.RECVONLY);
    } else {
      direction = Direction.reverse(Direction.INACTIVE);
    }

david-morales:
I can reproduce the issue also with Safari

PR #1589, opened 2020-05-27
Fix Firefox SDP negotiation and Simulcast

jcague:
**Description**

This PR fixes two issues in Firefox:
- When negotiating SDPs when Erizo sends its Offer first (#1577)
- Simulcast was not working anymore due to an issue in the SDPSemantics.

Not needed.

### DIFF
--- a/erizo_controller/erizoClient/src/webrtc-stacks/FirefoxStack.js
+++ b/erizo_controller/erizoClient/src/webrtc-stacks/FirefoxStack.js
@@ -41,16 +41,12 @@ const FirefoxStack = (specInput) => {
 
   that.enableSimulcast = sdp => sdp;
 
-  const baseCreateOffer = that.createOffer;
-
-  that.createOffer = (isSubscribe, forceOfferToReceive = false, streamId = '') => {
+  that.preCreateOffer = (isSubscribe = false) => {
     let promises = [];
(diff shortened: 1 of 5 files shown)


erizo_controller/erizoClient/src/webrtc-stacks/FirefoxStack.js
```diff
@@ -41,16 +41,12 @@ const FirefoxStack = (specInput) => {
 
   that.enableSimulcast = sdp => sdp;
 
-  const baseCreateOffer = that.createOffer;
-
-  that.createOffer = (isSubscribe, forceOfferToReceive = false, streamId = '') => {
+  that.preCreateOffer = (isSubscribe = false) => {
```
lodoyun, line 44: Good catch! I would suggest a better name though... Maybe `prepareCreateOffer` or `stackPrepareOffer`?

extras/basic_example/public/script.js
```diff
@@ -77,9 +77,7 @@ const startBasicExample = () => {
     video: !audioOnly,
     data: true,
     screen,
-    attributes: {},
-    videoSize: [640, 480, 640, 480],
```
lodoyun, line 81: 👍  This probably makes sense for the basic Example since we run into incompatibilities all the time

PR #1595, opened 2020-06-08
Give a better structure to client logs

jcague:
**Description**

This PR changes the way we structure the client's log lines. I also reduced the level of many of those logs from INFO to DEBUG, so it might affect how people usually track issues in the client. But I hope it will improve the way they detect them with all these changes.

Not needed.

### DIFF
--- a/erizo_controller/erizoClient/src/ErizoConnectionManager.js
+++ b/erizo_controller/erizoClient/src/ErizoConnectionManager.js
@@ -25,7 +25,6 @@ const log = Logger.module('ErizoConnection');
 class ErizoConnection extends EventEmitterConst {
   constructor(specInput, erizoId = undefined) {
     super();
-    log.debug('Building a new Connection');
     this.stack = {};
 
     this.erizoId = erizoId;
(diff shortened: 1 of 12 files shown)


erizo_controller/erizoClient/src/ErizoConnectionManager.js
```diff
@@ -96,8 +96,12 @@ class ErizoConnection extends EventEmitterConst {
     }
   }
 
+  toLog() {
```
lodoyun, line 99: 💯

erizo_controller/erizoClient/src/Room.js
```diff
@@ -53,7 +53,7 @@ const Room = (altIo, altConnectionHelpers, altConnectionManager, specInput) => {
       stream.pc.removeStream(stream);
     }
 
-    log.debug('Removed stream');
+    log.debug(`message: Removed stream, ${stream.toLog()}`);
```
lodoyun, line 56: I'm wondering if a `toLog()` method would be helpful here - logging `that.roomID`, and maybe other Room level state at some point.

PR #1602, opened 2020-06-24
Implement Unified Plan

jcague:
**Description**

This PR adds the concept of Transceivers as the new way to attach a MediaStream to a WebRtcConnection in Erizo. 

It also adds all the remaining stuff to start supporting Unified Plan.

Not needed.


erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -222,46 +239,82 @@ bool WebRtcConnection::werePacketLossesRecently() {
...
+          return transceiver->getMediaStream()->getId() == stream_id;
         }
-        return isStream;
-      }));
+        return false;
     });
+    if (transceiver_it != connection->transceivers_.end()) {
+      std::shared_ptr<MediaStream> stream = (*transceiver_it)->getMediaStream();
+      (*transceiver_it)->resetMediaStream();
+      bool isStream = stream->getId() == stream_id;
```
lodoyun, line 278: Is this a sanity check? Wouldn't we know by now that this stream is the stream we want to remove? Just checking I'm not missing something

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -181,6 +182,22 @@ bool WebRtcConnection::createOfferSync(bool video_enabled, bool audio_enabled, b
...
+    std::string stream_id("");
+    std::string transceiver_id(transceiver->getId());
+    if (!transceiver->isInactive()) {
+      if (transceiver->isSending()) {
+        direction = StreamDirection::SENDONLY;
+      } else {
+        direction = StreamDirection::RECVONLY;
+      }
+      stream_id = transceiver->getMediaStream()->getId();
+    }
```
lodoyun, line 196: Shouldn't we be setting the direction to `StreamDirection::INACTIVE`?  I know we probably don't contemplate that case but nothing prevents us from filling in the data structure correctly.

erizo/src/erizo/WebRtcConnection.cpp
```diff
@@ -222,46 +239,82 @@ bool WebRtcConnection::werePacketLossesRecently() {
...
+        [&media_stream, &added, connection] (const auto &transceiver) {
+      if (!transceiver->hasMediaStream() && !added) {
+        transceiver->setMediaStream(media_stream);
+        added = true;
+        ELOG_DEBUG("%s message: Added mediaStream reusing transceiver, id: %s, transceiverId: $d",
+          connection->toLog(), media_stream->getId().c_str(), transceiver->getId());
+      }
+    });
+    if (!added) {
+      auto transceiver = std::make_shared<Transceiver>(std::to_string(connection->transceivers_.size()));
```
lodoyun, line 255: I think this is the way Chrome sets the ids now. I'm fine with it until we find out if a more sophisticated way to set id's is worth it 👍

erizo/src/erizo/WebRtcConnection.cpp
jcague: yes, StreamDirection::INACTIVE the default value, I set it on line 186.

erizo/src/erizo/WebRtcConnection.cpp
jcague: yes, you're right, this is totally unncessary

PR #1606, opened 2020-06-29
Add more info to logs

jcague:
**Description**

Add more info to logs about metadata and options.

Not needed.

### DIFF
--- a/erizo_controller/erizoJS/erizoJSController.js
+++ b/erizo_controller/erizoJS/erizoJSController.js
@@ -94,11 +94,11 @@ exports.ErizoJSController = (erizoJSId, threadPool, ioThreadPool) => {
     }
   };
 
-  const getOrCreateClient = (erizoControllerId, clientId, singlePC = false) => {
+  const getOrCreateClient = (erizoControllerId, clientId, options = {}, singlePC = false) => {
     let client = clients.get(clientId);
     if (client === undefined) {
       client = new Client(erizoControllerId, erizoJSId, clientId,
-        threadPool, ioThreadPool, !!singlePC);
+        threadPool, ioThreadPool, !!singlePC, options);
       client.on('status_event', onConnectionStatusEvent.bind(this));
       clients.set(clientId, client);
     }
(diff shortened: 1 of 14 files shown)


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -250,7 +252,7 @@ exports.ErizoJSController = (erizoJSId, threadPool, ioThreadPool) => {
     updateUptimeInfo();
     let publisher;
     log.info('addPublisher, clientId', clientId, 'streamId', streamId);
-    const client = getOrCreateClient(erizoControllerId, clientId, options.singlePC);
+    const client = getOrCreateClient(erizoControllerId, clientId, options, options.singlePC);
```
lodoyun, line 255: I assume we do this because we want to have the options passed when adding the publisher. I think, however, that we should still use that options structure and not pass `options.singlePC`. We can do whatever check we need in `getOrCreateClient` to avoid any problems.

erizo_controller/erizoJS/models/Connection.js
```diff
@@ -38,12 +38,12 @@ class Connection extends events.EventEmitter {
     this.mediaConfiguration = 'default';
     //  {id: stream}
     this.mediaStreams = new Map();
+    this.options = options;
+    this.metadata = this.options.metadata || {};
```
lodoyun, line 42: Do we still need this? I've seen this PR uses `this.options.metadata` in general

erizo_controller/erizoJS/models/Subscriber.js
```diff
@@ -88,7 +88,8 @@ class Subscriber extends NodeClass {
   }
 
   close(sendOffer = true) {
-    log.debug(`msg: Closing subscriber, streamId:${this.streamId}`);
+    log.debug(`msg: Closing subscriber, streamId:${this.streamId}, `,
```
lodoyun, line 91:
```suggestion
    log.debug(`message: Closing subscriber, streamId:${this.streamId}, `,
```

PR #1607, opened 2020-06-30
Centralize the way we get configuration flags in BasicExample

lodoyun:
**Description**

This PR cleans up the way we get configuration flags in the BasicExample by creating a new global structure to hold them and parsing all of them when the window is loaded.

### DIFF
--- a/extras/basic_example/public/script.js
+++ b/extras/basic_example/public/script.js
@@ -53,11 +53,13 @@ function startRecording() {
       room.startRecording(localStream, (id) => {
         recording = true;
         recordingId = id;
+        window.recordingId = recordingId;
       });
     } else {
       room.stopRecording(recordingId);
       recording = false;
     }
+    window.recording = recording;
   }
 }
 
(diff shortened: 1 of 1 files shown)


extras/basic_example/public/script.js
```diff
@@ -5,17 +5,39 @@
 /* eslint-disable no-param-reassign, no-console */
 
 const serverUrl = '/';
-window.localStream = undefined;
-window.room = undefined;
-window.recording = false;
-window.recordingId = '';
+let localStream;
+let room;
+let recording = false;
```
jcague, line 10: We are not setting `recording` and `recordingId` to `window` anymore. I think it is good because we don't have a case that needs them as window parameters.

extras/basic_example/public/script.js
```diff
@@ -5,17 +5,39 @@
...
+let recordingId = '';
+const configFlags = {
```
jcague, line 12: 💯

extras/basic_example/public/script.js
lodoyun: 👍  That was my reasoning but I'll add them for consistency

Issue #1609, opened 2020-07-10
DTLS issue in Chrome Canary v86


alokp:
DTLS handshake process appears to be failing when the client is run in latest Chrome Canary v86. Appears to run fine in Chrome Beta and older. 

I tested this by running the official docker container with this command

`PUBLIC_IP=127.0.0.1 MIN_PORT=30000; MAX_PORT=30050; docker run --name licode -p  3000:3000 -p $MIN_PORT-$MAX_PORT:$MIN_PORT-$MAX_PORT/udp -p 3001:3001  -p 8080:8080 -e "MIN_PORT=$MIN_PORT" -e "MAX_PORT=$MAX_PORT" -e "PUBLIC_IP=$PUBLIC_IP" -e "NETWORK_INTERFACE=eth0" lynckia/licode`

basic example connected fine when run from Chrome v83, but when run in Chrome Canary _connectionstatechange_ event was raised with failed. Erizo logs had message about the handshake failing. Here are the server logs from that session

```
2020-07-10 19:39:38.467  - INFO: ErizoController - message: socket connected, socketId: wtawAiSW9HuE8pSvAAAB
2020-07-10 19:39:38.482  - DEBUG: ErizoController - Channel - message: token received
2020-07-10 19:39:38.484  - INFO: RPCPublic - message: consumed token, tokenId: 5f08c3f832a28302249dd44a, roomId: 5f08c37332a28302249dd448, serviceId: 5f08c366beb2a2d3c5563b26
2020-07-10 19:39:38.485  - INFO: ErizoController - message: getting singlePC configuration, singlePc: false, global: 
2020-07-10 19:39:38.485  - DEBUG: ErizoController - Client - message: Adding listeners to socket events, client.id: 92423243-bc3e-4b79-97e8-4ad9ef7f85e2
2020-07-10 19:39:38.485  - INFO: ErizoController - message: client connected, clientId: 92423243-bc3e-4b79-97e8-4ad9ef7f85e2, singlePC: false
[...]

PR #1610, opened 2020-07-11
Update x509 cert version to 3

jcague:
**Description**

This should fix issues with latest versions of Chrome (v86+), like the one mention in #1609 .

Not needed.


erizo/src/erizo/dtls/DtlsClient.cpp
```diff
@@ -210,7 +210,7 @@ int createCert(const std::string& pAor, int expireDays, int keyLen, X509*& outCe
   X509_EXTENSION* ext = X509_EXTENSION_new();
 
   // set version to X509v3 (starts from 0)
-  // X509_set_version(cert, 0L);
+  X509_set_version(cert, 2L);
```
lodoyun, line 213: 👍

PR #1612, opened 2020-07-14
New Client API to update active simulcast layers for publishers

jcague:
**Description**

New Client API to update active simulcast layers for publishers.

We can now decide which Simulcast layers are active. It can only be applied to publishers.
```
localStream.updateSimulcastActiveLayers({0: true, 1: false});
```

In this example we are configuring 2 spatial layers bitrates, limiting the lower layer to 80 Kbps and the higher to 430 Kbps.


doc/client_api.md
```diff
@@ -321,6 +322,15 @@ localStream.updateSimulcastLayersBitrate({0: 80000, 1: 430000});
...
 In this example we are configuring 2 spatial layers bitrates, limiting the lower layer to 80 Kbps and the higher to 430 Kbps.
 
+## Update Simulcast Active Layers
+
+We can decide which Simulcast layers are active. It can only be applied to publishers.
+```
+localStream.updateSimulcastActiveLayers({0: true, 1: false});
+```
+
+In this example we are configuring 2 spatial layers bitrates, limiting the lower layer to 80 Kbps and the higher to 430 Kbps.
```
lodoyun, line 332:
```suggestion
In this example we are disabling layer 1 while the other layer (0) is active.
```

PR #1619, opened 2020-07-23
Refactor websocket reconnections to ack messages

jcague:
**Description**

The current mechanism to avoid losing messages in ErizoController was not 100% reliable. We found issues when we were not able to detect a reconnection. This new approach sends ACKs for each signaling message to better recovering from not detected reconnections too.

Not needed.


erizo_controller/erizoController/erizoController.js
```diff
@@ -102,13 +102,17 @@ Object.keys(opt.options).forEach((prop) => {
   }
 });
 
+const SOCKET_IO_PING_INTERVAL = 10000;
```
lodoyun, line 105: I'm curious about having this constants here. If they are not used anywhere else I think it's fine just to pass the values in the `socket.io` `listen` call. Otherwise, if we think it's useful, I would export them to the config file.

lodoyun (approved):
This looks good to me 👍

erizo_controller/erizoController/erizoController.js
jcague: This is an unusual file because we do lots of different things before starting with the main logic. Some constants are defined even in some lines below. IMO there is no reason to export them as config variables so I decided to put all the constants together in the same place to make it the code a bit more comprehensive.

PR #1626, opened 2020-08-19
Fix problems with logs

lodoyun:
**Description**
This PR fixes a couple of problems introduced recently in logs.
1. Fix the _Connecting to Room_ log. It now properly logs the token and we remove the RoomId since it is never present and it can be confusing.
2. Logs would produce an error if a stream does not have attributes


jcague (approved):
good catch!

PR #1629, opened 2020-09-04
Add first negotiation tests

jcague:
**Description**

Add first SDP negotiation tests.

Not needed.


test/package.json
```diff
@@ -4,8 +4,12 @@
   "private": true,
   "dependencies": {
     "aws-sdk": "^2.102.0",
+    "axios": "^0.20.0",
     "chai": "^3.5.0",
+    "cli-progress": "^3.8.2",
+    "csv-parse": "^4.12.0",
```
lodoyun, line 10: I don't see where we're using this. Do we need it?

test/negotiation/log4cxx.properties
```diff
@@ -0,0 +1,78 @@
+# Set root logger level to WARN and its only appender to A1.
+log4j.rootLogger=WARN, A1
```
lodoyun, line 2: 💯  to having log level files specific for these tests

test/negotiation/utils/NegotiationTest.js
```diff
@@ -0,0 +1,473 @@
...
+        ctx.checkClientHasNotFailed();
+      });
+    };
+
+    ctx.subscribeToErizoStreamStep = function() {
+      describe('Subscribe to One Erizo Stream', function() {
+        let offer, answer;
+
+        before(async function() {
+          remoteStream = await ctx.createRemoteStream();
```
lodoyun, line 338: I think `remoteStream` in this context is a bit confusing. I think maybe `erizoStream` would be more descriptive since it's a stream that originates in Erizo in the context of the negotiation.

test/negotiation/utils/NegotiationTest.js
```diff
@@ -0,0 +1,473 @@
...
+    };
+
+    after(async function() {
+      await page.close();
+      ctx.erizo.removeAllStreams();
+      ctx.erizo.close();
+    });
+
+    before(async function() {
+      const htmlPath = '/Users/jcague/development/licode/extras/basic_example/public/index.html';
```
lodoyun, line 167: Consider making this dependent of the path of the process, or a similar thing to avoid having to update the path here.

test/negotiation/utils/NegotiationTest.js
```diff
@@ -0,0 +1,473 @@
...
+function sleep(ms) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
+
+let browserInstalled = false;
+let revisionInfo;
+
+const installBrowser = async () => {
```
lodoyun, line 24: I'm not sure about having this here. I would probably separate it and have it be a prerequisite for running the tests.  That script could, for instance, take environment variables as input for the chromeVersion and/or the platform

lodoyun (approved):
This looks very good to me. I think it's going to be a very important tool.

test/package.json
jcague: good catch!

test/negotiation/utils/NegotiationTest.js
jcague: yes, you're right that it will be more descriptive 👍

test/negotiation/utils/NegotiationTest.js
jcague: good catch!

test/negotiation/utils/NegotiationTest.js
jcague: I moved this code to a new `BrowserInstaller.js` file

PR #1632, opened 2020-09-15
Remove FuturesManager

lodoyun:
**Description**

This PR removes `FuturesManager` since we are using a different mechanism to clean up the futures.


jcague (approved):
Good catch!

PR #1636, opened 2020-09-22
Fix memory leaks and optimize code

jcague:
**Description**

I'm currently investigating cases of unsubscriptions that finish with a timeout between ErizoController and ErizoJS. While investigating those cases I found out some minor leaks and tried to optimize some logic paths.

I think it is important to look into what I'm doing when closing the WebRtcConnection because we now use a promise to do it. I've tested it locally and I'm considering testing it in a dev environment with multiple browsers.

I'm also thinking of implementing async/await instead of chaining promises, but I'll do it in a different PR.

Not needed.

### DIFF
--- a/erizoAPI/WebRtcConnection.cc
+++ b/erizoAPI/WebRtcConnection.cc
@@ -85,7 +85,7 @@ void WebRtcConnection::closeEvents() {
   }
   async_ = nullptr;
   future_async_ = nullptr;
-  ELOG_DEBUG("%s, message: Closed, pendingRefs: %d", toLog(), refs_);
+  ELOG_DEBUG("%s, message: Closed Events, pendingRefs: %d", toLog(), refs_);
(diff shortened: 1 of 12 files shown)


erizoAPI/WebRtcConnection.cc
```diff
@@ -98,8 +85,29 @@ void WebRtcConnection::close() {
   }
   async_ = nullptr;
   future_async_ = nullptr;
+  ELOG_DEBUG("%s, message: Closed, pendingRefs: %d", toLog(), refs_);
```
lodoyun, line 88: Consider saying specifically that we closed the events `message: ClosedEvents`.

lodoyun (approved):
This looks good to me 👍 . I think this addresses most (if not all) the cases I've encountered and I like the solutions proposed.

PR #1638, opened 2020-09-30
Update remove client in erizoJS

lodoyun:
**Description**

This PR changes the way `removeClient` works in erizoJS. 
Before, when a Client left, erizoController would issue the commands to erizoJS to remove all published streams and subscriptions from that particular Client and then it would request the removal of the client itself. This approach lent itself to a number of race conditions.
With this PR, erizoController will only have to request the removal of the Client and only remove locally the data structures associated with that client (published and subscribed streams). ErizoJS will now process this removeClient as a command to remove the subscriptions and the published streams in the same operation.

The new `NodeManager` makes it easier for erizoJS to find published streams owned by a Client.

### DIFF
--- a/erizo_controller/erizoController/erizoController.js
+++ b/erizo_controller/erizoController/erizoController.js
@@ -418,10 +418,13 @@ exports.deleteRoom = (roomId, callback) => {
     return;
   }
 
+  // delete all clients and their streams
+  room.forEachClient((client) => {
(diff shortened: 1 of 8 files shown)


jcague (approved):
LGTM! The only nit-picking comment is about why we name NodeManager instead of PublisherManager, which I think would be less verbose if we had methods like add(), remove(), etc.

PR #1641, opened 2020-10-05
Fix some ICE connectivity issues

jcague:
**Description**

I've been investigating some issues that led to having different ICE states between browser and Erizo in some cases. Some of the issues were solved here, while others need to be fixed in nICEr (I'm also preparing a PR there).

The major issue was that Erizo considered all Host candidates as failed because they were using their private IP network, even if we define a Public IP one. For those cases, we now automatically add the Public IP as the Host candidates' address when it is set in the config file.

As part of this work, I've been experimenting a bit with ICE-LITE. I haven't completed it yet because it is low priority, as Firefox does not seem to support ICE-LITE MCUs. The good news is that I've solved some issues that I found during the process.

Not needed.

### DIFF
--- a/erizo/src/erizo/NicerConnection.cpp
+++ b/erizo/src/erizo/NicerConnection.cpp
@@ -112,7 +112,7 @@ int NicerConnection::ice_connected(void *obj, nr_ice_peer_ctx *pctx) {
   }
   conn->updateIceState(IceState::READY);
   conn->nicer_->IceContextFinalize(conn->ctx_, pctx);
-
+  conn->nicer_->IcePeerContextDumpState(pctx, 0);
   return 0;
 }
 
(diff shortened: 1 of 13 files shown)


erizo/src/erizo/NicerConnection.cpp
```diff
@@ -112,7 +112,7 @@ int NicerConnection::ice_connected(void *obj, nr_ice_peer_ctx *pctx) {
   }
   conn->updateIceState(IceState::READY);
   conn->nicer_->IceContextFinalize(conn->ctx_, pctx);
-
+  conn->nicer_->IcePeerContextDumpState(pctx, 0);
```
lodoyun, line 115: 👍

lodoyun (approved):
This looks good to me 👍

PR #1642, opened 2020-10-08
Add stats about the delay added to running tasks in Workers

jcague:
**Description**
We wanted to add even more information about the tasks that are being run within Erizo's Workers. I also added more data about the time it takes for these tasks to be completed.

Not needed.


lodoyun (approved):
This looks good to me 👍

PR #1643, opened 2020-10-09
Add more info to StreamFailed events

jcague:
**Description**

This PR adds more info about the connection when a Stream failed because the ICE connection failed.

Not needed.


erizo_controller/erizoClient/src/ErizoConnectionManager.js
```diff
@@ -36,6 +36,7 @@ class ErizoConnection extends EventEmitterConst {
     this.sessionId = ErizoSessionId;
     this.connectionId = spec.connectionId;
     this.qualityLevel = QUALITY_LEVEL_GOOD;
+    this.reachedConnectivity = false;
```
lodoyun, line 39: I'm wondering if `reachedConnectivity` is the best way to call this. I don't know... `wasAbleToConnect`, `wasPreviouslyConnected` seem clearer to me but it's your call :)

lodoyun (approved):
This looks good to me. Just a name suggestion but it's not really important. 👍

PR #1644, opened 2020-10-09
Add metrics about promises duration and delay

jcague:
**Description**
Add metrics about promises duration and delay to see how long these promises last and if there is some delay when they are resolved.

Not needed.


erizoAPI/MediaStream.cc
```diff
@@ -117,6 +117,13 @@ std::string MediaStream::toLog() {
   return "id: " + id_;
 }
 
+void MediaStream::computePromiseTimes(erizo::time_point scheduled_at,
+    erizo::time_point started, erizo::time_point end) {
+  promise_durations_.add(started - scheduled_at);
```
lodoyun, line 122: I think this is inverted compared to what we do with tasks. Shouldn't the  `delay = started - scheduled_at`, meaning the delay before the promise starts and  `duration = end - started` meaning the duration of the actual promise?

lodoyun (approved):
This looks good to me. Just a question regarding `duration` vs `delay` in the comments. 👍

erizoAPI/MediaStream.cc
jcague: It is not inverted, but that's true that I should use a better naming instead of started/scheduled_at/ended, and I'll update the description for the Prometheus metrics.

PR #1646, opened 2020-10-16
Two minor fixes in metrics and tests

jcague:
**Description**

This is a small PR to implement two small fixes:
- We were not populating Prometheus metrics about promises when there were no connections/streams.
- We were having some warnings when running tests.

Not needed.

### DIFF
--- a/erizo_controller/erizoJS/erizoJSController.js
+++ b/erizo_controller/erizoJS/erizoJSController.js
@@ -675,13 +675,7 @@ exports.ErizoJSController = (erizoJSId, threadPool, ioThreadPool) => {
   };
 
   const sumArrays = (array1, array2) => {
-    if (array1.length === 0) {
-      return array2;
-    }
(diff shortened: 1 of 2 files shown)


erizo_controller/erizoJS/erizoJSController.js
```diff
@@ -738,6 +732,7 @@ exports.ErizoJSController = (erizoJSId, threadPool, ioThreadPool) => {
       });
     });
     initMetrics();
+    console.log(metrics);
```
lodoyun, line 735: Is this for your testing?

lodoyun (approved):
This looks good to me  👍

erizo_controller/erizoJS/erizoJSController.js
jcague: this was for local testing, I should remove it, thanks!

PR #1649, opened 2020-10-19
Https by default in Licode and BasicExample

davidguv:
**Description**

Changes the default configuration to https to avoid problems with the browsers.

Since browsers require https to access the camera and microphone. It is mandatory to change the default configuration of Licode in order to test it. 

https://blog.mozilla.org/webrtc/camera-microphone-require-https-in-firefox-68/
https://www.chromestatus.com/feature/5703419427815424
https://www.digicert.com/blog/https-only-features-in-browsers/
https://caniuse.com/mdn-api_mediadevices_getusermedia_secure_context_required

[x] It includes documentation for these changes in `/doc`.

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -11,10 +11,13 @@ Both options require that you have [docker](https://docs.docker.com/installation
 
 The easiest way to run licode is to use the [image we provide](https://hub.docker.com/r/lynckia/licode/) in Docker Hub. In this case you have only to execute the run command. But now the image name is lynckia/licode:*version* where `version` is the release you want to use:
 
-	PUBLIC_IP=<YOUR_ACTUAL_IP> MIN_PORT=30000; MAX_PORT=30050; sudo docker run --name licode -p  3000:3000 -p $MIN_PORT-$MAX_PORT:$MIN_PORT-$MAX_PORT/udp -p 3001:3001  -p 8080:8080 -e "MIN_PORT=$MIN_PORT" -e "MAX_PORT=$MAX_PORT" -e "PUBLIC_IP=$PUBLIC_IP" -e "NETWORK_INTERFACE=eth0" lynckia/licode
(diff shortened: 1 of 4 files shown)


aalonsog:
Taking into account that browsers impose the use of https for getting the user media, I agree with this approach.

lodoyun:
You don't have to change the configuration to test it locally (The https/wss requirement not enforced for localhost), which is the main idea of the docker container and the focus of the configuration.

I'm aware that this is not a solution that works in every scenario, neither is the one proposed here since it uses the default certificates that you have to make sure to accept for both wss (erizoController) and https (basicExample). Accordingly, the reasoning for this being the way it is was that, with no https ,you can test it locally as it is and, if you want to deploy it in production, you should get in there and configure everything the way you want (listen https or not, use a proxy, properly open the ports, etc.).

davidguv:
Thanks for your comment!

I tried to use lycode with a docker but found these problems:

- Docker Licode with Firefox on localhost/127.0.0.1 does not work for me.
WebRTC: ICE failed, add a TURN server and see about:webrtc for more details

- Chrome work good with localhost/127.0.0.1

- In [licode-docker-doc](https://github.com/lynckia/licode/blob/master/doc/docker.md) talks about <ACTUAL_IP>, anything other than 127.0.0.1 would not work. Maybe is better change to 127.0.0.1 in the doc?

lodoyun:
we could default to 127.0.0.1 in the start script if the env variable is not present, that's probably a good solution.
I just double checked and run the last version of the container and had no problems with Chrome or Firefox. There could be something else that is causing problems for you.

davidguv:
Any ip other than 127.0.0.1 would not work for http. 
Do you mean that another ip other than 127.0.0.1 starts in https?

harikt:
I think using env will solve the problem . I send a PR https://github.com/lynckia/licode/pull/1651 which may help and close this issue also.

lodoyun:
I'm not convinced about  #1651 either. Again, the default start script for the container has a very clear goal, it's a way for people to quickly set up a licode server locally and play with it.
If you are going to be setting it up for production or in a _staging_ environment you are going to have to change a number of things and passing all of them through environment variables is not the best way to do it.
IMO, a more sensible approach is to use a [different entry_point script ](https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime) for the container. You can start by using the one we have and modify it to use a different config file, make sure the components start in the right order, etc.

davidguv:
Perhaps the improvement would be to include in the documentation that the dockerfile and default configuration is exclusively for testing on localhost. 

https://licode.readthedocs.io/en/master/docker/

Would other files (Dockerfile, licode_default.js, initDockerLicode.sh) be interesting in this repository for connection from different machines?

PR #1651, opened 2020-10-29
update ssl and port for docker in erizo

harikt:
Hello,

Apologies. I am new to docker, licode and this is my first patch thinking this will work for everyone. Hoping for the best :) .

When playing with docker, the websocket was always using the IP addresss and port 8080 without ssl . So licode didnot worked on real server. In local system things may be different.

So currently added env value which will be replaced.

No

### DIFF
--- a/extras/docker/initDockerLicode.sh
+++ b/extras/docker/initDockerLicode.sh
@@ -189,6 +189,9 @@ if [ "$ERIZOAGENT" == "true" ]; then
   if [[ ! -z "$ERIZOCONTROLLER_SSL" ]]; then
     sed -i "s/config\.erizoController\.ssl = .*/config\.erizoController\.ssl = $ERIZOCONTROLLER_SSL;/" /opt/licode/licode_config.js
   fi
+  if [[ ! -z "$ERIZOCONTROLLER_HOSTNAME" ]]; then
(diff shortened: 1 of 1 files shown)


harikt:
How this works :  from https://licode.readthedocs.io/en/master/docker/

```
PUBLIC_IP=<YOUR_ACTUAL_IP> MIN_PORT=30000; MAX_PORT=30050; sudo docker run --name licode -p  3000:3000 -p $MIN_PORT-$MAX_PORT:$MIN_PORT-$MAX_PORT/udp -p 3001:3001  -p 8080:8080 -e "MIN_PORT=$MIN_PORT" -e "MAX_PORT=$MAX_PORT" -e "PUBLIC_IP=$PUBLIC_IP" -e "NETWORK_INTERFACE=eth0"  -e "ERIZOCONTROLLER_PORT=$ERIZOCONTROLLER_PORT" -e "ERIZOCONTROLLER_SSL=$ERIZOCONTROLLER_SSL" -e "ERIZOCONTROLLER_HOSTNAME=$ERIZOCONTROLLER_HOSTNAME" lynckia/licode
```

Probably the env values can be reduced to `CTRL_*` if preferred.

harikt:
Based on reply on https://github.com/lynckia/licode/pull/1649#issuecomment-719507311 

> IMO, a more sensible approach is to use a [different entry_point script](https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime) for the container. You can start by using the one we have and modify it to use a different config file, make sure the components start in the right order, etc.

PR #1656, opened 2020-11-12
Enable basic features by default in basic example

jcague:
**Description**

This PR enables Simulcast, Single Peer Connection, and Offers From Erizo by default when using the basic example. They can be disabled manually with the corresponding URL queries set to 0. For example, if we want to disable Simulcast we would use this URL: `https://<licode_url>/?simulcast=0`.

I'm also adding some delay after starting RabbitMQ because it was failing to me almost 100% of the time.

Not needed.


lodoyun (approved):
This looks good to me 👍

PR #1665, opened 2020-12-02
Fix a crash in depacketizer that happened when we reset it

ferhez:
after calling reset last_payload_ is null_ptr, so it is crashing.

**Description**


jcague (approved):
good catch!

PR #1669, opened 2020-12-23
Fix Firefox onnnegotiation needed

lodoyun:
**Description**

This PR fixes an issue caused by Firefox behaving differently than Chrome when issuing `onnegotiationneeded`. Chrome calls it once per track in the stream while Firefox seems to call it only once per stream. This was causing issues in all Firefox connections.


jcague (approved):
good catch!

PR #1675, opened 2021-01-18
Add Performance Metrics about ErizoJS to Prometheus

jcague:
**Description**

Add perf metrics about ErizoJS (requests to subscribe and unsubscribe streams) and export them to Prometheus through ROV.

This PR adds the structure needed to seamlessly add new metrics about the duration of tasks performed in ErizoJS and I add use it to calculate the performance of subscriptions/unsubscriptions.

Not needed.


lodoyun (approved):
This looks good to me 👍  I like the use of both `perf_hooks` and `fast-stats` 👏

PR #1677, opened 2021-01-19
Fix new url format for mongodb

lodoyun:
**Description**
This PR updates the default config file and the nuve installation script to adapt to the new format for the mongodb url.


jcague (approved):
LGTM! good catch!

PR #1684, opened 2021-03-04
Updated mongo and conan versions

aalonsog:
The original objective of this PR was to update the MongoDB version installed by Licode scripts to MongoDB v4. This is needed for supporting the new implementation of transactions included in commit a9a29f8cbf4ef0b1940f349879fdff19bfec4d69.

However, during the update I discovered an Erizo installation issue raised by the incompatibility of the current versions of bzip2 and conan. Thus, I decided to update also the conan version to 1.34. This update implicates the modification of how the conan compiler is passed to the *format_include_paths* function. This is why I've also modified conanfile.py.

Summarising the PR modifies the following: 

- Updated MongoDB version installed by installUbuntuDeps.sh to MongoDB 4 and changed installation for using binaries (installation using package manager crashed the docker image compilation due to systemd requirement).
- Updated conan version installed by installUbuntuDeps.sh to conan 1.34
- Updated conanfile.py for being compatible with the new conan version


scripts/installUbuntuDeps.sh
```diff
@@ -94,16 +94,29 @@ install_apt_deps(){
...
   sudo chown -R `whoami` ~/.npm ~/tmp/ || true
 }
 
+install_mongodb(){
+  if [ -d $LIB_DIR ]; then
+    echo "Installing mongodb-org from tar"
+    sudo apt-get install -y libcurl4 openssl liblzma5
+    wget -P $LIB_DIR https://fastdl.mongodb.org/linux/mongodb-linux-x86_64-ubuntu2004-4.4.4.tgz
+    tar -zxvf $LIB_DIR/mongodb-linux-x86_64-ubuntu2004-4.4.4.tgz -C $LIB_DIR
+    sudo cp $LIB_DIR/mongodb-linux-x86_64-ubuntu2004-4.4.4/bin/* /usr/local/bin/
```
lodoyun, line 111: would symbolic links work? I don't particularly like copying the binaries there when they may change.
Also, this may create new problems with anyone that already has a mongo version installed. We could consider using a suffix or a preffix for the binaries. For every other thing I think we either use a ubuntu package or we keep it in `LIB_DIR` without affecting `/usr/bin` or `/usr/lib`

lodoyun: I would also add a way to clean everything up in the `cleanup` function

lodoyun (commented):
This looks good. I do have a couple of suggestions to make the new mongo installation less problematic.

scripts/installUbuntuDeps.sh
aalonsog: Done!! Thanks for the suggestion :)

jcague (approved):
LGTM!

PR #1687, opened 2021-03-17
Fix linking with bzip and zlib2 when building erizo

lodoyun:
**Description**

This PR addresses issues I experienced when linking against some boost dependencies: bzip2 and zlib.
We no longer need to use  `@conan/stable` for things that are in the Conan Index.

Additionally, I remove the conan brew installation for mac, that is not needed anymore.


jcague (approved):
LGTM!

PR #1690, opened 2021-03-26
Handler API

Arri98:
**Description**

In this pull request the possibility of adding handlers to the pipeline without modifying the mediaStream file is introduced. The objective is to facilitate the processing of the audio and video streams inside the pipeline. To desing these new handlers they have to extend the CustomHandler interface created inside the Handler.h file, that has the same functions that the Handler class has plus a new function getPostion(), that returns the position where the handler has to be inserted inside the pipeline. This position is seleted by the handler designer and can not be changed. Handlers can be inserted into the pipeline into three different positions:

- Beginning: At the start of the pipeline, before the rest of handlers.
- Middle: In the middle of the pipeline.
- End: After all the handlers.

These handlers are grouped into profiles that can be imported and selected through the API as documented in `/doc`.

To add handlers to the API they have to be copied to the erizo/src/erizo/handlers folder and add the name of the class and the name you want for them inside the API into the erizo/src/erizo/handlers/handler.py file. Then erizo has to be recompiled. Handlers are imported by the erizo/src/erizo/handler/HandlerImporter class which creates and stores pointer to the handlers. When running the ./installErizo script a python script modifies the HandlerImporter class to add the handlers selected in the handler.py file to the API. Once this is done the new handlers are available and can be selected by creating a series of profiles that can be later selected throught the API. This proccess of adding handlers and creating profiles is detailed in the doc file.

A test handler has been added which logs data from the packets received, as an example for handler development.

Added the possibility of adding handler to the pipeline by creating profiles in the licode.config file and the selecting them through the API.

[Yes] It includes documentation for these changes in `/doc`.

### How does it affect Erizo Client?
Erizo Client reads the new handlerProfile parameter, which contains the selected handler profile for the mediaStream, and sends it to the erizoJS 

### How does it affect Nuve?
It does not affect Nuve

### How does it affect Erizo Controller / Erizo Agent / Erizo JS?
When creating a new webRtcMediaStream ErizoJS controller reads the config file to get the name of the handlers associated to the selected profile.

### How does it affect ErizoAPI / Erizo?
The mediaStream recieves a new parameter, maps where each pair represents a filter and contains its name and parameters sended by the user. It has a default empty value for retrocompatibility.  Also when adding the handlers to the pipeline, in the three points specified before, a function calls pipeline->add_front function for all the handlers that have to be inserted in that position.
Erizo API has a new node addon that receives a vector of maps, containing the filter names and parameters and creates the handlers and inserts then in a  map that we then pass to the mediaStream.


doc/client_api.md
```diff
@@ -331,6 +331,26 @@ localStream.updateSimulcastActiveLayers({0: true, 1: false});
 
 In this example we are disabling layer 1 while the other layer (0) is active.
 
+
+## Add handlers to the mediaStream
```
jcague, line 335: Part of this documentation is outdated. Also, I'd rather move it to `server_api.md`.

erizo/src/erizo/MediaStream.cpp
```diff
@@ -447,7 +451,7 @@ void MediaStream::initializePipeline() {
   pipeline_->addService(packet_buffer_);
 
   pipeline_->addFront(std::make_shared<PacketReader>(this));
-
+  addHandlerInPosition(BEGINNING, handler_pointer_dic, handler_order);
```
jcague, line 454: I like these three options. An alternative naming would be TOP, MIDDLE, and BOTTOM, which might give a more accurate description of where the handler will be placed in the "network" stack.

erizo/src/erizo/handlers/HandlerImporter.h
```diff
@@ -0,0 +1,41 @@
...
+};
+
+
+class HandlerImporter: public HandlerImporterInterface {
+    DECLARE_LOGGER();
+ public:
+    HandlerImporter();
+    void loadHandlers(std::vector<std::map<std::string, std::string>> customHandlers);
+ private:
+    enum HandlersEnum {LoggerHandlerEnum};
```
jcague, line 34: To make this even easier for developers, this Enum could become a vector and have an initial step where the new Handlers should self-register to that vector. It could work if the HandlerImporter is a Singleton, for instance.

But we can do it in a different PR.

erizo/src/erizo/handlers/handlers/LoggerHandler.h
```diff
@@ -0,0 +1,29 @@
...
+    void write(Context *ctx, std::shared_ptr <DataPacket> packet) override;  // Process packet sent to client
+    Positions position() override;  // Returns position to place handler.
+    void enable() override;  // Enable handler
+    void disable() override;  // Disable handler
+    std::string getName() override;  // Returns handler name
+    void notifyUpdate() override;  // Recieves update
+ private:
+    RtcpHeader* rtcp_head;
+    RtpHeader* rtp_head;
+    bool isEnabled = true;
```
jcague, line 25: We prefer `snake_case` vs `camelCase` in the C++ part of Licode. We still see some camel cases there, but those are still inherited from the past.

erizo/src/erizo/handlers/HandlerImporter.h
```diff
@@ -0,0 +1,41 @@
...
+    std::map<std::string, std::shared_ptr<erizo::CustomHandler>> handlers_pointer_dic = {};
+    std::vector<std::string> handler_order = {};
+};
+
+
+class HandlerImporter: public HandlerImporterInterface {
+    DECLARE_LOGGER();
+ public:
+    HandlerImporter();
+    void loadHandlers(std::vector<std::map<std::string, std::string>> customHandlers);
```
jcague, line 32: consider using `custom_handlers` here

erizo/src/erizo/handlers/HandlerImporter.h
```diff
@@ -0,0 +1,41 @@
...
+ private:
+    enum HandlersEnum {LoggerHandlerEnum};
+
+    std::map<std::string, HandlersEnum> handlersDic ={{"LoggerHandler", LoggerHandlerEnum}};
```
jcague, line 36: consider using `handlers_dic` here

erizo/src/erizo/handlers/handlers/LoggerHandler.cpp
```diff
@@ -0,0 +1,92 @@
...
+            } else if (packet->priority == LOW_PRIORITY) {
+                ELOG_DEBUG("Packet with low priority");
+            }
+            ELOG_DEBUG("------------------------");
+            ctx->fireRead(std::move(packet));  // Gives packet ownership to next handler in pipeline
+        }
+    }
+
+    void LoggerHandler::enable() {
+    isEnabled = true;
```
jcague, line 79: correct indentation

erizo/src/erizo/handlers/handlers/LoggerHandler.cpp
```diff
@@ -0,0 +1,92 @@
+
+#include "LoggerHandler.h"
+
+#include "../../rtp/RtpHeaders.h"  // Imports headers utilities
+
+namespace erizo {
+
+    DEFINE_LOGGER(LoggerHandler, "rtp.LoggerHandler");  // Defines handler logger name
```
jcague, line 8: Indentation: we prefer 2 spaces

erizo/src/erizo/handlers/HandlerImporter.cpp
```diff
@@ -0,0 +1,28 @@
...
+// Created by licode20 on 22/3/21.
+//
+
+#include "HandlerImporter.h"
+namespace erizo {
+
+    HandlerImporter::HandlerImporter() {}
+
+    void HandlerImporter::loadHandlers(std::vector<std::map<std::string, std::string>> customHandlers) {
+        for (unsigned int i = 0; i < customHandlers.size(); i++) {
```
jcague, line 11: there are multiple names in camelCase, but we prefer snake_case in C++

erizo/src/erizo/MediaStream.cpp
```diff
@@ -1040,6 +1045,18 @@ void MediaStream::enableSlideShowBelowSpatialLayer(bool enabled, int spatial_lay
   });
 }
 
+void MediaStream::addHandlerInPosition(Positions position,
+       std::map<std::string, std::shared_ptr<erizo::CustomHandler>> handlers_pointer_dic,
+       std::vector<std::string> handler_order) {
+    for (unsigned int i = 0; i < handler_order.size() ; i++) {
+        std::string handlerName = handler_order[i];
+        if (handlers_pointer_dic.at(handlerName) && handlers_pointer_dic.at(handlerName)->position() == position) {
```
jcague, line 1053: consider the handler_name (snake_case) instead of handlerName

jcague (approved):
I don't see any issue with this code. I left some minor comments about styling and things the linter did not found. But it looks good to me, thanks for the contribution @Arri98 !

erizo/src/erizo/handlers/HandlerImporter.cpp
```diff
@@ -0,0 +1,28 @@
+//
```
lodoyun, line 1: nit: We don't use this comment anywhere else in the code. While we are not 100% consistent I wouldn't add a new thing

erizo/src/erizo/MediaStream.cpp
lodoyun, line 454: I agree with the other reviewer :). I think we need better names for this. I propose a different way which would be more self-explanatory to me: `AFTER_READER`, `MIDDLE`, `BEFORE_WRITER`. Wdyt?

lodoyun (approved):
I don't have that much to add seeing the other review and I don't want to repeat all the points, just a couple of comments 🙂 . I do think we could experiment in the future to find ways (if possible) to reduce even further the amount of code from licode itself that the handler developers have to modify but this is a good solution 👍

erizo/src/erizo/MediaStream.cpp
Arri98: Ok, great sugestion. I was also looking for a better name but couldn't find anything that clear.

doc/client_api.md
Arri98: Changed

erizo/src/erizo/handlers/HandlerImporter.h
Arri98: Changed all camelCase

Arri98:
I jsut changed the names that you suggested and fixed the indentation aswell as the documentation

PR #1692, opened 2021-04-09
[WIP] ICE-TCP support. Another attempt.

vpoddubchak:
My attempt to make ICE-TCP work.
I took 2 commits from jcague's PR and fixed some problems.
As result: Chrome is sending data to the licode when UDP protocol i sblocked on client side.

But I still have problems with receiving stream from licode. 
Any help is appreciated.

### DIFF
--- a/erizo/src/erizo/NicerConnection.cpp
+++ b/erizo/src/erizo/NicerConnection.cpp
@@ -461,7 +461,19 @@ void NicerConnection::onCandidate(nr_ice_media_stream *stream, int component_id,
     default:
       break;
   }
-  cand_info.netProtocol = "udp";
+  cand_info.netProtocol = cand->addr.protocol == IPPROTO_UDP ? "udp" : "tcp";
+  if (cand->addr.protocol == IPPROTO_TCP) {
+    switch (cand->tcp_type) {
+      case nr_socket_tcp_type::TCP_TYPE_ACTIVE:
+        cand_info.tcpType = TCP_ACTIVE;
+        break;
+      case nr_socket_tcp_type::TCP_TYPE_PASSIVE:
+        cand_info.tcpType = TCP_PASSIVE;
+        break;
+      default:
+        cand_info.tcpType = TCP_SO;
+    }
+  }
   cand_info.transProtocol = ice_config_.transport_name;
   cand_info.username = ufrag_;
   cand_info.password = upass_;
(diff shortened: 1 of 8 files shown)


vpoddubchak:
@jcague can you please give me advice where to move with it ?

vpoddubchak:
Hi @jcague.  ICE-TCP support is becoming a burning issue for us.  We have a customer with a crazy firewall and all UDP connections from licode will randomly drop after a while.  We see a few ICE-RESTARTS but after a while, licode stops trying.  We tested with solutions based on Janus or MediaSoup and they work fine, and we confirmed that it's because TCP candidate is selected (no relay).  That's why we are in a rush to get ICE-TCP working asap with Licode (which is the best SFU for sure!)  We will assign few developers on this, but would love to get guidance from you first.  Please Help!

jcague:
@vpoddubchak I'm sorry I didn't see your message before. What issues do you see when receiving data from Licode?

jcague:
just a question that could minimize the problem, have you used TURN TCP to solve the issue? in that case, why is that solution not working?

vpoddubchak:
Thanks for looking into it!

> What issues do you see when receiving data from Licode?

I do such test with build from this PR:
1. Block udp traffic by such commands on dev machine (client and server is running there): 
  - ```iptables -I INPUT -p udp -s 0/0 -d 0/0 -m multiport ! --dports 80,443,53 -j DROP```
  - ```iptables -I INPUT -p udp -m udp --sport 53 --dport 1024:65535 -j ACCEPT```
2. Enter the to room by Chome client. Publish is happening by TCP protocol (ice-tcp is working for publisher)
3. Enter to the room by second tab of Chrome. Sometimes I see 1 frame from publisher, sometimes just black screen. In webrtc-internals of publisher I see that it drops bitrate to zero just after subscriber has been connected. In stats of subscriber I can see that icepair is TCP-based and use prflx tcp candidate. but data is not receiving, it stuck after 1 or 2 seconds.

Interesting thing: when I do p.2 after unblocking UDP traffic, all looks good: publisher is publishing by TCP, subscriber is receiving by UDP.

I thought it is RembManager blocks incoming traffic after subscriber appears because it is not designed to work with TCP. I tried to switch it off by removing remb from sdp of licode, but I see no difference.

> have you used TURN TCP to solve the issue? in that case, why is that solution not working?

We tried but it does not work. This is because client always can get UDP stun candidate and start to use it to receive data, so relay candidates ignores... Even during icerestart it takes udp stun again and again...

We are trying such workaround: before icerestart, configure peerConnection to force relay candidates. In this case, client takes udp stun on first attempt, but takes relay candidate after icerestart. This works in our lab tests but we did not test it in their "crazy network" yet. Anyway, it is just workaround and we need ice-tcp for such networks.

vpoddubchak:
@jcague tag you just to make sure you see my previous message.

jcague:
@vpoddubchak Today I was able to establish connectivity between Erizo and the client using your PR, but I had to update it to the latest changes in the master branch and used a different version of the nICEr library (I implemented a fix to use TCP). I can submit the changes to this PR.

Anyway, I was thinking that maybe the problem is bigger than just adding support to TCP candidates. There is another limitation in our current implementation of ICE (which is based on Firefox) as it does not support ICE aggressive nominations, and I guess that is the other part that would make it work for your case. We can try with the changes I implemented (using the latest code in Licode) and if it does not work I'm thinking on a bigger change in ICE.

vpoddubchak:
Thanks for doing it!
Submit whatever you want. 
I also tried to update nICEr and nrappkit (see my latest PRs) but it did not help with ice-tcp

vpoddubchak:
Hello @jcague!
Just checking if this is still on your radar as I am about to resume this activity on my side.  
Appreciate your answer & guidance.

PR #1694, opened 2021-04-16
Update client with setStreamPriorityStrategy

lodoyun:
**Description**

This PR fixes a race condition by properly updating the `Client` in ErizoController when `setStreamPriorityStrategy` is called.


jcague (approved):
good catch!

PR #1696, opened 2021-04-19
Stop padding if the distributor does not assign all the available

lodoyun:
bitrate

**Description**

This PR addresses an issue where erizo would keep generating padding unnecessarily when using the `StreamPriorityBWDistribution`.
If the distributor does not use all the available bitrate, the `PaddingManager` would still assign padding bitrate to the mediaStreams. We don't actually need that since the distributor has already assigned all the needed bitrate according to the defined strategy.
This PR addresses this by instructing the `PaddingManager` to stop generating padding when there's unneeded available bitrate after a pass of the distribution algorithm.

### DIFF
--- a/erizo/src/erizo/bandwidth/StreamPriorityBWDistributor.cpp
+++ b/erizo/src/erizo/bandwidth/StreamPriorityBWDistributor.cpp
@@ -12,7 +12,8 @@
 namespace erizo {
 DEFINE_LOGGER(StreamPriorityBWDistributor, "bandwidth.StreamPriorityBWDistributor");
 
-  StreamPriorityBWDistributor::StreamPriorityBWDistributor(StreamPriorityStrategy strategy): strategy_{strategy} {
+  StreamPriorityBWDistributor::StreamPriorityBWDistributor(StreamPriorityStrategy strategy,
+  std::shared_ptr<Stats>stats): strategy_{strategy}, stats_{stats} {
   }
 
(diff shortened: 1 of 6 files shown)


erizo/src/erizo/rtp/QualityManager.cpp
```diff
@@ -349,7 +349,7 @@ void QualityManager::forceLayers(int spatial_layer, int temporal_layer) {
 }
 
 void QualityManager::enableSlideShowBelowSpatialLayer(bool enable, int spatial_layer) {
-  if (enable_fallback_below_min_layer_ == enable && slideshow_below_spatial_layer_ == spatial_layer_) {
+  if (enable_fallback_below_min_layer_ == enable && slideshow_below_spatial_layer_ == spatial_layer) {
```
jcague, line 352: good catch! ;)

erizo/src/erizo/bandwidth/StreamPriorityBWDistributor.cpp
```diff
@@ -99,6 +100,7 @@ void StreamPriorityBWDistributor::distribute(uint32_t remb, uint32_t ssrc,
           remb, stream_info.stream->getId().c_str(), remaining_bitrate);
     }
   }
+  stats_->getNode()["total"].insertStat("unnasignedBitrate", CumulativeStat{remaining_bitrate});
```
jcague, line 103: 👍

jcague (approved):
LGTM

PR #1698, opened 2021-04-21
QualityManager cleanup

jcague:
**Description**

Remove some old code from QualityManager that was affecting the new Stream Prioritization code:

- QualityManager did not enable padding anymore, so I removed that code.
- ConnectionQualityCheck will not affect QualityManager anymore. We saw weird and complex behavior due to spurious packet losses.

Not needed.


lodoyun (approved):
This looks good to me 👍

PR #1701, opened 2021-04-23
Properly check is slideshow is enabled in QualityManager

lodoyun:
**Description**

This PR fixes a bad check in QualityManager to avoid calling `enableSlideshowBelowSpatialLayer` if the new call does not change the configuration


jcague (approved):
good catch!

PR #1703, opened 2021-04-27
Fix Safari publish problem

vpoddubchak:
After this [commit](https://github.com/lynckia/licode/pull/1659) Safari start to have problems with publishing.
As I understand, the same problem was with Firefox and it has been [fixed](https://github.com/lynckia/licode/pull/1669).

I did the same for Safari and it works in our environment.


erizo_controller/erizoClient/src/webrtc-stacks/SafariStack.js
```diff
@@ -0,0 +1,16 @@
...
+  const that = ChromeStableStack(specInput);
+
+  that._updateTracksToBeNegotiatedFromStream = () => {
+    that.tracksToBeNegotiated += 1;
+  };
+
+  return that;
+};
+
+export default SafariStack;
```
lodoyun, line 16: We need a new line at the end here

lodoyun (commented):
Good catch. I think this is a good solution 👍.
Please fix the linter issues. I think it's only the one I commented.
Thanks!

erizo_controller/erizoClient/src/webrtc-stacks/SafariStack.js
vpoddubchak: Done

lodoyun (approved):
👍

PR #1711, opened 2021-05-11
Fix issues with socket.io v2.4.0

jcague:
**Description**

Version 2.4.0 introduced some issues when the requests are proxied and the Origin header is set. Also, we now use a fixed version socket.io-client instead of a local copy of the library.

Not needed.


lodoyun (approved):
This looks good to me 👍

PR #1712, opened 2021-05-18
Adjust bandwidth estimation stats for stream priority distributor

lodoyun:
**Description**

This PR improves the use of stats in `StreamPriorityBandwidthDistributor` by:
1. Updating the stats stored in MediaStream more often (every 100ms at most)
2. Improving the stats stored in MediaStream so not only the last 500ms are taken into account but using the highest value for each second inside the 3 second window we are currently using
3. Adding to the calculated desired bitrate the same offset used in `QualityManager`

This PR also updates the tests accordingly. 


jcague (approved):
LGTM

PR #1714, opened 2021-05-25
Fix potential data race when updating targetIsMaxVideoBw

lodoyun:
**Description**

Remove the check that could cause a data race


jcague (approved):
LGTM

PR #1718, opened 2021-05-31
Fix the way we set the stream priority in ErizoAPI

jcague:
**Description**

I introduced a bug when solving conflicts between branches. This PR should fix it.

Not needed.


lodoyun (approved):
Good catch! this looks good to me.
Thinks like this make me wonder if we should be thinking of other less error prone ways to do this.

PR #1720, opened 2021-06-03
Support for multiple simulcast configurations for published streams in one connection

lodoyun:
**Description**

This PR moves simulcast configuration from the Stacks to Streams. That allows for publishing multiple simulcast streams with different configurations in the same connection.

### DIFF
--- a/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
+++ b/erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
@@ -96,40 +96,30 @@ const BaseStack = (specInput) => {
     return sdpInput;
   };
 
-  that.enableSimulcast = (sdpInput) => {
-    log.error('message: Simulcast not implemented');
-    return sdpInput;
-  };
-
-  const setSpatialLayersConfig = (field, values, check = () => true) => {
-    if (that.simulcast) {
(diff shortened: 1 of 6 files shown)


erizo_controller/erizoClient/src/webrtc-stacks/BaseStack.js
```diff
@@ -203,24 +194,71 @@ const BaseStack = (specInput) => {
...
+      .catch((e) => {
+        log.warning(`message: Error setting simulcast layer configs, error: ${e}`);
+      });
+  };
+
+  that.setSimulcastLayersConfig = (streamInput) => {
+    log.debug(`message: Maybe set simulcast Layers config, simulcast: ${JSON.stringify(that.simulcast)}`);
+    if (streamInput.hasSimulcast() && streamInput.getSimulcastConfig().spatialLayerConfigs) {
+      streamInput.stream.transceivers.forEach((transceiver) => {
+        if (transceiver.sender.track.kind === 'video') {
```
jcague, line 235: I know this is not from this PR, but I've seen issues with Unified Plan because transceiver.sender can be null sometimes, can you add a check here to avoid issues?

lodoyun:
Better version of this in #1726

PR #1721, opened 2021-06-04
Fix mediaStream priority initialization

lodoyun:
**Description**

This PR fixes an issue where a mediaStream would be initialized with an empty string when no priority is passed to erizoAPI constructor


jcague (approved):
LGTM

PR #1725, opened 2021-06-17
Fix stream republishing

jcague:
**Description**

We were not handling MIDs properly when the media section was reused. And that affected cases where clients republished their streams.

I also removed a couple of annoying things related to logs and repeated negotiation attempts that were always failing.

Not needed.


lodoyun (approved):
Good catch! LGTM 👍

PR #1727, opened 2021-06-21
Take into account passing "undefined" in updatesimulcastlayersbitrate

lodoyun:
**Description**

This addresses an unintended change in behaviour in updatesimulcastlayersbitrate. If we explicitly pass `undefined`, now the `maxVideoBW` will be assigned, which is the closest to what it meant before the maxVideoBW changes.


jcague (approved):
LGTM!

PR #1732, opened 2021-06-24
Update Basic Example with more options

jcague:
**Description**

I added more options to the Basic Example to give more flexibility when publishing and subscribing.

Not needed.


lodoyun (approved):
This looks good to me 👍

PR #1733, opened 2021-06-30
Fix issues with setMaxVideoBW and updateSimulcastLayersBitrate

lodoyun:
**Description**

This PR fixes 2 issues:
1. `updateSimulcastLayersBitrate` could update the structure passed as parameters if the bitrate specified was over the max. This could cause problems in the application if it happened to repeatedly use the same structure.
2. We were not updating the layers properly when setting MaxVideoBW. That could lead to a case where some of the layers did not comply with the maxVideoBW. 


erizo_controller/erizoClient/src/Stream.js
```diff
@@ -617,16 +625,14 @@ const Stream = (altConnectionHelpers, specInput) => {
   that.updateSimulcastLayersBitrate = (bitrates) => {
     if (that.pc && that.local) {
       // limit with maxVideoBW
-      const limitedBitrates = bitrates;
+      const limitedBitrates = Object.assign({}, bitrates);
```
jcague, line 628: good catch!

jcague (approved):
LGTM

PR #1735, opened 2021-07-01
fix(rtcp): repair GenericOneByteExtension memory leak.

johzzy:


erizo/src/erizo/rtp/RtpHeaders.h
```diff
@@ -197,6 +197,9 @@ class GenericOneByteExtension {
     char *payload = buf + 1;
     sprintf(data, "%*.*s", length, length, payload);  // NOLINT
   }
+  ~GenericOneByteExtension() {
```
jcague, line 200: good catch!

jcague (approved):
This looks good to me! thanks for the contribution!

PR #1736, opened 2021-07-12
Implement target connection bandwidth

lodoyun:
**Description**

This PR implements a way to set a `target bandwidth` for connections in Licode. This target can be set with as part of a strategy as `connectionTargetBw` or by calling  `room.setConnectionTargetBandwidth` in the client.
The main use of target bitrate is to provide a goal or a target for the bandwidth estimation. Licode will generate padding to try to reach the target bitrate.

### DIFF
--- a/erizo/src/erizo/rtp/RtpPaddingManagerHandler.cpp
+++ b/erizo/src/erizo/rtp/RtpPaddingManagerHandler.cpp
@@ -114,8 +114,11 @@ void RtpPaddingManagerHandler::recalculatePaddingRate() {
   bool can_send_more_bitrate = (kBitrateComparisonMargin * media_bitrate) < estimated_bandwidth;
   bool estimated_is_high_enough = estimated_bandwidth > (target_bitrate * kBitrateComparisonMargin);
   bool has_unnasigned_bitrate = false;
+  bool has_connection_target_bitrate = connection_->getConnectionTargetBw() > 0;
   if (stats_->getNode()["total"].hasChild("unnasignedBitrate")) {
-    has_unnasigned_bitrate = stats_->getNode()["total"]["unnasignedBitrate"].value() > kUnnasignedBitrateMargin;
+    has_unnasigned_bitrate =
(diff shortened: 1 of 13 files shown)


erizo/src/erizo/rtp/RtpPaddingManagerHandler.cpp
```diff
@@ -183,6 +186,7 @@ int64_t RtpPaddingManagerHandler::getTotalTargetBitrate() {
       }
       target_bitrate += media_stream->getTargetVideoBitrate();
     });
+  target_bitrate = std::max(target_bitrate, static_cast<int64_t>(connection_->getConnectionTargetBw()));
```
jcague, line 189: 👍

erizo_controller/erizoJS/models/Client.js
```diff
@@ -23,6 +23,7 @@ class Client extends EventEmitter {
     this.ioThreadPool = ioThreadPool;
     this.singlePc = singlePc;
     this.streamPriorityStrategy = Client._getStreamPriorityStrategy(streamPriorityStrategy);
+    this.connectionTargetBw = options.connectionTargetBw || 0;
```
jcague, line 26: I'm not sure I like this third way of adding it. Shouldn't we obtain the value from the streamPriorityStrategy?

```js
this.connectionTargetBw = this.streamPriorityStrategy.connectionTargetBw;
```

jcague (approved):
LGTM! 

I have a question: I understand that we can enable either it by calling setConnectionTargetBw(), by setting connectionTargetBw in the strategy, or by passing it as an option when publishing or subscribing the first stream, am I right?  I'm not sure I like the third option.

And I guess that the last API that is called updates the value, with no other preference or relationship between APIs.

I like having it

erizo_controller/erizoJS/models/Client.js
lodoyun: The value may have been updated, erizoController sends the last version with every addPublisher or addSubscriber. Here we're accounting for the case where this client was not present in this erizoJS

PR #1738, opened 2021-07-13
Fix initializing connection target bw

lodoyun:
**Description**

This PR fixes cases where the connectionTargetBw was not being properly taken from the strategy definition


jcague (approved):
LGTM

PR #1740, opened 2021-07-21
Close all connections before exiting erizoJS

lodoyun:
**Description**

This PR aims to fix a crash when exiting ErizoJS after all clients leave.
The crash is caused by a race condition when connections are not closed properly.


jcague (approved):
It looks good to me. I like this way of closing connections in order.

PR #1741, opened 2021-07-28
Add Backoff and ramp up periods in padding manager

lodoyun:
**Description**

This PR introduces improved backoff and ramp up periods for `RtpPaddingMangerHandler`, the goal is for the manager to avoid congesting the network whenever possible.

It only takes into account differences in bandwidth estimation. If the estimation decreases, it is detected as potential network problems and padding will stop for some time. After that backoff period, it will gradually ramp up (if the estimation does not decrease again) until it sends the calculated `target_padding_bitrate`.


jcague (approved):
This looks good to me! quite self-explanatory!

PR #1745, opened 2021-08-16
Update Mac deps and reduce CI time

jcague:
**Description**

Log4cxx and other dependencies have been updated recently (and not so recently), so I wanted to spend a PR to update a bunch of them.

I also use this PR to reduce the time it takes for CI to build and run tests. I removed the build debug version from the docker image.

Not needed.


scripts/installUbuntuDeps.sh
```diff
@@ -17,10 +17,10 @@ FAST_MAKE=''
 gcc_version=0
 
 check_version(){
-  if [[ $(lsb_release -rs) == "18.04" ]] || [[ $(lsb_release -rs) == "20.04" ]] 
-  then 
+  if [[ $(lsb_release -rs) == "18.04" ]] || [[ $(lsb_release -rs) == "20.04" ]]
```
lodoyun, line 20: Do we still need to keep this? I'm asking because I'm not sure this will even work with older gcc versions

scripts/installErizo.sh
```diff
@@ -115,16 +150,22 @@ then
   install_erizo_controller
   install_spine
 else
-  while getopts “heacstfd” OPTION
+  while getopts “heEaAcstfd” OPTION
```
lodoyun, line 153: 👍

lodoyun (approved):
This looks good to me. Just a question in the comments

PR #1746, opened 2021-08-31
Add ice disconnected timeout

lodoyun:
**Description**

We've detected instances where the PeerConnection IceState is stuck at `disconnected` after being connected for a while.
This PR adds a timeout to that state and triggers an ice restart if we detect it's stuck.


jcague (approved):
LGTM

PR #1751, opened 2021-09-07
Fix connection failed event 

lodoyun:
**Description**

Properly capture the `failed` message from the connection and emit a `connection-failed` event.


jcague (approved):
good catch!

PR #1766, opened 2021-11-08
Hack to avoid make sure we pass candidates in libnice even when there's only mdns

lodoyun:
**Description**

This PR introduces a hack to fix [this issue](https://gitlab.freedesktop.org/libnice/libnice/-/issues/69) until we can use an updated version of libnice with [this fix](https://gitlab.freedesktop.org/libnice/libnice/-/merge_requests/182).
In short, when connecting with Chrome and libnice with `ice-lite` there's a change we're not passing any remote candidates to libnice because they are mdns candidates and they are not valid. Instead of resolving those candidates (which may not always work), this hack just passes a "dummy" candidate to get libnice to progress to the next state.

### DIFF
--- a/erizo/src/erizo/LibNiceConnection.cpp
+++ b/erizo/src/erizo/LibNiceConnection.cpp
@@ -411,6 +411,17 @@ bool LibNiceConnection::setRemoteCandidatesSync(const std::vector<CandidateInfo>
     if (cinfo.hostPort == 0) {
       continue;
     }
+    std::string host_address = cinfo.hostAddress;
+    int host_port = cinfo.hostPort;
+
+    if (host_address.find("local") != std::string::npos) {
+      //  HACK for mdns. Theres currently an issue in libnice where if no valid candidates
+      //  are received - More details https://gitlab.freedesktop.org/libnice/libnice/-/merge_requests/182
+      //  We are forcing a "dummy" candidate to unlock the process.
+      ELOG_WARN("%s hostAddress is mdns, using dummy candidate", toLog());
+      host_address = "127.0.0.1";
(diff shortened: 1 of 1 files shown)


jcague (approved):
LGTM, it would be interesting to have reminders in GitHub to review hacks in the future

ranthwar-vcloudx:
@lodoyun This fix is working only for chrome and Edge....FF and safari is breaking with this changes....Can you suggest us any alternative fix for safari and FF....

lodoyun:
This is a problem specific to using libnice and, I've only been able to reproduce it with ice-lite.
How are you reproducing this? This is a pretty rare issue since, I've only seen it happen when both Licode and the browser are in AWS.
If that's your issue, disabling ice-lite (`config.erizo.iceLite = false`) or using nicer (`config.erizo.useNicer = true`) will fix it.

ranthwar-vcloudx:
Thanks for the reply.

We are using latest libNice module along with Trickle set to True..We are using Normal desktop browsers for FF,Safari.

Thanks..Will try those with that options...

Will there be any other issues/side affects if i turned above flags..??

Regards
Ram

lodoyun:
Using ice-lite and libnice helps with an issue that we detected that could cause users to go through TURN unnecessarily. Wether you ever experience that issue depends on your Licode deployment (particularly where your TURN servers are), you may not run into it at all. Other than that, I'm not aware of any other issue with libnice when iceLite is turned off. 
Nicer may be a bit riskier long-term since we're focusing on libnice now and don't plan to actively update Nicer in the future.

ranthwar-vcloudx:
Thanks @lodoyun .

PR #1767, opened 2021-11-15
Make connection quality level depend on strategy

jcague:
**Description**

Make Connection Quality Levels depend on the BW distributor strategy.

Possible quality levels are now: 'good', 'low' and 'very-low'.


erizo/src/erizo/bandwidth/StreamPriorityBWDistributor.cpp
```diff
@@ -32,6 +32,10 @@ void StreamPriorityBWDistributor::distribute(uint32_t remb, uint32_t ssrc,
   }
   ELOG_DEBUG("Starting distribution with bitrate %lu for strategy %s", remaining_bitrate, getStrategyId().c_str());
   strategy_.reset();
+
+  bool has_no_default_levels = false;
```
lodoyun, line 36: These variable names are really confusing to me. I think it's because I think of `fallback` and `slideshow` as "default levels". An alternative would be `has_no_spatial_levels` and `distribute_bitrate_to_temporal_levels` since we always expect temporal layers there anyway.
Other options are probably fine as long as we avoid `default` as it's not clear what default is in the context.

lodoyun (approved):
👍  I like this simple idea to get that signal from the distributor but I have two comments:
1. Variable names suggestion in comments.
2. Consider adding to the unit tests to cover this. I think we already have cases that cover both options and it would be easy to just check the output of the `tooLowBandwidhEstimation` call.

PR #1771, opened 2021-11-18
Give the option to reuse rooms with the same name

jcague:
**Description**

We were having race conditions when creating a room from two different endpoints at the same time. This PR gives the option to reuse rooms with the same name when we call createRoom() to simplify other's codebases.

N/A


lodoyun (approved):
This looks good to me 👍

PR #1772, opened 2021-11-23
Fix crash when calling enableFallback/Slideshow below min layer

lodoyun:
**Description**

`enableFallbackBelowMinLayer` and `enableSlideshowBelowMinLayer` are called from `StreamPriorityBWDistributor` in the `WebRtcConnection` thread. It is possible for the asyncTask in `MediaStream` to make those calls when the pipeline is already close but the `MediaStream` has not been destroyed yet. This results in a crash since it will end up calling `selectLayer` in `QualityManager` that tries to access the already removed pipeline directly.


jcague (approved):
LGTM

PR #1774, opened 2021-11-24
Add packet stats and improve padding bitrate assignment

lodoyun:
**Description**

This PR solves a potential issue when deciding which streams will send padding. The PaddingManager now filters and only assigns bitrate to those subscribed streams that have video that is not muted.

This PR also adds logs to keep track of packets received and sent per SSRC while also accounting for the video bitrate with or without padding.

### DIFF
--- a/erizo/src/erizo/MediaStream.h
+++ b/erizo/src/erizo/MediaStream.h
@@ -157,6 +157,8 @@ class MediaStream: public MediaSink, public MediaSource, public FeedbackSink,
   bool isAudioMuted() { return audio_muted_; }
   bool isVideoMuted() { return video_muted_; }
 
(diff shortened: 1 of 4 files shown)


erizo/src/erizo/MediaStream.h
```diff
@@ -157,6 +157,8 @@ class MediaStream: public MediaSink, public MediaSource, public FeedbackSink,
   bool isAudioMuted() { return audio_muted_; }
   bool isVideoMuted() { return video_muted_; }
 
+  bool canSendPadding() { return !isPublisher() && hasVideo() && !isVideoMuted(); }
```
jcague, line 160: 👍

jcague (approved):
Good catch! I also like the new additions to the stats.

PR #1776, opened 2021-12-12
Notify about low conn quality levels if bw is not enough for first layers

jcague:
**Description**

I was not taken into account if we distributed BW into layers but with not enough bandwidth to send the very first temporal layer.

Not needed.


lodoyun (approved):
This looks good to me 👍

PR #1777, opened 2021-12-14
Fix to playing files and rtsp streams

Arri98:
The reproduction of rtsp streams and files was failing as the externalInput was not up-to-date with other changes. When subscribing to the externalInput stream erizoJS failed with the following error:

TypeError: this.publisher.getMaxVideoBW is not a function
    at Subscriber.setMaxVideoBW (/licode/erizo_controller/erizoJS/models/Subscriber.js:54:42)
    at new Subscriber (/licode/erizo_controller/erizoJS/models/Subscriber.js:33:10)
    at ExternalInput.addSubscriber (/licode/erizo_controller/erizoJS/models/Publisher.js:49:24)
    at Object.that.addSubscriber (/licode/erizo_controller/erizoJS/erizoJSController.js:352:28)
    at /licode/erizo_controller/common/amqper.js:125:36
    at Message.<anonymous> (/licode/erizo_controller/node_modules/amqp/lib/queue.js:209:28)
    at Message.emit (events.js:210:5)
    at Message.EventEmitter.emit (domain.js:476:20)
    at Queue._onContent (/licode/erizo_controller/node_modules/amqp/lib/queue.js:550:25)
    at AMQPParser.self.parser.onContent (/licode/erizo_controller/node_modules/amqp/lib/connection.js:142:32)

Also the video was not being played because the payloadType did not match the negociated one. 

In the negotiation vp8 had payloadType 100 but in the MediaProcessor.cpp  the payloadType was being set to 96.
Captura de pantalla de 2021-12-14 13-35-24


erizo/src/erizo/media/MediaProcessor.cpp
```diff
@@ -509,7 +509,7 @@ int OutputProcessor::packageVideo(unsigned char* inBuff, int buffSize, unsigned
         rtpHeader.setTimestamp(av_rescale(pts, 90000, 1000));
     }
     rtpHeader.setSSRC(55543);
-    rtpHeader.setPayloadType(96);
+    rtpHeader.setPayloadType(100);
```
lodoyun, line 512: This won't work in general since the PT of VP8 is defined in  `rtp_media_config.js`.
The proper solution here would be to pass the mappings in the `ExternalInput` constructor, just like we do for `ExternalOutput` and handle it properly.

erizo_controller/erizoJS/models/Subscriber.js
```diff
@@ -51,9 +51,14 @@ class Subscriber extends NodeClass {
         Math.min(this.publisher.getMaxVideoBW(), this.maxVideoBW) : this.maxVideoBW;
       this.maxVideoBW = updatedMaxVideoBW;
     } else {
-      updatedMaxVideoBW = this.publisher.getMaxVideoBW();
+      // eslint-disable-next-line no-lonely-if
+      if (this.publisher.getMaxVideoBW) {
+        updatedMaxVideoBW = this.publisher.getMaxVideoBW();
+      } else {
+        updatedMaxVideoBW = 0;
```
lodoyun, line 58: This means all subscribers will have a maxVideoBW of 0 for this stream. Have you checked that this causes no issues? AFAIK we don't really support not having a value here. It does have less implications when not in simulcast like in this case, but it's weird nonetheless. We would be generating REMBs with 0.
I would implement a solution where we set a maxVideoBW for the ExternalInput.

lodoyun, line 58: This means all subscribers will have a maxVideoBW of 0 for this stream. Have you checked that this causes no issues? AFAIK we don't really support not having a value here. It does have less implications when not in simulcast like in this case, but it's weird nonetheless. We would be generating REMBs with 0.
I would implement a solution where we set a maxVideoBW for the ExternalInput.

erizo_controller/erizoJS/models/Publisher.js
```diff
@@ -222,7 +222,7 @@ class Source extends NodeClass {
       if (this.ei && this.mediaStream.periodicPlis) {
         clearInterval(this.mediaStream.periodicPlis);
         this.mediaStream.periodicPlis = undefined;
-      } else {
+      } else if (this.mediaStream.setPeriodicKeyframeRequests) {
```
lodoyun, line 225: 👍

lodoyun, line 225: 👍

lodoyun (changes requested):
Thanks for looking into the problems. Now that we have a clear diagnose of what's happening, I would prefer the root issues to be resolved in order to merge this.

PR #1797, opened 2022-02-04
Update webrtc library

lodoyun:
**Description**

We use google's webrtc code for a few things, including fec and bandwidth estimation. This PR updates the code.
To make that update work this PR also:
* Updates to gcc10 to be able to build webrtc + abseil properly in Ubuntu 20.04. Sadly, this implies that I had to start ignoring some warnings (i.e. they are no longer errors). Those warnings are all from libraries that we can't or won't update at the moment (libav, nan)
* Updates libsrtp

Regarding Licode functionality this should affect:
* Sender Bandwidth estimation
* Receiver side bandwidth estimation
* Fec receiver handler. 


scenesea:
Hi, it's very helpful that webrtc library is updated. 
But i have a question, what version or commit id of webrtc libray does this update relate to?

PR #1265, opened 2018-07-18
Fix for recording audio and screensharing streams

zafergurel:
**Description**
ExternalOutput does not record audio-only or video-only (screensharing) streams.

The changes introduced in this commit lets ExternalOutput to be aware that the stream is an audio-only or video-only stream.

Also please look at https://github.com/lynckia/licode/pull/1254 .
The comments about this PR made this PR possible.

Issues #1203 and #1185 are related and can be fixed with this PR.

[x] It needs and includes Unit Tests

**Changes in Client or Server public APIs**
ExternalOutput in erizoAPI has been changed. 
A new method named setHasAudioAndVideo has been added to ExternalOutput (ExternalOutput.h and ExternalOutput.cc).

PR #1808, opened 2022-03-11
Fixes to recording

Arri98:
**Description**

Fixed some issues with recording:

- Licode would not record streams that didn't have both video and audio. This was fixed by @[zafergurel](https://github.com/zafergurel) in #1265. Made minor changes for the linter so all credits to him.
- When stopping the recording, a wrong ID was passed resulting in the stream continued being recorded. This was fixed by @[nvazquezg](https://github.com/nvazquezg) in #1565
-  When checking if the stream had audio the condition if (getAudioSinkSSRC() == 0) was always returning false even if the stream had audio. In this if block the codec was set to PCMU so if other codec was used the resulting audio is corrupted. 

Also removed the toggleRecording button in the basic example and changed it for individual buttons so the user can select the stream to record.

No changes

No changes

### DIFF
--- a/erizo_controller/erizoAgent/log4cxx.properties
+++ b/erizo_controller/erizoAgent/log4cxx.properties
@@ -34,7 +34,7 @@ log4j.logger.dtls.DtlsSocketContext=WARN
 log4j.logger.dtls.SSL=WARN
 
 log4j.logger.media.ExternalInput=WARN
-log4j.logger.media.ExternalOutput=WARN
+log4j.logger.media.ExternalOutput=DEBUG
 log4j.logger.media.InputProcessor=WARN
 log4j.logger.media.OneToManyTranscoder=WARN
 log4j.logger.media.OutputProcessor=WARN
--- a/erizo_controller/erizoAgent/log4cxx.properties
+++ b/erizo_controller/erizoAgent/log4cxx.properties
@@ -33,8 +33,8 @@ log4j.logger.dtls.DtlsFactory=WARN
 log4j.logger.dtls.DtlsSocketContext=WARN
 log4j.logger.dtls.SSL=WARN
 
-log4j.logger.media.ExternalInput=DEBUG
-log4j.logger.media.ExternalOutput=DEBUG
+log4j.logger.media.ExternalInput=WARN
+log4j.logger.media.ExternalOutput=WARN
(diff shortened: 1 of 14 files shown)


erizo_controller/erizoJS/models/Publisher.js
```diff
@@ -98,11 +98,16 @@ class Source extends NodeClass {
...
     const eoId = `${url}_${this.streamId}`;
+    const hasVideo = options.hasVideo === null || options.hasVideo;
+    const hasAudio = options.hasAudio === null || options.hasAudio;
+
     log.info(`message: Adding ExternalOutput, id: ${eoId}, url: ${url},`,
       logger.objectToLog(this.options), logger.objectToLog(this.options.metadata));
     const externalOutput = new erizo.ExternalOutput(this.threadPool, url,
       Helpers.getMediaConfiguration(options.mediaConfiguration));
     externalOutput.id = eoId;
+    externalOutput.setHasAudioAndVideo(hasAudio, hasVideo);
```
lodoyun, line 109: Why not just pass `hasAudio` and `hasVideo` in the constructor just like we do for MediaStream?

lodoyun (commented):
In general I think this looks good. I do have a question that I feel would avoid adding new methods to set the hasVideo and hasAudio flags when we already have that info when the object is created

erizo_controller/erizoJS/models/Publisher.js
Arri98: Makes sense, I will update it

erizo_controller/erizoAgent/log4cxx.properties
```diff
@@ -34,7 +34,7 @@ log4j.logger.dtls.DtlsSocketContext=WARN
 log4j.logger.dtls.SSL=WARN
 
 log4j.logger.media.ExternalInput=WARN
-log4j.logger.media.ExternalOutput=WARN
+log4j.logger.media.ExternalOutput=DEBUG
```
lodoyun, line 37: consider making this `WARN` again

lodoyun (approved):
Looks good to me 👍  but please fix the log level as indicated in the comment
Thanks!

zafergurel:
Thanks @Arri98 for giving credit and for your contribution.

PR #1812, opened 2022-04-25
Fix recording not stopping

Arri98:
There was an error where when recording where the recording would not stop and files kept growing even though licode gave no error.  What I found is that the promise that removed the ExternalOutput was always pending and never resolved and the callback was never executed but, the recording would stop on most occasions. After removing the lines shown in the PR the callback is always called and the recording stops all the times I tested it. 

To check this you check that the log inside the callback in the removeExternalOutput() function on line 115 of erizo_controller/erizoJS/models/Publisher.js is never reached. 

I don´t know the root issue of this problem, and my manual testing is limited so this might cause other issues with removing that check. I will leave this PR as a draft, as a point of discussion and potential fix until a better solution is found or further testing is done. 

### DIFF
--- a/erizo/src/erizo/media/ExternalOutput.cpp
+++ b/erizo/src/erizo/media/ExternalOutput.cpp
@@ -105,6 +105,7 @@ void ExternalOutput::syncClose() {
   if (!recording_) {
     return;
   }
+  recording_ = false;
   // Stop our thread so we can safely nuke libav stuff and close our
   // our file.
   cond_.notify_one();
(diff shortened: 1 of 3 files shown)


Arri98:
After further investigation i have found the root of the problem. In the ExternalOuput, when closing it, we would wait for the recording thread to finish before freeing memory allocated to libav componentes. However, . The file im going to comment is /src/erizo/media/ExternalOutput.cpp.

In line 110 and 111 we unlock the recordingThread and wait for it to finish. However as the recording_ flag is never set to false it  never leaves the loop in line 545. Queues are not depleated and libav componentes are not liberated in subsecuent lines. The debug message in line 134 never appears as we never reach this point. 

Also as we never closed the externalOuput, in the AsyncCloser class inside the ExternalOutput.cc we never fulfilled the promise, so the callback was also never executed. 

For the these reasons the audio and video queues would never deplete and the last seconds of the recording were lost. Also there was a small memory leaks and some open connections after closing recorders.

lodoyun (approved):
This look good to me 👍  Good catch

PR #1877, opened 2023-03-22
Fix google closure compiler for macosx

lodoyun:
**Description**

The version we were using of the google-closure-compiler crashed in Mac OS X ventura because of an issue with a dependency - [upx](https://github.com/upx/upx/issues/612)

This PR updates it to a more recent version resolving the issue.


Arri98 (approved):
Tested on MacOSX Ventura and compiles without errors.

